Issue #16, opened 2018-11-08
PATTERN - Data Protection Banner


sturobson:
We need to create/replicate the GDPR data protection banner for use across all EMBL sites.

Currently the pattern can [be found here](https://www.ebi.ac.uk/style-lab/websites/patterns/banner-data-protection.html).

the existing pattern uses data attributes for the recreating the content using the CSS `content` selector.

This does not get read by screen readers (VO on MacOS tested), so we need to use 'actual content' moving forward. I don't see this as a problem? A WP plugin(?) could be created for just this thing? For example.

Also from an accessibility (a11y) viewpoint it needs to be near the top of the DOM on page load, so this is something we would need to add to documentation for people who would be implementing this.

sturobson:
Ah, wait a minute -- I'm making use of `content` in CSS to put the `data-message` in place.

Are you saying/thinking that we get JS to inject it where it needs to be? 

(Yes, I've not looked to deeply at the current implementation, and when straight to `content`).

sturobson:
![screenshot 2018-11-08 at 10 57 32](https://user-images.githubusercontent.com/925197/48194524-3074a700-e345-11e8-8c2d-00cf0f88d232.png)

sturobson:
1. do inline links in text that has a background colour stick with the blue or match the text colour and make use of the underline to denote a link?
2. is this a good place to use the green/blue CTA?
3. Is the dark gray better than the black?

(note: that width of text has been increased to a max-width of 60em, so it doesn't look too far away from the button (unless you're on a massive monitor).
4. Should the text just run up to the button regardless of monitor size?
5. Would it be better to have a 'floating' notice that has a max-width so the text is editable?

khawkins98:
Had a look at this, I think it's good and plenty for now (any design comments from Mark aside). I'll make a new issue to write out some user stories around what the banner needs to do (some of these are special to our fragmented developer scenario).

khawkins98:
There's a couple little rough things here I'll make a PR for:
- Banner is `absolute` and not `fixed`
- Banner content area spills out to the whole screen

I think we can also leverage `vf-body` and `vf-grid` for some reusability. 

Currently:

![image](https://user-images.githubusercontent.com/928100/48254557-a42cb780-e40a-11e8-8167-82dc751a6c37.png)



Updated:

![image](https://user-images.githubusercontent.com/928100/48254362-249ee880-e40a-11e8-94f0-1e88aa1dde9f.png)

PR #18, opened 2018-11-08
wip: data protection (GDPR) component

sturobson:
An updated version of the data protection banner, using the current v2.0 brand colours and typography.

This is in relation to #16 

notes:

- The JS is a little rudimentary and probably needs more work.
- This doesn't set any cookies (see above).
- The green colour choice for a battery is just a suggestion

questions:

- Do we have a defined 'use this' for the content that needs to be displayed in this notice?
- I would assume that the phrasing and terminology used would be the same across all sites using this?

### DIFF
--- a/assets/scss/styles.scss
+++ b/assets/scss/styles.scss
@@ -63,6 +63,7 @@ html, button {
 @import 'vf-breadcrumbs/vf-breadcrumbs.scss';
 @import 'vf-factoid/vf-factoid.scss';
 @import 'vf-snippet/vf-snippet.scss';
+@import 'vf-data-protection-banner/vf-data-protection-banner.scss';
 @import 'vf-code-example/vf-code-example.scss';
 @import 'vf-page-header/vf-page-header.scss';
 @import 'vf-link-list/vf-link-list.scss';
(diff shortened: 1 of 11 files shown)


khawkins98:
I think it's good and plenty for now (any design comments from Mark aside). 

Few comments/use cases that extend from the current EBI VF setup:
- Cookie
  - Obv need a vanilla JS way to set/store cookie
  - Cookie needs to pre-fixed `siteNameID` + `dataProtectionVersion` + `data-protection-accepted`
- Message
  - Only shown if the `siteNameID` + `dataProtectionVersion` isn't `true`
  - Uses a default text + links unless a custom message is provided
  - Default message should be loaded from a centrally managed script (so we can globally update the message and version number without re-deploying 700 sites
      - We could use the [existing announcement JS file](https://dev.ebi.emblstatic.net/announcements.js) that lives on the CDN

Long story/short: we could go with the PR for now, do any design tweaks from Mark, and then make a ticket to implement the above (soon)

sturobson:
Cool. I’ll merge this now and add a new ticket on the 12th November.

PR #20, opened 2018-11-09
Update vf-data-protection-banner rough edges

khawkins98:
for #16

### DIFF
--- a/components/blocks/vf-data-protection-banner/vf-data-protection-banner.config.yml
+++ b/components/blocks/vf-data-protection-banner/vf-data-protection-banner.config.yml
@@ -1,5 +1,6 @@
 title: Data Protection Banner
 label: Data Protection
+preview: '@preview--nogrid'
 context:
   data-service-id: MyService
   data-protection-version: 0.1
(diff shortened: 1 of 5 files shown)


sturobson:
So, we've hit on an issue that I didn't foresee when implementing a 'simple' grid. 

Some items will fill the height of their grid item dependant on their siblings.

screen shot 2018-11-11 at 10 44 46

sturobson:
I think this also shows a problem of how we're exposing the grid in the preview files.

sturobson:
@khawkins98 if you're happy with this now - I think we can push this to `develop`

khawkins98:
Just made an edit to use `.vf-grid` in place of `.embl-grid` in the HTML markup. 

Functionally it's the same(*) and it leaves open the possibility of having this be a non-`vf-embl` specific pattern. 

*There's a lack of left/right padding on mobile, but I think that's more a symptom of an issue in `vf-grid`? do we need this code:

```
 @media (max-width: 1299px) {
    .vf-grid {
      grid-column: 1 / -1;
    }
  }
```

sturobson:
There's a lack of padding on mobile as `preview-nogrid.hbs` doesn't have a `vf-body` on it - which would have the left right padding which goes from 16px to 'as big as it needs'

khawkins98:
Sounds good -- I have a couple of ideas about "messaging" in the VF, but think this is plenty good enough for me.

PR #21, opened 2018-11-09
Deploy dev branch to dev CDN, pattern library

khawkins98:
As we're still in early alpha, it probably makes sense to deploy all edits to the develop branch (particuarly so the edits get out to the CDN)


khawkins98:
I think Stu's not around today, so running with this.

PR #38, opened 2018-11-13
updates to remove the iframe preview padding

### LINKED ISSUE #37: BUG - remove the left and top padding from the preview iframe
At some point, I thought it was a good idea to add a little padding to the iframe (I think it's mainly because I don't like the border). 

Looking at certain patterns - that choice looks completely silly. This should be removed.

sturobson:
This removes the 12px left/top padding that is on the preview iframe.

This closes #37

### DIFF
--- a/assets/scss/_shame.scss
+++ b/assets/scss/_shame.scss
@@ -2,7 +2,7 @@
 
 .Preview-iframe {
   border: 1px solid #b3b3b3;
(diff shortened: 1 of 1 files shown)


khawkins98:
Looks good. Hopefully the queziness underlying this will be eliminated once we improve the layout.

PR #42, opened 2018-11-20
Add new responsive logo pattern variant

khawkins98:
Specifically done to test Tabea's new compact logo styles, but could see this approach being useful to other users of the VF.

resize-logo

### DIFF
--- a/components/blocks/vf-global-header/vf-global-header.hbs
+++ b/components/blocks/vf-global-header/vf-global-header.hbs
@@ -1,5 +1,5 @@
 <div class="vf-global-header">
-  {{render '@vf-logo'}}
+  {{render '@vf-logo--responsive'}}
 
   {{render '@vf-navigation--tertiary'}}
 </div>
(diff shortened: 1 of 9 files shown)


khawkins98:
Thinking on this a bit more, it would make sense if the logo optimised to be smaller -- was smaller in px as well as on physically smaller device.

resize-logo-2

markboulton:
This is great. As we switch this, we should also reduce the height of the top bar so the exclusion zone around the logo remains consistently above the minimum. The challenge might be that the exclusion zone in digital is typically smaller than that of print, so we'll need to chat with Tabea about it.

sturobson:
So, when Tabea showed this to me as an idea - it got to the point where EMBL disappeared and the logo was the only thing seen.

Shall we work on doing that now … or should we pull this in as a v0.0.1a? 

I'm happy with either.

khawkins98:
Nice idea -- maybe we make the smallest logo just the logo to spur discussion?

PR #45, opened 2018-11-21
patterns/embl-meta-tags

khawkins98:
Adds an initial spec for the meta tags. These can be queries by other patterns (such as the breadcrumb or global nav) to provide appropriate navigation and look/feel.


sturobson:
I'll merge these in, but have some questions or reservations…

Is this more a 'tool' than a design pattern/component? 

If it's not a 'tool' then do we need another folder within the component library for things like this?

Do we need to show these in the component list? Or can we hide It and include it in the documentation instead?

sturobson:
I'll add those as an issue to discuss further.

khawkins98:
Good questions, "tools" might make sense. We'll probably have a few of these... metatags, analytics helpers... the no-js is probably also another one.

sturobson:
At least we can pull in patterns into other documentation pages so we can 'hide' them from 'components' but still pull in their actual code / output elsewhere

PR #47, opened 2018-11-22
Add add a link to the pattern readme boilerplate

khawkins98:
Makes it easier to get to the online information from github. 

Does nothing to help existing patterns 😢 ... that's a #todo, or if I get a few mintues later.

#changelog #templates


khawkins98:
Hrm, of course then the pattern library would link to itself. Will give this more thought...

Issue #24, opened 2018-11-11
BUG: vf-tabs js querySelector on null


khawkins98:
The js for vf-tabs will generate an error if matching html isn't found:

The code ([vf-tabs.js#L5](https://github.com/visual-framework/vf-core/blob/master/components/blocks/vf-tabs/vf-tabs.js#L5))
```
  const tablist = tabbed.querySelector('.vf-tabs__list');
```

Can result it (when no tab pattern is present):
```
  Uncaught TypeError: Cannot read property 'querySelector' of null
```

You can see it anywhere the script.js is being used.

I'm not quite sure the best way to deal with exiting when no content found, but probably something like:

``` js
(function() {
  // Get relevant elements and collections
  const tabbed = document.querySelector('.vf-tabs');
  const tabbedContent = document.querySelector('.vf-tabs-content');
  if (tabbed === null || tabbedContent === null) { 
    console.log('tabs or content not found');  // we probably need a vfIfDebugOn?
    return;
  }
  const tablist = tabbed.querySelector('.vf-tabs__list');
  const tabs = tablist.querySelectorAll('.vf-tabs__link');
  ...
```

PR #49, opened 2018-11-23
Fix uncaught type error in vf-tabs 

khawkins98:
For #24

This also:
- allows vfTabs to be reinvoked/called (dynamic content)
- resets active tabs on init

#changelog #patterns #javascript


sturobson:
closes #24

Issue #28, opened 2018-11-12
QUESTION - Do we publish patterns independently or as part of one big release?


sturobson:
We will be [making use of Lerna](https://github.com/lerna/lerna) to help publish the components of the visual framework to NPM.

One decision that needs making...

Do we allow each component to be published independently? 
ie: if we update the buttons we can push the new version of that package up with it's version bump

Or do we publish the component library as a whole(fixed)?
ie: if we update the buttons we push a version bump to all packages.

A part from 'hot fixes', would it be better for our audience to 'schedule' releases (in time) which means a 'publishing as a whole' would work better?

Another idea is we could start with indepentently (as we're in alpha) and move to fixed when the VF is at Beta or live?

khawkins98:
Adding a footnote that this versioning method could (should?) also apply the VF WP theme (`vf-wp`).

So `vf-wp:2.0.10` would expect atleast `vf-core:^2.0.0`. And the `vf-wp` for EMBL site would also require `vf-embl:^2.0.0`.

That's my take, anyways.

sturobson:
I was talking about this w/ @dbushell yesterday. Unless it's a breaking change (v1.2.3 to v2.0.0) where we are breaking the HTML / class name convention (mainly, but there could be other reasons for the breaking change) ... David was thinking that a link to the CDN of the (versioned) .css file would be a easier / better solution.

(I guess that idea should really be it's own issue within the`vf-wp` repo)

khawkins98:
Sounds about right to me. The generic `vf-wp` theme will be quite general in usage, so it probably makes sense to use the kitchen sink CSS+JS from the CDN.

That also makes it possible to have an admin field of "Want a different look? Link to your VF 2.0.X compatible CSS and JS here."

sturobson:
@khawkins98 as part of the build process to put the new build the visual framework live - can we add a CHORE (new label needed) to make it spit out the CSS and JS to a CDN somewhere>

khawkins98:
@dbushell Sounds great. I think that would make the devs at EBI (who are responsible for maintenance) really happy, too. 

@khawkins98 Do you mean the kitchen sink CSS/JS, or the individual component CSS/JS?  The former is already there at:
- https://dev.assets.emblstatic.net/vf/scripts/scripts.js
- https://dev.assets.emblstatic.net/vf/css/styles.css

@pwalter-ebi Just occurred to me we probably need to do a version number on those, a la: 
[...]

PR #55, opened 2018-11-28
Document interoperability of pattern versions

khawkins98:
Relates to #28

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,14 @@ We support the last two major releases with bug fixes and branding. New features
 
 Where's version 1.X, you ask? That's the [EMBL-EBI Visual Framework](https://github.com/ebiwd/EBI-Framework) from where this concept [originated, and evolved](https://blogs.embl.org/communications/2018/09/12/faster-scientific-websites-through-reusability/).
 
+### Versioning of patterns
+
+Pattern versions follow the major releases for compatibility. That is:
+
+- All patterns for `v2.0` should be comptabile with eachother;
+- It is not intended to use a `vf-pattern:2.1.x` pattern with `vf-core:2.0.x`; and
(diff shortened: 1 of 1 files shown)


khawkins98:
@sturobson now that we're ready to npm (🍾), following up on previous convos about pattern minor/major version compatibility. I think this is where we landed?

sturobson:
I think we're both seeing versioning points differently

`v1.2.3` … the 1 being the major release, the 2 being the minor release and the 3 being the patch release.

therefore, I think:- 

anything that's 2.x.x should work with each other as we shouldn't be introducing breaking changes 
here.

there's chance to do pre-release commits with Lerna too - so perhaps we can start including those (in the future) as 'this is new, it shouldn't break anything, but it's here if you want to check'

Also - we should look at getting the Visual Regression testing set up so it's more automated - I'll look.at this. This way we can see if changing something in `vf-button` has broken the `vf-job-description--snippet` pattern that uses it.

sturobson:
I'll merge this in now to get things underway to lead to more discussion over versioning and how other patterns would/could use newer/older patterns

khawkins98:
Funny. Was just about to comment. 

I think what you say here is fine -- generally the overall architecture should remain the same for all `v2.x`s.... just like how we've now gone from `v1.x` to `v2.0` for `vf-core`. And while minor versions might not be plug-and-play compatible, they should fundamentally be able to work together... just like how a  jQuery 3.2 plugin might require jQuery core 3.1+.

So revised language:

```
### Versioning of patterns

Pattern versions follow the major releases for compatibility. That is:

- All patterns for `v2.0` should be architecturally compatible with each other;
- Some patterns might require specific minor versions, a `vf-pattern:2.2.x` might require `vf-core:>2.1.x`; and
- Minor versions of packages should generally be inter-compatible to use `vf-pattern-a:2.0.12` with `vf-pattern-b:2.0.2`
```

PR #61, opened 2018-11-28
Adds LernaJS for individual package deployment to NPM

sturobson:
I think we're at a point where we can merge this in.

Need to document 'process' as there's an additional bit of cli work needed for new patterns (as it's a scoped org)


sturobson:
I'll get to these conflicts

sturobson:
the conflicts fixed --- 

- everything is back to it's flat folder structure
- 'boilerplates' are it's own thing still -- did we discuss what these were?

PR #82, opened 2018-12-06
Pattern/meta information

sturobson:
should this actually be - `vf-article-meta`? or something else entirely? 

- we need to look at making a 'grid' for articles - as the positioning of this on a desktop width browser needs some other content to be told what to do.


khawkins98:
Do think we should call it `vf-article-meta-information` as it should be clear this is separate from any page meta information pattern.

Also thought we need to account for linking of the name and image, have taken a stab at that.

We probably also should consider the addition of position/title as well, but that can wait until someone squeaks...

PR #83, opened 2018-12-06
updates factoid

sturobson:
adds a 24px margin-bottom to the factoid. 

With this and the embl-grid container and vf-divider spacing added yesterday we're

a) pushing down
b) needing to standardise the spacing
  - element
  - block
  - container

c) last-child - does it have a margin or not, or does it rely on it's parents padding / bottom margin?


khawkins98:
last-child margin always makes me nervous, I'd say lets see how far the container padding get us.

PR #94, opened 2018-12-11
Refactor - sans serif typography

sturobson:
updates `font-size`, `font-weight`, and `line-height` for the Plex sans-serif typeface:

- updates design tokens
- updates Sass variables
- removes Sass monospace code and `if/else` statement in the `_typography.scss` mixin
- updates any `@include set-type` to new sizes
- updates any classnames in components to include new sizes

- adds depreciated status
- marks job-description snippet, news item as depreciated.

### DIFF
--- a/assets/scss/styles.scss
+++ b/assets/scss/styles.scss
@@ -8,9 +8,13 @@
  *
  */
 
+
+
 @import 'vf-global-custom-properties.scss';
+
 @import 'vf-global-variables.scss';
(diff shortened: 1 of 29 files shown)


sturobson:
Yeah, because of the font size changes and reduction in the number thereof. I refactored it - which has incurred some classname changes. 

Would’ve been a perfect text case for visual regression testing :(

PR #98, opened 2018-12-11
Fix/vf global header centering

khawkins98:
A few follow-on fixes


sturobson:
as `vf-global-header` should be inside `vf-header` we should rely on `vf-header` for the margin-bottom to push down the next item in the page structure.

PR #100, opened 2018-12-12
Feature/pattern defaults

khawkins98:
Proof of concept on having `-default` classes for patterns.


khawkins98:
I think this is redundant now. Closing.

PR #103, opened 2018-12-12
creates a badges component to replace tags - deprecates some packages

sturobson:
adds `vf-badge` component to be used instead of `vf-tag`

deprecates: 

- `vf-tags`
- `vf-job-description` (we need to see how we can do this _just_ for the snippet, as it's wholesale at the moment).
- `vf-snippet`
- `vf-news-item`

in those components it:

- adds a deprecated `vf-badge` to the components `.yml` file.
- adds a link to the new patter in the components `.md` file.

This PR also:

- moves some fractal CSS to the `_shame.scss` files
- updates the fractal `.nunj` patterns to use `vf-badge` instead of `vf-tag`
- adds a `patternType` for deprecated so they're listed in the patterns, but at the bottom.


sturobson:
note: this is should not be a  **breaking change** as we are still including `vf-tag` but upgrading to `vf-badge` should be on the priority list for applications and sites using it.

PR #106, opened 2018-12-17
Make typography sizing high-specificy in styles.scss

khawkins98:
This way if/when a dev does drop a `.vf-text--body-r`, it will take priority.

### DIFF
--- a/assets/scss/styles.scss
+++ b/assets/scss/styles.scss
@@ -49,9 +49,7 @@ html, button {
 
 
 /* All Visual Framework Elements */
-@import 'vf-heading/vf-heading.scss';
 @import 'vf-badge/vf-badge.scss';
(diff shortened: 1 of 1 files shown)


assets/scss/styles.scss
```diff
@@ -98,17 +96,17 @@ html, button {
...
-
-
 /* All Visual Framework Deprecated Patterns */
-
 @import 'vf-tag/vf-tag.scss';
 @import 'vf-job-description/vf-job-description.scss';
 @import 'vf-news-item/vf-news-item.scss';
 @import 'vf-snippet/vf-snippet.scss';
 
+/* All Visual Framework Utility and high-specificity patterns */
```
khawkins98: Move these after deprecated patterns to maintain priority of utility patterns.

PR #123, opened 2018-12-26
Support multiple tabs

khawkins98:
Also allows for nested tabs, cleans up some JS a bit too.

Previously if you had more than one set of tabs on a page, it just fell apart.

I think some of this starts to raise a question about how much JS we should be writing... is there some sort of lightweight framework that can do things like tabs for us .. making a new issue for that.


sturobson:
Neat feature - although I'd've thought it's not the best UX to nest tabs into tabs? At least there's the option now :)

khawkins98:
Accidental support 😁 -- by supporting more than one tab on the page, it also supports nested tabs.

PR #126, opened 2019-01-04
Docs/onboarding

khawkins98:
A bunch of documentation tweaks to aid in the onboarding.

It also links off to the "pre-alpha" child theme, we can find/replace those as the concept evolves and we have a final name/correct URL.


components/embl-logo/assets/logo-medium.svg
```diff
@@ -0,0 +1,56 @@
+<?xml version="1.0" encoding="utf-8"?>
```
sturobson: I'll look at some SVGO gulp plugin to avoid:

a) the need to optimise this by hand
b) further PRs with all that Adobe cruft in the SVG

gulpfile.js
```diff
@@ -1,5 +1,18 @@
...
+
+// -----------------------------------------------------------------------------
+// Configuration
+// -----------------------------------------------------------------------------
+
+const SassInput = './assets/scss/styles.scss';
+const SassOutput = './public/css';
+const autoprefixerOptions = { browsers: ['last 2 versions', '> 5%', 'Firefox ESR'] };
+const config = JSON.parse(fs.readFileSync('./package.json'));
+global.vfName = config.vfConfig.vfName;
```
sturobson: can you talk me through this bit later?

khawkins98: Yep. It's more about the child theme support and thinking that maintenance would be easier when we can get VF-Core and the child theme to use as close to identical config files as possible.

PR #132, opened 2019-01-07
Fix/vf tabs

### LINKED ISSUE #129: BUG: vf-pattern variants or resources tab not working
steps taken to get to this:

```
git fetch origin
git reset --hard origin/develop
npm prune
npm install
gulp dev
```

navigate to the vf-summary pattern

expected behaviour:

clicking on a variant shows that variant
clicking on a resource tab shows that file

neither work

khawkins98:
Fixes #129. 

Basically the problem was that when there was more than one tab set on a page and the first tab set wasn't JS tabs, it was throwing off indexes. 

This fix adds in data attributes:
```
<ul class="vf-tabs__list" data-vf-js-tabs>
```
```
<div class="vf-tabs-content" data-vf-js-tabs-content>
```

This means when there are non-JS tabs, data attributes can be left off and all works well.

### DIFF
--- a/components/vf-tabs/vf-tabs.hbs
+++ b/components/vf-tabs/vf-tabs.hbs
@@ -1,5 +1,5 @@
 <div class="vf-tabs">
-  <ul class="vf-tabs__list" data-vf-tbas_list>
+  <ul class="vf-tabs__list" data-vf-js-tabs>
     <li class="vf-tabs__item">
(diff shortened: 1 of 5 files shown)


khawkins98:
☝️ This is also a first proof-of-concept for using `data-vf-js-*` to indicate when our optional JS should be used, or you can leave the data attribute off and the framework will let you do your own js, even if you want to use the framework JS elsewhere.

PR #134, opened 2019-01-09
Support images wider than 180px

khawkins98:
Makes news images fit an intended size. Also updates the `vf-summary` to use its own news image (not the deprecated one).


components/vf-summary/vf-summary--news-has-image.scss
```diff
@@ -12,6 +12,7 @@
   .vf-summary__image{
     grid-column: 1 / span 1;
     grid-row: 2 / span 4;
+    max-width: 100%;
```
sturobson: Do you think we should abstract this to:

```
img {
  height: auto;
  max-width: 100%;
}
```

Or would that potentially break any consumers site that's using Bootstrap/Foundation? 

I know it'd potentially break their site if not using those frameworks.

PR #149, opened 2019-01-15
deprecates vf-contact / introduces vf-summary--profile

sturobson:
with the article / Wordpress pages I noticed that what was the 'contact' pattern had been updated.

This was a good opportunity to refactor it into a summary variant. This PR does the following:

- creates a `vf-summary--profile` pattern
- deprecates the `vf-contact` pattern 

notes: the `vf-contact` pattern name might need to be repurposed at some point for ¯\_(ツ)_/¯ something to do with contacting EMBL … so we need to have a think about that part of deprecating patterns for the future too, I guess.


khawkins98:
`vf-contact-use-this-one` :P

PR #153, opened 2019-01-16
Pattern/embl breadcrumbs tweaks

khawkins98:
Fixes a few rough edges:
- Handling null URLs
- Handle "no match found"
- Don't link to the active term


khawkins98:
🍻

Issue #158, opened 2019-01-17
PATTERN: embl-content-hub-loader


khawkins98:
We should move the Content Hub loading JS/pattern from the static pages to the EMBL VF. 

We also need to document when user should use the HTML from the pattern library vs loading from the content hub ... i.e. "Here's the `vf-footer` pattern, but don't copy this code, copy the ContentHub link instead"

PR #159, opened 2019-01-17
Load header footer from ContentHub

khawkins98:
A pre-step for in with #158 (PATTERN: embl-content-hub-loader)


khawkins98:
Now updated with an `embl-content-hub-loader` pattern:
- For #158
- Migrate JS from static pages repo
- Updates to use VF approaches including data-embl-js-content-hub-loader
- Updates templates to no longer load js from static pages repo

sturobson:
I wonder if we need to add `gulp component` generated files that aren't being used?

khawkins98:
Yeah, I don't know. 

- Part of me: delete them, they're junk.
- Other part of me: since it's primarily a CSS framework, better to leave them in as that way you know it wasn't an error .. "this page was intentionally left blank"

If you look at my non-css patterns (boilerplates) you see that I seem to change my mind often.

Feelings?

sturobson:
yep - "since it's primarily a CSS framework, better to leave them in as that way you know it wasn't an error .. "this page was intentionally left blank" - let's go with that.

khawkins98:
Added a new PR for the documentation about empty files: #164

PR #164, opened 2019-01-17
Add "This page was intentionally left blank"

khawkins98:
Following conversation in #158, this add notes about leaving empty sass files to:
- A few existing patterns
- The pattern generator

### DIFF
--- a/tools/component-generator/templates/_component.scss
+++ b/tools/component-generator/templates/_component.scss
@@ -1,5 +1,12 @@
 // <%= componentName %>
 
+// **Thinking about deleting this file?**
+// If your component needs no CSS/Sass, we still recommend leaving the
+// scss files in place. As this is primarily a CSS framework, it is better to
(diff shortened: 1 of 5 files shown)


tools/component-generator/templates/_component.scss
```diff
@@ -1,5 +1,12 @@
 // <%= componentName %>
```
sturobson: I wonder if we add this into the component generator - so that it's a thing to be deleted, rather than added?

khawkins98: I think that's what I've done here? Or did you mean that we should do something else?

sturobson:
I wonder if it would be good to add these as a defaults to the component generator too

tools/component-generator/templates/_component.scss
sturobson: Gah! My eyes!

PR #174, opened 2019-01-21
updates vf-inlay

sturobson:
This PR:

- changes the HTML element for a vf-inlay from an `<article>` to a `<section>`.
- adds a `:last-of-type` to `vf-inlay` so that it should work as per designs


sturobson:
@dbushell note: I've changed the HTML element from an article to a `<section>` now

PR #175, opened 2019-01-22
refactors vf-breadcrumbs - breaking change

sturobson:
This patter moves the lists for breadcrumbs inside a nav element with added aria where appropriate for a11y.

It also splits out the `--with-related` Sass to a separate partial and updates the styles to be more like the Sketch file.

Sketch file: 
screenshot 2019-01-22 at 09 42 44

Before:
screenshot 2019-01-22 at 09 21 34

After:
screenshot 2019-01-22 at 09 35 19


sturobson:
@khawkins98 / @dbushell note the HTML changes for this pattern

Issue #162, opened 2019-01-17
FEATURE: Make `vf-banner` more reusable


khawkins98:
The [`vf-banner'](https://dev.beta.embl.org/guidelines/visual-framework/dev-docs/components/detail/vf-banner--phase.html) pattern could be made more flexible.

`--gdpr` banner should be expanded into something more generic, perhaps `--modal`; this could have various states, attributes:
  - position (static|fixed|bottom|top)
  - state (static|dismissable|blocking)
    - if blocking, `esc` to accept,close (y|n) 
    - dismiss button text (default=OK|string|disabled)
      - cookie support (y|n)
        - cookie name (`vf-modal__mycookie`)
        - cookie version (`vf-modal__mycookie-version`)
    - extra button support(`<a href="#">string1</a>\<a href="#">string2</a>`)
  

 This could look like:
```
<div class="vf-banner vf-banner--fixed vf-banner--bottom vf-banner--modal" 
data-vf-js-banner-state="static|dismissable|blocking" data-vf-js-banner-esc-close="y|n"
data-vf-js-banner-cookie-text="Dismiss" 
data-vf-js-banner-cookie-name="mysite_gdpr"
data-vf-js-banner-cookie-version="0.1"
data-vf-js-banner-extra-button="<a href='#'>string1</a>\<a href='#'>string2</a>">
    <div class="vf-banner__content">
        // ContentHub import here, or custom HTML
    </div>
</div>
```

khawkins98:
Some related/dup tickets out there: #62 #26

PR #176, opened 2019-01-22
Pattern/banner modal variant

khawkins98:
More to do over time here, but a lot of progress and in a functional state.

- deprecate old --gdpr variant
- adds --modal
- adds cookie storage
- allows JS object to be used to make banner

For #162


sturobson:
are a modal and a banner two separate things though?

I'd say yes :)

khawkins98:
Yeah, there's a good bit of hang on work here too... especially how do we convey something that has multiple JS and CSS options ... maybe I'll have a look at some API documentation approaches.

Issue #179, opened 2019-01-23
separation of patterns by 'ownership'


sturobson:
I think it might be cleaner to move `EMBL specific` patterns to their own 'space' in the layout of patterns?

I think the best way to do this in the UI is to duplicate the layout with blocks/containers/pages/deprecated 

We cannot do this in the folder structure as it'd be a pain for lerna etc.

I think we should be able to do this with the way you have created the lists in the Fractal UI.

Next step is for me to try this - if it works - I'll raise a PR.

khawkins98:
Copying a comment over from the above PR:

> ties in with the need to document our user types/need (#127)
>
> That is:
> 1. We have EMBL site builders that won't much care if a pattern is "VF block" or "EMBL block"
> 2. We have distinctly non-EMBL users that want to avoid all EMBL specific stuff
> 
> Ideally we could do something like:
> 1. If building for EMBL (dev.beta.embl.org/../dev-docs/), then show EMBL patterns
> 2. If building generic VF (visual-framework.github.io/vf-core/), then hide EMBL patterns
> 
> But that has implications beyond just showing the patterns but also which patterns are used and > what's in `vf-core/index.scss`

As a mid-term step, thought is to use the above PR (splitting out patterns by type) for alpha.2 and then revisit a more holistic solution for alpha.3

khawkins98:
I think this has been largely addressed and it's a general thing we're working on. Being bold and closing.

Issue #127, opened 2019-01-06
DOCS - List core audiences/user types


khawkins98:
**What:**
List out our core user types

**Why:**
- Reminds us who our users are
- When someone asks "why have you done x?" or "why won't you do x?" this would be helpful
- Informs how we should build vf-core and whatever other repos

khawkins98:
Ties in with #114 (governance).

khawkins98:
We now have thoughts on audience for vf-core; from: https://github.com/visual-framework/vf-core/issues/321#issuecomment-476579608

> - Lower-level developers
>   - need to: build a design system for orgs that work with lifescience info 
>   - interface: docs, code, command line utils
> - Maintainers of design systems
>   - need to: add a vf-core pattern to an existing vf-core compatible system
>   - primary interface: npm, online docs
> - Those curious/learning/evaluating the VF
>   - need to: understand what vf-core is
>   - primary interface: the static site, github readme, examples of orgs using the VF 
> 
> What's not above: people building a site using generic VF-core CSS+JS ... I don't think we want that, instead — if there is demand — we might have a `vf-boilerplate`, but that's a TBD.

We still need to fully consider the EMBL use case, but that's after beta.1

khawkins98:
I think we've half addressed this -- the other part is an expansion on "build a design system for orgs that work with lifescience info" and the needs there. 

That is, we've listed the audiences for vf-core, but not the use cases. 

I've previously discussed that use case in https://blogs.embl.org/communications/2018/09/12/faster-scientific-websites-through-reusability/

So distilling some points:

- Reusable, structured CSS, JS
- A pattern library
- Sane defaults for new websites
- Make a "look and feel" more consistent
- Provide versioned upgrades 
- Run with and alongside other major tools (Foundation, Bootstrap and from the JS side Angular, React)
- Easily add your own patterns, tweak existing patterns
- Use it without a fixed look (doesn't have to look like EMBL)

Need to stash those in the docs somewhere, too,

PR #180, opened 2019-01-23
refactors component list UI to separate EMBL specific patterns

sturobson:
This PR:

- updates the nunjucks files to separate the EMBL patterns from VF Core patterns
- prefixes existing `pattern-type` with `embl-` so they get picked up

### DIFF
--- a/components/embl-boilerplate-page/embl-boilerplate-page.config.yml
+++ b/components/embl-boilerplate-page/embl-boilerplate-page.config.yml
@@ -2,4 +2,4 @@ title: EMBL Boilerplate page
 label: EMBL Boilerplate
 preview: '@preview--nogrid'
 context:
-  pattern-type: boilerplate
+  pattern-type: embl-boilerplate
--- a/components/embl-breadcrumbs-lookup/embl-breadcrumbs-lookup.config.yml
+++ b/components/embl-breadcrumbs-lookup/embl-breadcrumbs-lookup.config.yml
@@ -3,7 +3,7 @@ label: EMBL Breadcrumbs
 preview: '@preview--blocks'
 status: alpha
 context:
-  pattern-type: block
+  pattern-type: embl-block
   meta-who: Strategy and Communications
   meta-what: Visual Framework
   meta-where: EMBL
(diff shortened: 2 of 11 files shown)


khawkins98:
I think this ties in with the need to document our user types/need (#127)

That is:
1. We have EMBL site builders that won't much care if a pattern is "VF block" or "EMBL block"
2. We have distinctly non-EMBL users that want to avoid all EMBL specific stuff

This PR I think is a real benefit to type 2, but less so EMBL users.

That said, I don't think we can really do much better at this point.

Ideally we could do something like:
1. If building for EMBL (dev.beta.embl.org/../dev-docs/), then show EMBL patterns
2. If building generic VF (visual-framework.github.io/vf-core/), then hide EMBL patterns

But that has implications beyond just showing the patterns but also which patterns are used and what's in `vf-core/index.scss`

Suggest we go with this for now and revisit in #179 for alpha.3

khawkins98:
Ah, and one other thought: Will get further complicated as we add EMBL-EBI and EMBL Etc. specific patterns!

sturobson:
I don't think hiding things would be a way forward - as that might stop, for example, a uniprot specific pattern being adopted by another team and becoming a vf pattern (which we need to work out the process for). 

Rather than separate them as is, perhaps have it as … 

screenshot 2019-01-23 at 11 36 08

sturobson:
from that image - we'd just need to have some way to tighten up that whitespace a little.

sturobson:
tidied this up with the divider lines …

screenshot 2019-01-23 at 11 49 11

khawkins98:
One of the questions about adopting contrib patterns: would `uniprot-new-pattern` remain that, or become `vf-new-pattern`?  If we keep the `uniprot-` namespace it could slow down development of any uniprot specific features.

Also suspect that as `dev.beta.embl.org/../dev-docs/` matures, we'll want to remove some patterns (no need to tell EMBL users about the generic `vf-logo`)

khawkins98:
On the updated image: might be able to get mostly there by sorting the output by prefix.

Pseudo code:
```
for each pattern:
  if `vf-*`: <div>show pattern</div>
for each pattern:
  if `embl-*`: <div>show pattern</div>
for each pattern:
  if `ebi-*`: <div>show pattern</div>

PR #192, opened 2019-01-27
Include varaibale.scss in package.json

khawkins98:
And fix a few omissions in patterns.

The omission of variables.scss in npm has broken child themes a bit, suggest we consider a hot alpha.3 or at least a push to npm


khawkins98:
@sturobson  actually, with the way we’ve discussed versioning, I think we could:

1. Push to master
2. Push to npm

And we don't need to do a `vf-core` alpha.3 release as while the vf-core component template is changing, this is mostly about getting the patterns on npm fixed.

PR #195, opened 2019-01-29
Chore/update packages on npm

khawkins98:
Merge in newest version numbers, githead to correspond to npm


khawkins98:
As we're not `tagging` lerna edits to the git repo, it can't track when things have changed, so we're having to bump everything every time.

We should figure out if we can re-enable tags, but just get them to not show as releases.

PR #198, opened 2019-02-01
Only use half page grid gap for nav

khawkins98:
As we use background colours on hover, we need the padding-left on first child, so we use half the `--page-grid-gap` to fix alignment, see screenshots.

Does some other cleanup as well.

After:
image

Before:
image

Docs: http://localhost:3002/components/detail/vf-navigation--primary
Target use case: http://dev-vf-theme-prototype.pantheonsite.io/


sturobson:
hmm, I feel uneasy with using a 'page grid gap' for a patterns grid gap :sob:

Especially if we change page-grid-gap in the future and that makes this look out of place again.

Perhaps this is enough for me to start working on that 'local design tokens' thing - but there are other pressing matters I'm sure.

Let's catch up with this next week?

khawkins98:
Yeah, also ran through my head. Not sure if the grid fall is universally tired to menu item spacing? But happy with however we tweak it.

sturobson:
circling back to the initial use case for adding the spacing -- this would break the (currently defunct) masthead design. Also the spacing needs to be more akin to the rest of the vf-inlay spacing to work a bit better too.

Working on this still

sturobson:
Ok.

The spacing and potentially the grid, and other things are dependant on the parent quite a bit for this pattern.

So do we do this

```
/* A */
.vf-parent-class > .vf-navigation {
  /* code here */
}
```

or 

```
/* B */
.vf-navigation--parent-class {
  /* code here */
}
```

I think A is how we're doing most of it now. But where would we put that code in `parent-class.scss` or `vf-navigation.scss`? 

Using `.vf-navigation--parent-class` means we can put it either in `vf-navigation.scss` or it's own `vf-navigation--parent-class.scss` file.

I'm unsure which would be better for maintenance. Perhaps we should also include 'relies on this Sass file' in the comments for something like this moving forward too

khawkins98:
Think I'd agree that `A` is the better/consistent approach. 

As for where the styling lives — we're facing that idea of interlinkage elsewhere, too — I think in `vf-navigation.scss`, as it is vf-navigation that is supporting/enhancing vf-parent, and vf-parent should work regardless if vf-navigation is installed. 

Avoids a situation where to install, say, `vf-inlay` you also need to install `vf-navigation`, and as `vf-navigation` only includes a class reference to `.vf-inlay` you don't _have_ to install `vf-inlay`.

---

Last minute bonus thought: don't know if it's worth making a new `.scss` file convention for patterns of either:
- `vf-navigation--enhancements.scss` and/or
- `vf-navigation--enhancements__vf-inlay.scss`

sturobson:
how about 

`vf-navigation--global` to match the global header
`vf-navigation--main` for 'normal' navigation
`vf-navigation--additional` for the 'sub brand' navigation

PR #199, opened 2019-02-05
Update/embl group page

khawkins98:
Following the (forthcoming) navigation updates, realised our `embl-group-page` template was out of date. 

This is a first cut at improving it. 

(Note the EMBL global header is missing as the Content Hub is still offline)

image

The biggest questions are:

1. How should we set the background colour? 
    - `<div class="vf-body vf-u-background-color-gray-bright">`  feels like a bit of a hack? `vf-body__embl-group`?
1. How do we control the in/out of the background colour as you scroll down?
    - For this example I've done it like the below, but that may not be a good or easy approach for things like WP (note how David is currently manually escaping the inlay with `.vfwp-column-reset {  grid-column: 1 / -1; }`)

```html
  <div class="vf-body vf-u-background-color-gray-bright">
    <section class="vf-inlay vf-u-background-color-green">
      <header />
    </section>
    <section class="embl-grid embl-grid--has-centered-content vf-u-background-color-white">
        {{> '@vf-content'}}
    </section>
  </div>
  <div class="vf-body"> <!-- second vf-body 😒 to end the coloured background, a `vf-body__embl-group` might help this -->
    <section class="vf-news-container | embl-grid embl-grid--has-sidebar">
```


sturobson:
1. How should we set the background colour?
  We need something generic I guess - as would `vf-body__embl-group` be a little redundant where this could be used else where?
  `vf-body--bg-alt`?

2\. How do we control the in/out of the background colour as you scroll down?

I 'think' once it goes fully white it stays that way.
David's turned the `vf-inlay` into a separate 'container` effectively for each vertical strip of content I think

I think we need to work out how to employ `grid-column` a little better. As we can ID parts of the grid `[main-start] [main-end]` etc - I'm wonder if we can set 'most things' to `grid-column: main` which would fit a) within `vf-inlay` and b) `vf-body`.

As there's no sub grid there's no conflict (I'm not sure if there would be with sub grid) - so there's no clash of which `main` to use.

sturobson:
I've merged for now - we can circle back to this - hopefully as part of the next sprint

PR #203, opened 2019-02-06
pattern/vf-header-nav-site_additional

khawkins98:
Looks like a couple of the name changes in #198


sturobson:
I find it odd that GH allows the OP to merge a branch after additional changes made post 'review'.

All's good though

PR #204, opened 2019-02-06
initial pass at creating a pagination component

sturobson:


sturobson:
Cool! 👍, This is just a quick thing to get the markup in place more than the 'design' and any potential other pattern variants

Tweak away!

PR #209, opened 2019-02-07
Features/vf colors

sturobson:
makes a start on creating `vf-core-colors` for non-changeable UI things


khawkins98:
and maybe `vf-colors.yml` should be `vf-generic-colors.yml`?

sturobson:
are vf-colors generic though? I thought they were more tied to EMBL?

sturobson:
I've updated `vf-core-colors` to `vf-ui-colors` in the interim

khawkins98:
I think the `vf-core-colors` are a bit mixed between generic/embl at the moment ... `vf-color--gray-bright` is fairly generic to me, it's more the hex colours we're passing in are EMBL specific.

I still don't have a good feeling as to how we'll abstract out a generic colour palette ... but I think this is a good step.

PR #211, opened 2019-02-07
Update vf-form demo

khawkins98:
The structure of the vf-form pattern doesn't show the sub-pattern folders (not sure if fractal thing or how I've made a fork of the fractal theme), but for now this shows all the vf-form varianats on the main vf-form page.


sturobson:
Yeah - Fractal doesn't do this :( 

I think they're too 'small' to be individual patterns with their own folder though. 

Perhaps they could be refactored as variants

PR #214, opened 2019-02-07
Update vf-inlay.scss

khawkins98:
Gets the width of the sidebar to not default to wide, sets a min-width closer to Mark's designs.

Relates to #208

### DIFF
--- a/components/vf-inlay/vf-inlay.scss
+++ b/components/vf-inlay/vf-inlay.scss
@@ -73,7 +73,7 @@
     minmax(var(--page-grid-gap), 60px)
     [main-start]
       minmax(auto, 100%)
-      minmax(18.125em, auto)
+      minmax(auto, 22.125em)
     [main-end]
     minmax(var(--page-grid-gap), 60px)
   ;


components/vf-inlay/vf-inlay.scss
```diff
@@ -73,7 +73,7 @@
     minmax(var(--page-grid-gap), 60px)
     [main-start]
       minmax(auto, 100%)
-      minmax(18.125em, auto)
+      minmax(auto, 22.125em)
```
sturobson: why is auto and the magic number around the other way here? As flipping it fixed this mornings issue?

khawkins98: Don't think the earlier tweak sorted it: http://dev-vf-theme-prototype.pantheonsite.io

khawkins98: Actually, it looks like that line somehow became flipped back

https://github.com/visual-framework/vf-core/pull/208/files#diff-5b4dd4e538d4c78bf7ece45164ef9b05R44

sturobson: ah, ok - I'll merge

khawkins98: I'm not sure why it's currently flipped on `develop`:

- develop has it wrong: https://github.com/visual-framework/vf-core/blob/develop/components/vf-inlay/vf-inlay.scss#L76
- but I don't see why in git blame: https://github.com/visual-framework/vf-core/blame/5b4c84c2777fd0e0bf33a61bd9307f0df8b6219d/components/vf-inlay/vf-inlay.scss#L44

Issue #218, opened 2019-02-08
CHORE - Precommit hooks


khawkins98:
Would, of course, be good to do some precommit hooks to integrate the existing lint warnings/errors.

Could try [Husky](https://github.com/typicode/husky)

sturobson:
Husky looks much easier to use than the pre-commit node module I found. I'll take this one

PR #219, opened 2019-02-08
adds husky and scss-lint as a pre-commit.

sturobson:
notes:

After initial setup I had put `"precommit": "gulp scss-lint"` in the `package.json` `scripts` section but Husky suggested moving it - which I did 'automagically' with `./node_modules/.bin/husky-upgrade`as the messing in the terminal suggested. 

You may need to manually do the above for this to work - as I've not committed any files relative to `./node_module` in this PR as I'm not convinced it would be a good precedent to start doing that.

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -1,0 +1,56 @@
+{
+  "plugins": [
+    "stylelint-order",
+    "stylelint-scss"
+  ],
+  "extends": [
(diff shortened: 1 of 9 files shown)


sturobson:
This will close #218

sturobson:
Awesome stuff!

PR #221, opened 2019-02-08
Pattern/embl group page background

khawkins98:
Does a bunch of little things to:

- get the EMBL group pages looking more like Mark's design
- suggests setting the background on the parent container (so the background also goes behind the breadcrumb)
  - adds `.embl-group-page--bg-color` `.embl-group-page--vf-header--bg-color` (see comment at https://github.com/visual-framework/vf-core/commit/0fd03a67313af09df999eb38f75cbd5eb910fd10)
- the default `vf-inlay.hbs` is more generic and less like the `embl-group-page.hbs`
  - adds `.vf-inlay__content--full-width`
- fix some minor issues, linting

Screenshot of Mark's design, after, before:

image

### DIFF
--- a/components/embl-boilerplate-page/embl-boilerplate-page.hbs
+++ b/components/embl-boilerplate-page/embl-boilerplate-page.hbs
@@ -8,7 +8,7 @@
   <meta content="width=device-width, initial-scale=1, minimum-scale=1" name="viewport">
 
   {{> @vf-favicon}}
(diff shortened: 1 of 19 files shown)


sturobson:
On the Sketch files the phase banner is between the breadcrumbs and the inlay content - does this allow for that? Looking at the visuals the banner is under the inlay masthead

Issue #213, opened 2019-02-07
ENHANCEMENT - "This page intentionally left blank. See the pattern variants."


sturobson:
Before Christmas I was toying w/ creating `vf-pattern__example.hbs` which would be a more 'designed' version of the `collated: true` option.

As it defaults to … the default file we could either 

a) use `collated: true` and just put everything in one 'file' in the UI
b) create a separate `vf-pattern__example.hbs` file 
c) see if we can hide the `default` if empty
d) get rid of the component iframe(?) views and make the `vf-component.md` file the page.

khawkins98:
Adding this one to the backlog.

PR #225, opened 2019-02-11
Feature/update collated previews

khawkins98:
Addresses ideas in #213 and #6, including:

- hiding tabs for collated patterns
- forcing all blocks to have a collated appearance
- not using iframes for collated/blocks
- adding anchor links (we could use a pattern around those)
- also sneaks in a few things found around links in badges

After (left) / before (right):

image


sturobson:
This is looking good - I wonder if we'd be better placed to, eventually, have the relevant HTML underneath (this could then be all done in the `README.md` file which we could then pre-fill a little more with the pattern generator.

sturobson:
Let's merge this in now as it's a vast improvement on what we have there already. 

I'll open an issue about making the component page be mainly the .md file as well - I did have a branch before xmas - but it looks like I deleted it :(

PR #226, opened 2019-02-11
Further improve/simplify blocks/element preview

khawkins98:
Follows on #225 by:

1. Dropping `collated`
2. Rendering all `elements` `blocks` as "collated" with inline code snippets

Ties in with #213, #6 

image


sturobson:
screenshot 2019-02-11 at 12 32 46

makes the components files and resource section full width

PR #227, opened 2019-02-11
makes bottom margin of vf-inlay 100px

sturobson:
currently the space between vf-inlay__content and anything else is just 38px (no idea why that number) - the change in content grid looks awful and looking at the sketch files the spacing is 100px

This PR just updates that.


khawkins98:
Should we use `  @include margin--block(bottom, 100px);` here? 

I'm not really sure when we should/shouldn't use the `_margin.scss` mixins. Maybe that's a to-do follow-on doc task.

sturobson:
Ah, you're right I think. 

When we/I/you get around to implementing the per-pattern design tokens idea that was discussed in December - the use of the mixin would make much more sense (I hope).

PR #228, opened 2019-02-11
updates vf-form__select to match latest designs

sturobson:
this _should_ make the `select` look like the group pages sidebar `select` design.

I have amended the CSS from this codepen - https://codepen.io/chriscoyier/pen/RELVWj


khawkins98:
I don't think this is coming over quite right for me at least in FF/Chrome:

Chrome:

image

FF:

image

---

Slightly off topic, because of the way Fractal has a hard time with the sub-patterns in `vf-form`, I hard a hard time previewing this. I think I've found a way to address it with reasonable trade offs. I'll push that to a new branch.

PR #229, opened 2019-02-11
Fix/vf form subpatterns

khawkins98:
Builds off #228 to make the subpatterns work in fractal and allows them to be seperate npm packages, does two things:

1. changes the package suffix from `__select` to `--select` (the underscores freak fractal out)
2. moves the `__select.hbs` up a folder to `vf-form`, which means they'll not be in the npm install but, that doesn't feel horrible given you're likely to install `vf-form` + `vf-form--select`


sturobson:
I think we, perhaps, need to rethink form elements.

Are they one big thing / one single npm install, or are they all separate npm installs?

I think deciding that will allow us to flatten things a little and avoid us jumping through hoops for Fractal

PR #237, opened 2019-02-15
updates button link mixin to include borders and text color

sturobson:
this PR

- adds border colours for visited and hover on buttons mixin


components/vf-sass-config/mixins/_links.scss
```diff
@@ -25,6 +25,7 @@
 
   &:visited {
     background-color: $vf-link-visited-color;
+    border-color: $vf-link-visited-color;
```
khawkins98: For these, should the dev also be able to pass `$vf-link-visited-color` as a param to the mixin?

PR #238, opened 2019-02-18
adds comments pattern

sturobson:
This is a first pass at creating the relevant HTML required for a comments section as part of a blog.

This is based on, loosely, the design in sketch (but this could be out-of-date? @markboulton)

The HTML shouldn't change much so I feel this is 'good enough' for @dbushell to take from for the WP theme.


components/vf-comments/vf-comments.hbs
```diff
@@ -14,7 +14,7 @@
         <p>For me, any communications strategy should be deeply rooted in the organisational strategy. And yet, your closing question reminded me how the discussions we lead about communications can often result in much broader discussions on organisational goals, ambitions and approaches.</p>
       </div>
       <button class="vf-button vf-button--outline--primary | vf-comments__reply">Reply</button>
-      <a href="#" class="vf-button vf-button--outline--tertiary | vf-comments__reply">Optional action</a>
+      <a href="#" class="vf-button vf-button--outline--tertiary | vf-comments__action">Optional action</a>
```
sturobson: do you think this would be an action that takes you to a different page? 

I guess _really_ the reply button would be going to another page on most things as most things won't be a SPA ?

khawkins98: Yeah. Think in reality most would be new pages... Thinking things like: "edit" "flag as spam" ... I suppose we might eventually need more designs on those, but this felt like a reasonable first step/thought.

sturobson:
> And at the risk of being pedantic, should we call this `vf-disucssion` or `vf-article-comments` (to complement `vf-article-meta-information`)? Developer brain expected something different when seeing `vf-comments`.

I'm not overly possessive over `vf-comments` -- will there be chance that a 'comments system' would be used for more than 'articles'? 

Perhaps `vf-discussion` might be a better fit.

sturobson:
changed all the things so it's now `vf-discussion` as that might be more portable across platforms, needs and requirements.

PR #239, opened 2019-02-18
Add team lead to group page boilerplate

khawkins98:
Updates boilerplate to add team leader to top right, fixes styling of intro text.

image

### DIFF
--- a/components/embl-group-page/embl-group-page.hbs
+++ b/components/embl-group-page/embl-group-page.hbs
@@ -73,11 +73,11 @@
       <section class="vf-inlay__content vf-u-background-color-white">
 
         <main class="vf-inlay__content--main">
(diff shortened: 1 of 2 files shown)


sturobson:
neat-o.

aside: I wonder if the component library is the 'correct' place to put example pages?

I guess it is fine for now - but perhaps the these could be statically hosted pages somewhere else as part of a much larger 'documentation' site (or design system).

Having them inside the components feels a little icky and the `detail` page is a bit cumbersome having a long page inside a page. imho.

khawkins98:
Yeah, I'm a bit between lesser evils:

- cons: what you said
- pro: avoids "looking for a container/block? go here. looking for for a boilerplate? go here." often devs need both

As a bonus, we also can do `{{> '@vf-pattern'}}`

I like the idea of abstracting/pulling all the bits into some sort of design system site, might also help clean up the repo structure a bit... one to mull on.

PR #240, opened 2019-02-18
updates button design to include darker bottom 'border'

sturobson:
as per the buttons in the sketch file…

I've added box-shadow to the buttons so there's a 'pressed' effect when hovering along with a dark bottom border - so that it looks a little '3d'

### DIFF
--- a/components/vf-button/vf-button.scss
+++ b/components/vf-button/vf-button.scss
@@ -10,37 +10,39 @@
   margin: 0;
   padding: 16px 24px;
   text-decoration: none;
-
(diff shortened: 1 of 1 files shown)


components/vf-button/vf-button.scss
```diff
@@ -4,27 +4,51 @@
   @include set-type(button--r);
 
   appearance: none;
-  border: 0;
+  border-radius: 4px;
```
khawkins98: Not sure if we still need a `border: 0` before to reset any default `<button>` faff in random browser X, but I guess _if_ that's an issue we can sort later.

Issue #230, opened 2019-02-12
Because of pulling in #228 - the vf-form--select pattern needs fixing.


sturobson:
From @khawkins98's comment in #228 

> I don't think this is coming over quite right for me at least in FF/Chrome:
> 
> Chrome:
> 
> ![image](https://user-images.githubusercontent.com/928100/52585152-502fd580-2e34-11e9-97c1-1579e89e859a.png)
> 
> FF:
> 
> ![image](https://user-images.githubusercontent.com/928100/52585157-5625b680-2e34-11e9-957b-ddffd20bd2fc.png)
> 
> Slightly off topic, because of the way Fractal has a hard time with the sub-patterns in `vf-form`, I hard a hard time previewing this. I think I've found a way to address it with reasonable trade offs. I'll push that to a new branch.

sturobson:
the issue with browser ui and css conflicting is now resolved … 

but the pattern needs styling to better fit the rest of the form patterns

PR #241, opened 2019-02-18
updates select to match other inputs and their interactions

sturobson:
the initial `vf-select` refactor didn't really carry the theme of the rest of the `vf-form` patterns.

This PR updates it.


khawkins98:
Sorts #230

PR #242, opened 2019-02-18
adds simple search form

sturobson:
in the group pages blog design the side bar as an 'inline input and button'.

This pattern is for that.

The search input makes use of `flex-grow: 1;` which means the input will fill any space allowed - leaving the button to take up just enough space it needs.


khawkins98:
Seems like we might have multiple search variants:

- this one
- "Advanced"
- multi-step

Should it instead be `vf-search` || `vf-search__simple`?

Or maybe `vf-form__search-simple`?

sturobson:
I've gone with `vf-search and `vf-search--inline` here in the end. 

Should be good to go

PR #251, opened 2019-02-21
resizes a few form elements as they're a bit bigger than the sketch files

sturobson:
@dbushell noticed that the search button looks a little big. 

I took a look at the latest sketch files and by moving towards the Plex sans typeface we didn't update the sizes for the form inputs and buttons etc.

This changes that so they match the sketch files a little more.


khawkins98:
The buttons are also looking a bit big: https://dev.beta.embl.org/guidelines/visual-framework/dev-docs/components/detail/vf-banner.html

image

sturobson:
something's squiffy on that bottom banner? too. 

They should be smaller than that now

Issue #197, opened 2019-01-31
IDEA: Block use of hex, named colours


khawkins98:
Via https://github.com/visual-framework/vf-core/pull/196#pullrequestreview-198479803 

@sturobson :

> We should look to add the use of hex, or named colours to be a blocker in the linting - which in turn should disallow any pushes or merges or PRs until fixed.

PR #253, opened 2019-02-21
Feature/block use of hex and named colours

khawkins98:
- Address #197 and lints for use of hex or named colours.
- Converts named/hex colours to variables/tokens
- Cleans up a few leftover/unused variables

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -8,6 +8,8 @@
   ],
   "rules": {
     "string-quotes": "single",
+    "color-named": "never",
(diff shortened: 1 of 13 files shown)


khawkins98:
In `components/vf-code-example/vf-code-example.scss` I've opted to convert the colours to the VF colours, rather than disable liniting; as:

1. Our VF colours matched fairly closely
1. I guess it makes the "theme" be more consistent
1. It feels nice to not have those hex colours hanging around

Issue #258, opened 2019-02-22
changing the markup for `vf-masthead__heading`


sturobson:
I was thinking about this over coffee …

In the group pages we have, inside the `vf-inlay`, a `vf-masthead` with a heading, and now a subheading.

The subheading is a description of the group.

I'm wondering if it's a better fit as a `<span>` or similar of `vf-masthead__heading`.

Currently the markup is:

```
<div class="vf-masthead__title">
  <h1 class="vf-masthead__heading">Häring Group</h1>
  <p class="vf-masthead__sub-heading">Chromosome structure and dynamics</p>
</div>
```

I think we could change that to:

```
<div class="vf-masthead__title">
  <h1 class="vf-masthead__heading">Häring Group<span class="vf-sr-only">:</span> <span class="vf-masthead__heading--additional">Chromosome structure and dynamics</span></h1>
</div>
```

This way, a screen reader and a search crawler would read it in one parse. It would also benefit searches (not sure if applicable) as the description is part of the H1 and tells the story a little more.

If there's no objections to this - I'll make this change today.

sturobson:
This is a potential breaking change for: 

the `vf-wp-theme` and/or the content hub

khawkins98:
Conceptually I really like this.

PR #259, opened 2019-02-22
Updates/vf masthead 

sturobson:
ref: #258 

This component update

- reduces the HTML
- makes the group or team description part of the `H1`

This update does not

- check if `with-title-block` works correctly (I've also not seen this in recent designs)

### DIFF
--- a/components/embl-group-page/embl-group-page.hbs
+++ b/components/embl-group-page/embl-group-page.hbs
@@ -52,14 +52,7 @@
 
     <header class="embl-group-header__header embl-group-page--vf-header--bg-color">
 
-      <div class="vf-masthead">
-        <div class="vf-masthead__inner">
-          <div class="vf-masthead__title">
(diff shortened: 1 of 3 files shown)


sturobson:
Cool … We need to think about how we're going to implement theming a little? 

`<div class="vf-masthead" style="background-color: green;">` 

Currently I think we can go with CSS custom properties for it. With a fallback for oldIE of white on black or black on white or green headers and progressively enhance them with custom properties for other 'things'.

This worked well for the last couple of specimens I worked on at Monotype.

sturobson:
this one - kinda gets the subheading works within EMBL-group-pages - but breaks the masthead for the generic vf-inlay (which is where I'd like to move 'embl-group-pages__header' in the next release.

Unsure if you want to pull this in - or just make refuse and close - as it's a bit of a mixed bag

sturobson:
after all the pain, I'm burning this PR and looking to start a fresh for the next release.

PR #271, opened 2019-02-28
Add more missing files in pattern package.json files

khawkins98:


sturobson:
I wonder if there's any pre-check we can do for this ?

if $files (from folder) exist in package.json then a-ok or something. It's probably something that we (I) need to be vigilant about and do by hand though

PR #276, opened 2019-03-04
Feature/update embl ontology client

khawkins98:
This revises the `embl-breadcrumbs-lookup` pattern for [the ontology approach](https://gitlab.ebi.ac.uk/emblorg/backlog/issues/124).

Also included is that it looks up the parent-parents  of the active term ... there's more work to be done there.

### DIFF
--- a/components/embl-breadcrumbs-lookup/embl-breadcrumbs-lookup.hbs
+++ b/components/embl-breadcrumbs-lookup/embl-breadcrumbs-lookup.hbs
@@ -1,4 +1,4 @@
-{{> @embl-content-meta-properties }}
+{{> @embl-content-meta-properties meta-who="James Sharpe" meta-what="Sharpe Group" meta-where="EMBL Barcelona"}}
 
 <nav class="vf-breadcrumbs embl-breadcrumbs-lookup" aria-label="Breadcrumb" data-embl-js-breadcrumbs-lookup>
   <div class="vf-text embl-breadcrumbs-lookup--loading-text"></div>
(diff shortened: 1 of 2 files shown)


sturobson:
is there any strong reasoning why we've gone with `embl-breadcrumbs-lookup` rather than `vf-breadcrumbs--lookup`? 

Also as `-lookup` feels like it's a modifier, do you think there would be `embl-breadcrumbs` so that we would need the `--` for all the BEM'y stuff.

khawkins98:
In this case it's specifically for EMBL's approach/spec `var url = 'https://dev.beta.embl.org/api/v1/pattern.json?pattern=embl-ontology&source=contenthub';`

I can't really imagine a possibility where another org would likely have the same taxonomy/ontology. 

As for the naming ... I'm not really opposed to `embl-breadcrumbs--lookup`, but I think that would break Fractal. Also it's complementary/requires `vf-breadcrumbs`, so we can't do `embl-breadcrumbs` ...

I wonder if it's a modifier even? It's really more of an API call that builds on top of `vf-breadcrumbs`, I don't think I have a good naming approach for that.

khawkins98:
Maybe it's `vf-breadcrumbs--embl-lookup`, but that feels a bit 🤢 , and if it had to inside of `vf-breadcrumbs`, then anyone installing it would get EMBL-specific code.

sturobson:
with all, that. Let's keep it 'as is' :)

khawkins98:
🙃

PR #281, opened 2019-03-08
pattern/support-vf-link-in-vf-masthead

khawkins98:
This does two useful things:

- Drops styling on an `a` element in the CSS
- Supports usage of `vf-masthead__heading__link`

In practice that means it resolves common styling specificity issues when dealing with `a:visited` in other CSS files (like I'm dealing with when trying to integrate with the EBI VF 1.3 CSS).


khawkins98:
I've gone with `.vf-masthead__heading__link` which feels weird, but I think is right.

sturobson:
There's some overall HTML architecture of this component that I want to look at - including using the inlined `--additional` which I got mixed up on in the last sprint.

I'll merge this one - but the HTML is _likely to change_

PR #283, opened 2019-03-08
enhancement: Block border bottom colours on links

khawkins98:
A common clash with external css will be `a {border-bottom: 1px dotted; }` we can prevent this on on many links with `vf-*` classes by adding `border-bottom: 0` to the inline-link mixin.

It  seems a reasonable assumption?

### DIFF
--- a/components/vf-sass-config/mixins/_links.scss
+++ b/components/vf-sass-config/mixins/_links.scss
@@ -2,23 +2,39 @@
   $vf-link-color: $vf-link-color,
   $vf-link-hover-color: $vf-link-hover-color,
   $vf-link-visited-color: $vf-link-visited-color,
-  $vf-link-disabled-color: $vf-link-disabled-color) {
+  $vf-link-disabled-color: $vf-link-disabled-color,
+  $vf-include-normalisations: $vf-include-normalisations) {
+
+  /// @todo Add documentation about how/when/where to expect, use $vf-include-normalisations
+  @if $vf-include-normalisations == true {
+    border-bottom: none;
+  }
 
   color: $vf-link-color;
 
-  border-bottom: none;
-
   &:visited {
     color: $vf-link-visited-color;
+
+    @if $vf-include-normalisations == true {
+      border-bottom: none;
+    }
   }
 
   &:hover {
     color: $vf-link-hover-color;
+
+    @if $vf-include-normalisations == true {
+      border-bottom: none;
+    }
   }
 
   &[disabled] {
     color: $vf-link-disabled-color;
     cursor: not-allowed;
+
+    @if $vf-include-normalisations == true {
+      border-bottom: none;
+    }
   }
 }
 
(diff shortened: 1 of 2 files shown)


sturobson:
do you think we'd need some sort of group of CSS to 'fix' existing CSS decisions - rather than add them inline? I think a last-on-the-list of imports with a collection of 'fixes' would be an easier to deprecate way of doing things?

khawkins98:
How do you mean, like an `@import-css-fixes`?

khawkins98:
Would that be part of `.vf-link`  or a standalone pattern?

sturobson:
Yes, although `fixes` is probably not the right term 'normalise'? 

I think if we add them inline into the component they could get lost in the future when they're no longer needed. 

`vf-framework-normalise.scss` could be something that's periodically checked through with ease

I reckon -- thoughts?

sturobson:
perhaps a Sass `if/else` so it can be turned on or off?

If we do that - we can have it as part of the component and then have the documentation for it with that component too

khawkins98:
What about:

```css
@mixin inline-link(
  ...
  if ($vf-include-normalisations) {
    border-bottom: none;
  }
```

Or is that too spagheti code of an approach?

sturobson:
I think if we do that it would need to be at the pattern level rather than from the mixin API.

It would be less of a black box to have 

```
a {
  @include inline-links;

  @if $vf-include-normalisations == true {
    border-bottom: none;
  }
}
```

From this it would give:

- an on/off switch for all components that need normalising.
- code that's an explicit decision as it's written 'in the component' rather than tucked away in a mixin, added by accident.

khawkins98:
I like the spirit of it, but in practice we'd have to add that snippet to 17 patterns, it also makes it tricky for nested properties ... what if we:

```scss
@mixin inline-link(
  $vf-link-color: $vf-link-color,
  $vf-link-hover-color: $vf-link-hover-color,
  $vf-link-visited-color: $vf-link-visited-color,
  $vf-link-disabled-color: $vf-link-disabled-color,
  $vf-include-normalisations: $vf-include-normalisations) {

  @if $vf-include-normalisations == true {
    border-bottom: none;
  }

  color: $vf-link-color;

  &:visited {
    color: $vf-link-visited-color;

    @if $vf-include-normalisations == true {
      border-bottom: none;
    }
  }

  &:hover {
    color: $vf-link-hover-color;

    @if $vf-include-normalisations == true {
      border-bottom: none;
    }
  }

  &[disabled] {
    color: $vf-link-disabled-color;
    cursor: not-allowed;

    @if $vf-include-normalisations == true {
      border-bottom: none;
    }
  }
}
```

sturobson:
👍 I guess we need to document it somewhere -- so that it's 'known' -- SassDoc etc etc etc

khawkins98:
Ok, more tweaks added. I think it's ok for the moment, what ya think?

PR #282, opened 2019-03-08
adds vf-icon--avatar.svg and updates vf-summary--profile

sturobson:

PR #285, opened 2019-03-08
Refactor/vf summary  profile tweaks

khawkins98:
Some observations off of @sturobson PR #282, not all of these ideas may be correct:

- Add colours to SVG  …
    - Probably more trouble than its worth to try and style an `img src=".svg"`?
- Add a grid gap  …
    - Wasn't noticeable before with the white background
- Add `.vf-summary--profile__linked`  …
    - Allows us to use the mouse hover state so we know we're hovering on something clickable
- Use vf-summary__link instead of assuming __email will always be linked  …
    - Should allow for future flexibility
- Move --overlay to top of --profile  …
    - I think this feels more natural on tabbed navigation ... focuses on full profile and then contact
- Sneak in a fix for "ORCID ID"

And here's a gif to save some time:

summary

It still feels a little weird that there's no visual indication that there's a link beyond the "Contact" link, but I guess that's ok?


components/vf-summary/vf-summary--profile.scss
```diff
@@ -1,5 +1,5 @@
 .vf-summary--profile {
-  grid-column-gap: 0;
+  grid-column-gap: 16px;
```
sturobson: subject to change (as it was `0` when we were using a `.png`)

sturobson:
> Sneak in a fix for "ORCID ID"

There's some outstanding questions on this still

sturobson:
> Add .vf-summary--profile__linked

Do we want to swing this one by @markboulton, from my pov, it's … odd

I'd expect the name to have the link, but not _everything_ … it all goes a bit too blue.

Bluer than blue

khawkins98:
> > Sneak in a fix for "ORCID ID"
> 
> There's some outstanding questions on this still

How so?

khawkins98:
> > Add .vf-summary--profile__linked
> 
> Do we want to swing this one by @markboulton, from my pov, it's … odd
> 
> I'd expect the name to have the link, but not _everything_ … it all goes a bit too blue.
> 
> Bluer than blue

Yeah. It's not for elegant, maybe just the name guys blue, but we definitely need to indicate interactivity more than a cursor

sturobson:
re: ORCID

ORCIDs can be links - https://gitlab.ebi.ac.uk/emblorg/backlog/issues/131
updated HTML idea - https://gitlab.ebi.ac.uk/emblorg/backlog/issues/130#note_36168

sturobson:
> Yeah. It's not for elegant, maybe just the name guys blue, but we definitely need to indicate interactivity more than a cursor

This is how similar implementations I've seen do it.

khawkins98:
Thinking on this more: it feels weird to present things that look like links "orcid, phone" and can't be clicked on. 

I'm certain the UX will be frustrating if you're certain you clicked an orcid but go to a profile page. 

Or, if the intent is only to drive users to the person's profile page, why are we even showing the orcid at this level?

We could lower the pseudo information density and resolve some deeper issues. 

I also wonder why we don't just show the email? Maybe that's an EMBL vs EBI behaviour.

Maybe we can pick it up in person next week.

sturobson:
> Maybe we can pick it up in person next week.

Definitely one that can be worked through when we're all in the same room as there's conflicts and compromises on all of the options for what we can do

PR #286, opened 2019-03-09
normalise: underline links

khawkins98:
This follows up on #283, and normalises the `@mixin inline-link` to underline links.

Background:

1. We're relying on normalise+default browser behaviour to underline links (that is, we never explicitily say `a { text-decoration: underline}`
2. Many other frameworks reset `a, a:vistited { text-decoration: prop }`
3. It's reasonable that we expect where a `vf-link` type of class has been used that we want to not use a global `a:visited` style


khawkins98:
n.b. As we get further down the line of "integrating with site/service X" I think we'll have many more normalisation issues, so I've made a tag.

PR #288, opened 2019-03-09
Feature/embl breadcrumbs enhancements

khawkins98:
Various refinements:

- makes it so the spacing doesn't flash as breadcrumbs load
- replaces "breadcrumbs are loading" with "EMBL.org" 
- handle when there is no parent
- doesn't insert root who/what/where
- deals with network issues better

Probably about as far as I can take the concept until we load up more real data in the ontology/taxonomies.


components/embl-group-page/embl-group-page.hbs
```diff
@@ -42,7 +42,7 @@
 
     <section class="embl-grid">
       <nav class="vf-breadcrumbs embl-breadcrumbs-lookup" aria-label="Breadcrumb" data-embl-js-breadcrumbs-lookup>
-        <div class="vf-text embl-breadcrumbs-lookup--loading-text"></div>
+        <div class="vf-text embl-breadcrumbs-lookup--ghosting"></div>
```
sturobson: if this is now no longer text - do we need `vf-text` and if it is 'text' as it's loading in text should it be a `p` rather than a `div`?

khawkins98: Probably don't need the text. I'll double check

khawkins98: Ah, actually just failed to fully update the group-page.hbs example; should have been `  <div class="vf-list vf-list--inline | vf-breadcrumbs__list | embl-breadcrumbs-lookup--ghosting"></div>`

Sorted now.

PR #291, opened 2019-03-11
bug: badges don't need normalisations

khawkins98:
They should have bottom borders, no underlines.

Fixes bug introduced by #286

### DIFF
--- a/components/vf-badge/vf-badge.scss
+++ b/components/vf-badge/vf-badge.scss
@@ -63,7 +63,7 @@ $vf-badge--tertiary-color--text: set-color(vf-color-white) !default;
   border-color: $vf-badge--primary-color--border;
   color: $vf-badge--primary-color--text;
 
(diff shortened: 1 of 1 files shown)


sturobson:
¯\_(ツ)_/¯ I think the badge should have it's own hover, rather than the text

sturobson:
so … I'm thinking that there should be two categories for 'links`. Both having sub-categories.

I think this should fix some issues, as well as making it more clearer in the Sass files as to what we're using rather than a long string of variables.:

- `vf-badge` pattern, when it is use as, or part of, a link. 
- `vf-summary--people` (for example) that has lots of links in the content.


- text links
  - content text links
  - primary call-to-action text links
  - secondary links

- button links
  - for actual button links
    - outline buttons
    - normal buttons
  - for things that have their own design links (like `vf-badge`)

sturobson:
moving my comment to an issue, and merging this for now.

PR #293, opened 2019-03-12
Feature/embl core

khawkins98:
stashing this as a PR for now.

As we go down the line of having:

- a generic "VF core" pattern library
- the VF for EMBL Design Language pattern library/design system 
- a perhaps distinct EBI pattern library or subset

we'll need ways to generate independent kitchen sink CSS and JS (for example, `embl-core` doesn't need `vf-logo`, and `vf-core` doesn't need `embl-breadcrumbs`).  

Part of the answer will be the `vf-child-template` but I think for some of the core EMBL stuff we don't want to be faffing about with 3 or 4 repos, ya? So this is a first hot take towards how we might be able to do that.


khawkins98:
I think this one is well and truly and outdated idea. Closing.

PR #294, opened 2019-03-13
adds image in links list

sturobson:
This will add an image to the links list so that it can be used in patterns like the resources list for profile pages.

We might need to tweak to CSS.


khawkins98:
Not sure if "Ready-to-be-merged" so have left open.

PR #296, opened 2019-03-14
bug: Swab svgo fro svgmin

khawkins98:
The practical reason for this is that [gulp-svgo](https://www.npmjs.com/package/gulp-svgo) was "touching" the file even if there were no changes, causing an optimisation loop.

- svg edited/added
- svgo
- svg seen as edited
- svgo
- svg seen as edited
- svgo
- _ad infinitum_

[gulp-svgmin](https://www.npmjs.com/package/gulp-svgmin) doesn't produce the same behaviour and also uses `svgo` under the hood.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -42,7 +42,7 @@ const syntax_scss = require('postcss-scss');
 const gulpStylelint = require('gulp-stylelint');
 
 // Image things
-const svgo = require('gulp-svgo');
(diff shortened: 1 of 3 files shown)


sturobson:
Great! I've noticed all the new `.svg` files each time I make a branch from `develop` and just never committed them. This, at least, would stop dirtying up the terminal with new files to `git add`. 

As it's using SVGO under the hood it's perfect.

Issue #152, opened 2019-01-15
DOCS: Deprecation process


khawkins98:
1. [ ] What does it mean for pattern "deprecation" 
1. [ ] How to deprecate a pattern
1. [ ] Put that documentation somewhere

khawkins98:
Draft text:

# Deprecating patterns

Inevitably, code will be replaced and improved and patterns will need to be marked as deprecated so:

1. Developers know to stop using the pattern; and
1. Aren't removed so existing sites continue to function

## How to deprecate

In the pattern folder:

1. In `pattern-name.config.yml` add:
     - `status: deprecated`
1. In `pattern-name.scss` wrapp the sass with:
    - `html:not(.vf-disable-deprecated) {`
1. Update `README.md` with:
    - why the pattern has been deprecated
    - what pattern(s) developers should instead use

## Need to deprecate just one variant in a pattern

The exact process by which to do this is still being decided, but for now:

1. Add a note to the `--variant.hbs`
1. Wrap the relevant css in `html:not(.vf-disable-deprecated) {`
1. Hide the variant tab in `.config.yml`:
```
variants:
  - name: deprecated-variant
    hidden: true
```

khawkins98:
This documentation was added in #347. 

Still be good to have a gulp task for this, but that one's a bit lower priority. I don't think we'll forget about the need to automate, but @sturobson if you think we might, maybe make a seperate ticket and we can backlog?

Closing this.

sturobson:
let's raise the ticket when we've got the time to work on it 😎

Issue #128, opened 2019-01-06
DOCS - Add a "filesystem" tree map


khawkins98:
This should help with onboarding and just be useful in general. Probably for the README.md.

Probably we'd want to use `tree`, here's a command I got to in a few minutes: 
- ` tree  -L 2 . -I 'node_modules|cache|test_*|.git|.github|bin' -a`

Which gives (+some manual annotation)

## Filesystem notes

```
├── .editorconfig
├── CHANGELOG.md
├── CONTRIBUTING.md
├── README.md
├── SETTINGUP.md
├── assets  (your global Sass assembly file)
│   └── scss 
├── build  (dynamic folder used for rendering of static pattern library)
├── components  (all the components you wish to use)
│   ├── _previews (templates for the web pattern browser)
│   ├── vf-activity-group
...
│   └── vf-video-teaser
├── docs (documentation files for the web interface)
│   ├── changelog
│   └── guidelines
├── fractal.js (configuration for the web pattern library)
├── gulpfile.js
├── lerna.json 
├── package.json
├── public  (dynamic folder used for rendering of global CSS, JS, pattern assets)
└── tools 
    ├── component-generator (make new components)
    ├── css-generator (build requirement)
    ├── frctl-mandelbrot-embl-subtheme  (embl web interface)
    └── frctl-mandelbrot-vf-subtheme ( vanilla web interface)
```

☝️ this is probably less important for the `vf-core` and more so for the child theme template.

khawkins98:
Adding to backlog.

khawkins98:
Actually this exists already at https://github.com/visual-framework/vf-core/blob/develop/docs/contributing/file-structure.njk

Removing from backlog.

sturobson:
aside: I wonder if we can move `_previews` outside of `vf-component` and if we did what performance hit it would or would not create.

khawkins98:
Seems they're "just another component" so that means:

- They've go to go in `./components` ... but
- As they're a component we could call it `vf-previews` or
- Call it `vf-boilerplates`, which might be useful as a collection of generic page layouts

All that said, we've got 5 preview templates, I think we really only need 3, or maybe even 2. Potentially we don't even need any. Related: https://github.com/visual-framework/vf-core/issues/213

sturobson:
should we amend what we call 'a component' and decide it to mean:

> A component is something that is used to create web pages and products for EMBL

which would mitigate `_previews` from being a 'true' component?

[...]

PR #298, opened 2019-03-14
Doc cleanup

khawkins98:
An umbrella branch for all sorts of documentation work.

Addressed so far:
- Add deprecation text for #152

@sturobson Please feel free to add/edit your own ideas on here.


khawkins98:
Candidate tasks for this branch: https://github.com/visual-framework/vf-core/issues?q=is%3Aissue+is%3Aopen+label%3A%22Type%3A+Documentation%22

khawkins98:
Ok, a lot of cleanup of documentation here and adds some things for #152 and #128.

It's not the end of our documentation journey, but gets it to a point where:

- Adds a few basics
- Fixes a few typos
- Makes it so the meat of all VF-core documentation is in `/docs` (even if we move it elsewhere later)

Also not in this PR: making the documentation navigable, that has dependency on theming questions.

PR #307, opened 2019-03-16
pattern: Update hbs template for vf-intro

khawkins98:
This makes the hbs template more of a "real" template and more copy-pasteable into other things like 11ty.


khawkins98:
Once I get the method to integrate these into 11ty down, imagine we'll have quite a few more.

sturobson:
> Once I get the method to integrate these into 11ty down, imagine we'll have quite a few more.

You’re getting something with 11ty set up too?

Perhaps we should set up a larger trello or something. So we don’t end up doing similar the same tasks?

PR #310, opened 2019-03-18
Feature/variant modifiers

khawkins98:
I think this is a good balance between templating and usability for both those needing templates and to skim docs.

Idea is to introduce an approach of having a (where required) `vf-pattern--template.hbs`; so that:

Devs using `npm`'d templates:
`{{> @vf-heading--template large="true" title="Hello World"}}`

Devs needing html docs:
image

And then building the patterns:
```hbs
<!-- Variant modifiers from the pattern Yaml -->
{{#each vf-heading-modifiers as |modifier type|}}
  {{> @vf-heading--template modifier title=modifier.title }}
{{/each}}
```

```yml
  vf-heading-modifiers:
    - display: true
      title: This heading size is display and inverted
    - extra-large: true
      title: This heading size is extra large
    - large: true
      title: This heading size is large
    - regular: true
      title: This heading size is regular
    - small: true
      title: This heading size is small
```

### DIFF
--- a/components/vf-box/vf-box.hbs
+++ b/components/vf-box/vf-box.hbs
@@ -1,4 +1,4 @@
 <div class="vf-box">
-  <h2 class="vf-text vf-text--heading-l">Hello World</h2>
+  {{> @vf-heading--template large="true" title="Hello World"}}
   <p class="vf-text vf-text--body-r">This is some more content that would be in the box.</p>
 </div>
--- a/components/vf-heading/vf-heading--template.hbs
+++ b/components/vf-heading/vf-heading--template.hbs
@@ -1,0 +1,19 @@
+{{!--
+  Pass in your display type and title:
+  {{> @vf-heading--template small="true" title="I'm a small headline"}}
+--}}
+
+{{#if display}}
+  <h1 class="vf-text vf-text--heading-xl vf-text--invert">{{{title}}}</h1>
+{{else if extra-large}}
+  <h2 class="vf-text vf-text--heading-xl">{{{title}}}</h2>
+{{else if large}}
+  <h3 class="vf-text vf-text--heading-l">{{{title}}}</h3>
+{{else if regular}}
+  <h4 class="vf-text vf-text--heading-r">{{{title}}}</h4>
+{{else if small}}
+  <h5 class="vf-text vf-text--heading-s">{{{title}}}</h5>
+{{else}}
+  {{!-- catch all for if the heading type doesn't match --}}
+  <h6 class="vf-text vf-text--heading-s">{{{title}}}</h6>
+{{/if}}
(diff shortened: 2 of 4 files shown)


khawkins98:
n.b. I had tried another approach with slimmer template logic, but it would have been less useful as a template for other content: https://github.com/visual-framework/vf-core/pull/310/commits/50f65d3cd73d3cd9b055a7a108a31c81b5744c46

sturobson:
If we're introducing `--template` I think it would need to be across all patterns so that there's not any cognitive load of 'should I be using `--template` here or not?'.

My thought was to have `pattern--example.hbs` as the page that would have the loops in - leaving `pattern.hbs` to be the one that's used with whatever arguments we decide upon.


I'm in two minds about where all those `if/else` statements are better / worse than adding the classname. 

Either way this should show up any 'differences' in how we're naming things (if at all) … as I'm sure there could be confusion as to why `component-x.hbs` uses small/medium/large but `component-y.hbs` uses s/r/l (although I think we've been fairly consistent here.

khawkins98:
I think we do want to get to the point where all the patterns have _real_ templates. 

Had thought about doing `--example.hbs` but that felt awkward in practice; to try and sketch out:
- vf-pattern.hbs: hidden from pattern library
- vf-pattern--example: shown
- vf-pattern.scss, code applies to  vf-pattern--example

In reality, I think what we'd want to be able to do is that:
- `vf-patten.hbs` = just a template
- `README.md` generate by HBS and all the former demo stuff

Obv we're not quite there yet though.

sturobson:
- vf-patten.hbs = just a template
- README.md generate by HBS and all the former demo stuff

Is what I got working before xmas and it was pretty nice.

khawkins98:
Was just looking at the fractal docs.

Got excited:

> If present, the contents of the README file will be run through the full Fractal documentation parser before being displayed to the user in places such as the web UI. That means that the file will be rendered with the documentation template engine (Handlebars by default) before being passed through a Markdown parser.

Then sad: 
> WARNING
> Fractal does not support the use of YAML front-matter in README.md files.

https://fractal.build/guide/components/notes.html

Gonna have a quick poke though

sturobson:
I think it'd be nicer, easier to grok for developers using nom to do:
`{{> @vf-heading size="large" title="Hello World"}}`

rather than:
`{{> @vf-heading--template large="true" title="Hello World"}}`

I agree re: `--example` but I'm wondering if needing `--template`would be a case of 'why do I need to add 'template'?

khawkins98:
On `size="large" `, agree. Trade of is getting an `if (size == "large)` working in HBS. 

It's not hard but requires an additional helper to be installed on the client. Not an impossible ask, but one of those things. sigh.

khawkins98:
I think this approach can be reassessed and done far better after we've moved to nunjucks templates #302

PR #312, opened 2019-03-19
updates ORCID to be an abbr

sturobson:


khawkins98:
So, wrote something on this, lost it, then had some more thought. 

I've not yet asked  you why you think this is important, but my assumption is for 1) general accessibility and 2) to define an acronym.

I think it's both of those + representative of a general area where we could do better for EMBL and lead a bit for the wider life science community; that is: exactly what the VF is about. 

With all that in mind, here are my thoughts:

- I'm not sure it's productive to state what ORCID stands for, in much the same way it would be to `DVD=digital versatile disc`. 
- I also took a little time to track down what [ORCID's guidance](https://orcid.org/trademark-and-id-display-guidelines#02-c) is: 
  > The name of the ORCID organization is “ORCID” ... Do not use “Open Researcher and Contributor Identifier”
- While many (most?) of our users will know what ORCID is, I know that not all will. 
- Better than an acronym we could offer a nice concise explainer, for example, as Europe PMC [puts it](https://europepmc.org): 
  > ORCID is a unique identifier for researchers which distinguishes you from every other researcher, and makes it easier to find your work.
  - that text — with a couple tweaks — plus a link to the ORCID website would do be really helpful for those new to ORCID

☝️ that idea of a description is probably not a good fit for `<abbr>` so we could instead use something like `aria-describedby` + `<dfn>`; pseudo code:

```html
<span aria-describedby="whatis-orcid">ORCID</span>

...

<div id="whatis-orcid" class="vf-expaliner">
  <dfn>ORCID</dfn> is a unique identifier for researchers to distinguish themselves, and make it easier to find specific work; <a href="https://orcid.org" target="_blank" class="vf-explainer__link">learn more</a>
</div>
```
Couple of notes on that snippet:

1. I'm thinking it's the start of a new `vf-explainer` pattern that could be used in _many_ situations for things like "transcript isoform" and "lymphoblastoid cell lines"  that show up fairly high in our websites ... and potentially even used by more scientist-specific pages and data resources.
  - There might even be some sort of "data-vf-jargon-level=10" scoring that could allows users to specify their expertise. #possibleStupidIdea
1. Obviously we'd need some sort of tbd mechanism to hide/show/make accessible that block of text. 
1. I think there was a third point, but I've forgotten now. The above is probably plenty to chew on.

So maybe what this is really a pattern request.

sturobson:
With you finding –

I also took a little time to track down what ORCID's guidance is:
The name of the ORCID organization is “ORCID” ... Do not use “Open Researcher and Contributor Identifier”

I think we can safely close this and move forward with your other idea

PR #313, opened 2019-03-19
Refactor/vf masthead

sturobson:
**note:** this is a **draft** PR - subject to change and flux as the general idea is worked on

There's still a bit to do here -- but essentially we're moving towards containers having a max-width, instead of a grid to position them on the page. 

This kinda works ok for `vf-inlay` patterns but I think we could generalise it a little more so that it's more like one masthead to rule them all.

`vf-masthead`, and `vf-navigation` gets some new full width background colours with the `::before` pseudo class.

`vf-masthead--inlay` has the start of what i'm thinking we could do with theming.


components/vf-masthead/vf-masthead.scss
```diff
@@ -2,14 +2,45 @@ $vf-masthead__title-text--color: set-color(vf-color-white);
 
 // vf-masthead
 
-.vf-header .vf-masthead {
-  display: grid;
-  grid-template-columns: minmax(var(--page-grid-gap), auto) minmax(auto, 76.5em) minmax(var(--page-grid-gap), auto);
+
+.vf-inlay__header {
+  margin: 0 auto;
```
khawkins98: Hello, old friend.

components/vf-masthead/vf-masthead--inlay.hbs
```diff
@@ -1,25 +1,29 @@
...
+<style>
+:root {
+  /* --vf-masthead__color--background: none; */
+  /* --vf-masthead__color--foreground: #000000; */
+  /* --vf-masthead__bg-image: url('http://www.fillmurray.com/800/200'); */
+}
+</style>
+
+<header class="vf-inlay__header">
+  <div class="vf-masthead" style="background-color: var(--vf-masthead__color--background, #000000); color: var(--vf-masthead__color--foreground, #ffffff);">
```
khawkins98: I'm guessing defaults for these (`--vf-masthead__color--background`) would be set in the Sass somewhere, ya?

sturobson: So the custom properties in the `:root` Would need to be added to the document after the css is linked in the head and before the component. 

The defaults are inline in the mark up (black and white in the example), so if there’s no theme or that bit doesn’t get included there’s a ‘nice’ default (of course black and white is just here as an example. It could easily be green white.)

This pattern also allows for background images too. So I think it’s a more general ‘masthead’ pattern with its parent dictating how wide it is. For example. In vf-inlay it would only need some padding on the parent header really.

components/vf-masthead/vf-masthead.scss
sturobson: Yeah. The more I think about it. The more I’m thinking that whole page doesn’t need a grid right now. And we can use `:before` when we need colour or image ‘bars’.

components/vf-masthead/vf-masthead.scss
```diff
@@ -28,9 +59,19 @@ $vf-masthead__title-text--color: set-color(vf-color-white);
 
 
 .vf-masthead__heading {
-  @include set-type(heading--xl);
+  @include set-type(heading--xl, $custom-margin-bottom: 8px);
 
-  color: $vf-masthead__title-text--color;
+  color: inherit;
```
khawkins98: We might be careful on changing this to inherit. The specificity inheritance seems to defer to the parent `.vf-masthead` over another stylesheet's closer `h1 { color: #222 }` ... but I'm sure that might vary from browser to browser or what might be going on elsewhere. 

or maybe here we could say `color: var(--vf-masthead__color--foreground)`

components/vf-masthead/vf-masthead--inlay.hbs
khawkins98: Shouldn't we also set some sort of global default in the VF so that way if someone forgets to add a local `style=""` there's a reasonable background -- that and it would help if the EMBL green changes so we won't have the default colour inlined in all the sites.

components/vf-masthead/vf-masthead.scss
khawkins98: Looking at how you're envisioning the theming options, I guess `color: var(--vf-masthead__color--foreground` wouldn't work. 

Perhaps instead we could only support "white" and "dark" foreground colours ... really that should be all someone should need? And if they need some crazy yellow text on blue background, they probably would need to be doing some other crazy stuff, like setting a bunch of other things to yellow and blue.

sturobson: Comment 1: I need to do some testing and thinking
Comment 2: Previously with this I’ve made use of JS to define the foreground colour so that it matches the correct contrast checks for WCAG … (so yes, either dark or light - but determined by the background colour at run time).

khawkins98: yeah, let's not get hung up on this one ... it's something we can refine "in the future".

components/vf-masthead/vf-masthead--inlay.hbs
sturobson: Well, we can set another custom variable in the stylesheet. That would be the back up colour in the `var` rather than being explicit with black/white.

```
<div 
  class=“vf-masthead” 
  style=“
  background-color: var(—custom-defined-bg-color, var(—backup-bg-color))
  “
>
```

(The syntax might be off there - but you should get the idea)

khawkins98: I think maybe we could do what I'm talking about if we had in the pattern sass:

```sass
:root {
--vf-masthead__color--background: #{set-color(vf-color-green)};
}

...

.vf-masthead {
  background-color: var(--vf-masthead__color--background);

```

And then did in the pattern html
```html
<style>
:root {
  /* --vf-masthead__color--background: none; */
  ...
</style> 
  <div class="vf-masthead" style="background-color: var(--vf-masthead__color--background); ...
``` 

That should allow the VF default to be used, unless the variable has been overridden prior to the html.

khawkins98: Comment overlap 🤸‍♂️

sturobson:
… because there are no nice ways of adding a Sass map value as a CSS custom property I have adjusted the `gulp tokens` task to make CSS custom properties of the colours so we can use them.

components/vf-masthead/vf-masthead--inlay.hbs
sturobson: so, the only 'nice' way to do this (because I cannot see a way to put a Sass variable from a Sass map using a Sass function that works). Is by duplicating the Sass variables for colour as custom properties.

I think moving forward that'd be a good thing anyways.

khawkins98:
On the above commits, I got a bit twisted halfway through, so here’s a compare:: https://github.com/visual-framework/vf-core/compare/984147a5c3b672a5493f0826ff05659ad617387c...de70499eff5e96f36314ce7e66a62c1b638ca5cd

This is partly about this PR and documentation in general .. basically, this way we get this on top of the dynamic files:

```
// Do not edit directly: This file has been dynamically generated by
// `gulp theo` from the Visual Framwork `vf-design-tokens`
 
// Source: vf-design-tokens/variables/vf-font-families.yml
```

Otherwise it’s hard to know which of the `variables/*.scss` can/should be edited and which are “dynamic”.

sturobson:
> partly about this PR and documentation in general

should it be in a different PR then? So we're not doing too much in one place

gulpfile.js
```diff
@@ -204,28 +204,54 @@ gulp.task('svg', () => {
 // Design Token Tasks
 // -----------------------------------------------------------------------------
 
-theo.registerFormat("typography-map", result => {
```
sturobson: where's this gone? Wouldn't your comments change the typography maps too?

khawkins98:
> > partly about this PR and documentation in general
> 
> should it be in a different PR then? So we're not doing too much in one place

Well, we need to document this PR ... could do a seperate PR that you could merge in here and then add the comments for this PR...

components/vf-masthead/vf-masthead--inlay.hbs
```diff
@@ -1,25 +1,31 @@
...
-    <div class="vf-masthead">
-      <div class="vf-masthead__inner">
-        <div class="vf-masthead__title">
-          <h1 class="vf-masthead__heading">
-            <a class="vf-masthead__heading__link" href="http://dev-vf-theme-prototype.pantheonsite.io">Häring Group</a>
-          </h1>
-        </div>
-      </div>
+<style>
+:root {
```
sturobson: these won't be living here for much longer. I think we'd need a variant for a masthead with images that takes these variables.

And then documentation about theming (which I'm working on, before we cross streams)

sturobson:
I think I've not helped by muddying the waters with creating the custom properties generation, I guess we can put it all in this one, I'll be sure to document the PR a little better whether it's a draft or not, next time around so there's clarity :)

sturobson:
changed the language of the notice for generated Sass variables to:

```
// This file has been dynamically generated from design tokens
// Please do NOT edit directly.

// Source: vf-design-tokens/variables/vf-breakpoints.yml
```

I didn't think consumers would necessarily need to know how they're generated (the gulp task) and I thought removing "Visual Framework" is better suited as the design tokens c/sh/w/ould eventually be used by more than the component library.

PR #320, opened 2019-03-25
Concept for showing the documentation in the readme

khawkins98:
So this relates to the documentation discussions and is a further iteration on approach we might use to show the documentation in the readme area.

image

Upsides:
- the `.njk` file isn't used as a demo but a real template
- the demo is used in the readme and shown there

Downside/mided:
- the real README.md file won't have the rendered code but rather the code used to make the demo ... i guess that's a feature depending on your view of the use case.

### DIFF
--- a/components/vf-heading/README.md
+++ b/components/vf-heading/README.md
@@ -2,13 +2,29 @@
 
 ## About
 
+Basic headline styles.
+
 ## Demo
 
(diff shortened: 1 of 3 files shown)


khawkins98:
We could either keep this as an idea, or merge at build atop it.

khawkins98:
Think this has been fully eclipsed. Closing.

PR #327, opened 2019-03-28
Features/markdown filters

khawkins98:
Stashing this feature here. We could well need it soon, but at the moment we can live without. And it needs a bit more cleanup.

Adds ability to parse {% markdown %} segments.

I noticed some oddities that the nunjucks filters were passed to the docs parser, but the extension wasn't. Might need to open a ticket for it.


sturobson:
Yes!

I think we're going to need this when we start working on the beta for sure.

khawkins98:
Okay, have had a cleanup and used this as a demo in vf-content -- both to check for errors and as a demonstration.

From my side I think it's worth rolling.

PR #331, opened 2019-03-28
Add missing npm files 

khawkins98:
Collecting a few missing things

### DIFF
--- a/components/vf-logo/package-lock.json
+++ b/components/vf-logo/package-lock.json
@@ -1,6 +1,6 @@
 {
 	"name": "@visual-framework/vf-logo",
-	"version": "0.0.14",
+	"version": "0.0.29",
 	"lockfileVersion": 1,
 	"requires": true,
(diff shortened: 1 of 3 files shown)


khawkins98:
@sturobson I wonder if for the pattern default we'd make maintenance far easier if we just included all files

khawkins98:
There will inevitably be more borked `package.json`s to fix, but this resolves a couple things hurting right now in 11ty. 

And then we'll need to open a new issue to not track files and test that approach ... I'll whip that up real quick.

khawkins98:
I've gone ahead and bumped the  npm versions for these two components and pushed to npm. 

Believe that fits our versioning approach, where patterns don't have to be updated only on "release day"/tied to vf-core versions. https://github.com/visual-framework/vf-core/blob/6f321e9d0cd6aaa41c955805863ba0074a9c41a0/docs/guidelines/versioning.njk#L7

Issue #316, opened 2019-03-25
Chore - Make use of .yml for adding in data/content


sturobson:
There are a few patterns still that have 'hard-coded' content betwixt the HTML declarations. We need to move this out to the .yml file so that the patterns can be imported/included as needed in a SSG-site or single page boilerplate. 

- [ ] put all content in the .yml file
- [ ] make use of the .yml file to 'create' the variants of a component - rather than having several 'types' or if/else statements in the `view` template.

khawkins98:
Tasks to be done:

## YAML and Nunjuck-ise  all the components

Rather than list each component, maybe we can approach each "tier" as PRs:

- [ ] All vf-elements
- [ ] All vf-blocks
- [ ] All vf-components
- [ ] All vf-containers
- [ ] All vf-grids (is there anything to do here?)

- [ ] All embl-elements
- [ ] All embl-blocks
- [ ] All embl-components
- [ ] All embl-containers
- [ ] All embl-grids (is there anything to do here?)

## Documentation

Probably a few types of documentation.

- [ ] High level on the role of components as "real" templates
- [ ] Building a component with .njk + .yml 
- [ ] Consuming a component's template

## Templating

- [ ] Update the component generator

khawkins98:
We've (mostly you @sturobson) have made a lot of progress on updating components and I think this will be an ongoing process as we make new components and use components in sites. 

That is: we'll never fully "finish" this and we won't forget this, so closing.

PR #338, opened 2019-04-01
WIP: updates components to use yaml where possible

sturobson:
There's a mix of components using yaml for content/data and some that aren't.

This PR will normalise its use. It also hopes to make the nunjucks template more flexible


khawkins98:
From purely within the component, I think this is really nice. 

Still some pains in using it as a template within the pattern library, for example to use this as a template for a rounded pill with a custom text I need to pass all this junk: 

```njk
{% render "@vf-badge--outline", 
  {"vf_badge_classes": "vf-badge--primary vf-badge--outline", 
    "vf_badge_name": "test name", 
    "vf_badge_tags": "span"}  
%}
```

for `vf_badge_tags`, it would probably be pretty easy to add an "if not set default", but for `vf_badge_classes` it might require some _cleverness_

Aside, I wonder if it's helpful to namespace these? On one hand, it's a principle we've used elsewhere, but are the Yaml variables and Nunjucks template are already in a special per-pattern bubble? I think so?

khawkins98:
Looks like the defaults are available with the template by using `{{ _self.context['vf_badge_classes'] }}`

So it might just take a couple lines of 

`{% set class = vf_badge_classes || _self.context['vf_badge_classes'] %}` (untested)

khawkins98:
Referencing #316, as the review comment didn't seem to.

PR #342, opened 2019-04-03
Add dynamic sass imports

khawkins98:
Use case: right now if a component isn't "installed" it _must_ be disabled in vf-core/index.scss

That adds a configuration step to customisation and means you can't really ever npm install vf-core unless you want to not add/remove components (which, you really should be customising)

So this idea uses gulp+node as a middle layer to see what components are installed and then — if a pattern is missing — it ignores it at compile  time  in `vf-core/index.scss`.

So that means vf-core could be a kitchen sink of all core modules and you can add/remove components without customising. 

WIP and it needs cleanup.


khawkins98:
@sturobson Before I carry on cleaning this up, any 👍 / 👎  on the concept?

sturobson:
a quick one, whilst I mull this over some more, is `vf-core/index.scss` now `vf-componenet-rollup/index.scss`?

khawkins98:
Yep.

sturobson:
finally - he comments:

Looking good - ideally I think we'd want people to only install the components they need and not this `rollup`?

But this is a great catch 'em if you can if they're using this but not installing all the components.

khawkins98:
Did a little cleanup and tweaking, should be good on here. Will wait until the morning to merge.

PR #346, opened 2019-04-08
ENHANCEMENT - vf-core npm optimisations

khawkins98:
Follows up on #343 by:

1. Making `npm install visual-framework/vf-core` a useful thing so you can require `fractal.js`, the theme or other things without getting duplicate installs of all the vf-core components
2. Deconflicts naming with `components/vf-core`, which is now `components/vf-componenet-rollup`


khawkins98:
After this is merged, we should npm publish the vf-core for testing/use in 11ty.

sturobson:
> We'll also need to depreciate the old @visual-framework/visual-framework package.

run me by this one for clarity?

`vf-core` becomes `visual-framework` in terms of the npm package - is that right?

I may need ☕️

khawkins98:
> vf-core becomes visual-framework in terms of the npm package - is that right?

Yep, that's basically the idea.

Issue #333, opened 2019-04-01
CHORE - Track all component files for npm


khawkins98:
As discussed (briefly) at https://github.com/visual-framework/vf-core/pull/331#issuecomment-477588045, it would simplify component maintenance and increase reliability if we simply tracked all component files for npm.

The easiest (best?) way to do this would be to remove the `files { }` section, https://alligator.io/nodejs/package-json/:

> files: This optional key expects an array of files that are included when the project is added a dependency to another project. File names/paths can use a glob pattern and when the files key is not provided a default value of ["*"] is used, which means that all the files will be included. Not to worry though, certain files/folders such as .git and nome_modules are always ignored.

If that, suits we'd need to:

- [x] Update the component builder template
- [ ] Update all the components
- [x] Consider adding [`.npmignore`](https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package) to the template

sturobson:
Hmmm, my original thinking is that people don't necessarily need to install the `.md` files etc for a project (adding that 'bloat' people complain about with Node and NPM). 

In saying that, I think removing `files { }` will definitely improve us shipping things without the extra checks to make sure we've included any additional files added to the folder.

I'm happy with this move forward.

khawkins98:
Addressed by #388

Issue #178, opened 2019-01-22
QUESTION: How to deprecate a variant


khawkins98:
When one variant in a pattern is no longer to be used/deprecated, how should we do that?

We can achieve this in part by:

1. Adding a note to the `--variant.hbs`
1. Wrapping the relevant css in `html:not(.vf-disable-deprecated) {`
1. Hiding the variant tab in `.config.yml`:
```
variants:
  - name: deprecated-variant
    hidden: true
```

Is that sufficient? Can/should we do anything more? 

Related to: `DOCS: Deprecation process` #152

khawkins98:
I think we're close on this one, but we need to have a think before we close. Bumping to alpha.4.

sturobson:
We might want to circle back and see what other things could possibly be deprecated - design tokens for one.

khawkins98:
Added to the backlog.

PR #347, opened 2019-04-08
Chore/deprecation documentation

khawkins98:
Addresses #152 and #178  and follows up on the deprecation branch to document how/why/what on deprecation.

Would need to be merged into PR  #333


khawkins98:
Nice catch.

Issue #114, opened 2018-12-19
DOCUMENTATION - VF Governance


khawkins98:
As we enter the alpha releases we should start getting wider consensus on the approaches of the VF (note this is seperate from the EMBL brand).

So we should have some kind of governance, documented procedures, and some sort (or advisory group) that provides feedback and buyin, and results in some sort of documentation on:

1. What was agreed
2. Why it was agreed
3. Forever-hold-your-peace 

So we need to:

1. Clarify the remit
2. Find some core members (might expand later, better too few than too many)
3. Set up a first date 
4. Establish frequency (bi-monthly? quarterly?)

khawkins98:
I think we should address this sooner than later -- and I think it's mostly about governance, I've updated the description accordingly.

khawkins98:
The pressing need here has been addressed in the above PR. Closing this one, we can iterate on the text as the need arises.

PR #351, opened 2019-04-09
Documentation/governance

khawkins98:
This wouldn't be the end of governance text, but I think it's a good mark in the sand that captures the spirit. 

Addresses #114


khawkins98:
@sturobson If you've no objections here, I'll merge in around lunch time. We can always tweak/update as we go on.

PR #354, opened 2019-04-15
Refactor/vf-link

khawkins98:
Wanted to try following your approach to yamalise a pattern to see if I was following, but as typical it turned up a couple other things ... so:

1. Is this approach to the yaml/njk template consistent with your approach?

I also tried "demos" of the component in the README.md:

2. That largely worked fine, but I had an issue with the Markdown parser turning the code into paragraph text due to whitespacing generated by the Njk template. So I added a  `{% spaceless %}` filter that we can wrap components in to suck out all that empty space.


sturobson:
Hey - looks great!

one small thing:

I'm not sure that a link with secondary or tertiary theming would have a different 'disabled' colouring - as that could be quite confusing. 

So, I've edited the .njk file to put the `--disabled` as a generic class that can be added.

khawkins98:
Sounds reasonable. Was thinking that one day we might have some sort of "special thing" for disabled links, but no point in outputting classes that don't do anything. 

---

Doing this little test gave me _thoughts_ about how we might better use variants, but I think getting into that now might be distracting 🙃

sturobson:
> Doing this little test gave me thoughts about how we might better use variants, but I think getting into that now might be distracting

Want to spark up a new issue to discuss?

sturobson:
there's a couple of things in the CSS I want to look at before merging this in.

khawkins98:
Did you want to incorporate #355 in this PR?

sturobson:
@khawkins98 I haven't (yet) as I thought it would 'muddy' the PR. I guess I can change the title and description though :) 

todo:

- [ ] turn `vf-link` into `vf-u-link`
- [ ] create a `vf-u-link--primary` 
- [ ] update any instances of `vf-link` being used.

I'll start a separate branch for turning `vf-u-link` into `vf-componentName-link`

components/vf-link/vf-link.njk
```diff
@@ -1,15 +1,45 @@
...
+{# Here is where we are adding the vf-text--body modifier #}
+class="vf-link
+
+{% if theme %}vf-link--{{ theme -}}{% endif %}
+
+{# If you're styling the shape of the item, the classnames will render in this for loop #}
+{%- if style -%}
+  {% for styles in style %}
+    {# Link is disabled? #}
+    {%- if styles == "disabled" %}
```
khawkins98: I think we need to updated this to output `vf-link--disabled`, if I'm following?

sturobson: this is creating `vf-link--disabled` for me as well as adding the HTML5 'wotsit' of `disabled` to the element too.

Originally it was adding `vf-link--secondary--disabled`, so I moved the logic around a little to stop it from doing that.

Is it not working for you?

PR #367, opened 2019-04-25
fix: embl-grid-and-breadcrumbs

khawkins98:
Anticipates placement of breadcrumbs inside `embl-grid`


sturobson:
I was working on this as part of getting the pages to look nice.

merging.

PR #371, opened 2019-04-29
Fixes/inlay header errors

sturobson:
this gets us over the line - although I'm not a fan of the masthead codebase still - as I'm mixing vf-inlay__haeder in there -- feel free to take a look or merge if you're happy for this working


khawkins98:
Looks good. 

Agree it could use a refactor, it has been a bit brittle 😢

PR #372, opened 2019-04-29
refactor vf-blocks to use yaml and nunjucks where possible/sensible

sturobson:
- [x] Activity List
- [x] Article Meta Information
- [x] Box
- [x] Breadcrumbs
- [x] Discussion
- [x] Global Header
- [x] Lede
- [x] Links List
- [x] Navigation
- [x] Page Header
- [x] Search
- [x] Section Header
- [x] Summary
- [x] Tabs - *requires* a look a the JS as initial tab no longer opens, although HTML unchanged
- [x] Video
- [x] Video Teaser

Not being updated:

- Masthead
- Pagination
- Footer - this is more of a _container_ really so needs a little bit more of a refactor than adding yaml
- Code Example
- Plex Mono Font
- Plex Sans Font


sturobson:
can you take a look to see what is going on with `vf-tabs` now I've YAML'd it? The HTML, as far as I can see is exactly the same?

Does the tabs JS run before the yaml data gets injected? Can we define a 'default open tab'?

khawkins98:
Just pushed a fix. Issue was that there was one more "content" tab in the nunj template than associated headers, which was confusing the JS.

Also moved the nested tab demo out of the fixed template and into the yaml.

components/vf-tabs/vf-tabs.config.yml
```diff
@@ -8,16 +8,35 @@ context:
...
+      tab_content: Phasellus ac tristique orci. Nulla maximus <a class="vf-link" href="#">justo nec dignissim consequat</a>. Sed vehicula diam sit amet mi efficitur vehicula in in nisl. Aliquam erat volutpat. Suspendisse lorem turpis, accumsan consequat consectetur gravida, <a class="vf-link" href="#">pellentesque ac ante</a>. Aliquam in commodo ligula, sit amet mollis neque. Vestibulum at facilisis massa.
     - tab_title: Nested Tabs
       tab_heading: Section 4
       tab_number: 4
-      tab_content: Nam luctus, enim in interdum condimentum, nisl diam iaculis lorem, vel volutpat mi leo sit amet lectus. Praesent non odio bibendum magna bibendum accumsan. Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nam euismod, tortor nec pharetra ultricies, ante erat imperdiet velit, nec laoreet enim lacus a velit.
+      tab_content: >
```
sturobson: So - I can work on this by using the yaml/nunjucks -- like I have done with `vf-discussion` - or we could leave it in the yaml like this - i don't mind

khawkins98: Do you mean to support nested tabs? I don't necessarily think that _must_ be supported by our njk template, more that it's been done here to show (and test) that nested tabs are working.

sturobson: cool

PR #382, opened 2019-05-17
Refactor/vf masthead

sturobson:
**There _are_ breaking changes.**

This PR attempts to fix the problems we have had with the masthead working nicely for inlay pages, but not for other things. 

It moves out the `vf-inlay__header` to become `vf-header--inlay` within the `vf-header` pattern

It makes it so that the navigation works full-width or within `vf-header--inlay`.
It makes the `vf-masthead` take up as much room as needed.
It updates the `embl-group-page` example.

It also adds the ability to determine the text colour depending on the main colour of a background image (defined as part of the `.png` file path. 

It adds **documentation** for the all of the above.


sturobson:
I've only just caught this set of messages, changes, and approval :-/

Thanks @khawkins98 👍

PR #388, opened 2019-05-29
Adds npmingore to components

khawkins98:
This should greately ease maintance and reduce the number of times a file is left out of npm.


sturobson:
👍

Issue #324, opened 2019-03-26
BUILDS - split out gulp tasks


sturobson:
This is a sub-issue of #321 

> The gulp build is global.

(probably not the best sentence to quote, but heh)

If we're looking at building more functionality with the gulp tasks - it is probably time we look at splitting these out as separate 'things' in a `gulpfilejs` folder.

This might make it easier in the overall `gulpfile.js` file to see what is doing what, where and why - and we can easily hunt down the separate tasks when any editing is needed.

khawkins98:
Breaking up the gulpfile would probably be a necessary and helpful step in any situation.

PR #396, opened 2019-06-06
Feature/gulpfile optimisations

khawkins98:
This does a couple things:

1) Namespaces the gulp commands so that they won't conflict if the gulpfile is being included in another project
    - also updates docs to say "use `gulp vf-dev`"
2) Makes better use of variables to set the component path for sass, js, etc.

Both of these will be of particular importance to things like vf-eleventy ... a separate PR to come there. 

Keeping this as a draft until the vf-eleventy stuff is prepped.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -35,7 +35,7 @@ build:
   script:
     - npm install
     - npm install -g gulp-cli
-    - gulp build
+    - gulp vf-build
   artifacts:
     paths:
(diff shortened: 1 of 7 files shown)


sturobson:
as in you're in `gulpfile.js` do you want to take on #324 as I've not assigned it to me yet?

khawkins98:
👍  snagged that task and assigned to myself.

PR #398, opened 2019-06-07
Refactor/vf design tokens

sturobson:
this Pull Request does quite a bit under the hood:

- updates the `vf-design-tokens` `.yml` to include more information to aid generation of examples for the `vf-core` component library and the embl -`design-system` repos
- pulls `vf-design-tokens` out of the day-to-day gulp tasks and has it as it's own little thing.
- puts all the tokens generation inside of `vf-design-tokens`
- updates all the Sass variables so they're a bit more BEM'y
- takes the variables from `vf-design-tokens` folder rather than generate them(1)
- creates a `dist` folder of design tokens into Sass variables and .json
- generates `.ios.json` format so the design tokens can be imported into the `design-system` repo.

1) it could be worth looking at pulling the `vf-design-tokens` in as an nom module?


components/vf-design-tokens/gulpfile.js
```diff
@@ -0,0 +1,150 @@
...
+    'font-size': ${prop.getIn(["value", "font-size"])},
+    'font-weight': ${prop.getIn(["value", "font-weight"])},
+    'line-height': ${prop.getIn(["value", "line-height"])}
+  ),`
+  )
+  .sort()
+  .join("\n")}
+
+);
+{{#if prop.comment}}
```
khawkins98: I think this is a typo? It's generating a bit of junk in some `.scss` files.

khawkins98:
Added one comment about a possible typo, also we should probably add documentation somewher about updating/using design tokens. I added a little to the readme.

khawkins98:
Oh, and I guess we need to rename this to `  "name": "@visual-frameworks/vf-design-tokens",`? and depricate `vf-tokens` on npm?

components/vf-design-tokens/gulpfile.js
sturobson: yes - need to look at that :D

sturobson:
I've dropped this back in as 'vf-design-tokens' - I'm now thinking that there could be more questions like - "tokens, what are tokens?" whereas keeping 'design' might mitigate it a little - or make it a little more obvious.

PR #418, opened 2019-06-19
Fixes/colors

### LINKED ISSUE #417: Colour tokens are broken
Following #398 many (most?) of the colour tokens are busted; for example the [`vf-divider`](https://visual-framework.github.io/vf-core/components/preview/vf-divider.html) has no background.

https://github.com/visual-framework/vf-core/blob/d65c37340bbddf7b031d0dc5a2cd1efadf7c6389/components/vf-sass-config/mixins/_divider.scss#L4

When the token is:

https://github.com/visual-framework/vf-core/blob/d65c37340bbddf7b031d0dc5a2cd1efadf7c6389/components/vf-design-tokens/src/core.palette.alias.yml#L62

I'm not sure which way is right -- with one hyphen or two?

sturobson:
This closes #417 

I've BEM'd the BEM in this instance - I'm not convinced still … and as previously mentioned I think it should ultimately move towards a scale like font-weights - so it can allow for other shades or tints of the default colour to come through.

I think that will open a bigger kettle of fish with the current colours not following a scale but seemingly decided upon for … ¯\_(ツ)_/¯ reasons.

### DIFF
--- a/components/embl-breadcrumbs-lookup/embl-breadcrumbs-lookup.scss
+++ b/components/embl-breadcrumbs-lookup/embl-breadcrumbs-lookup.scss
@@ -9,7 +9,7 @@
 
 .embl-breadcrumbs-lookup--ghosting {
   animation: embl-breadcrumbs-lookup-ghosting-animation 4s ease-in-out infinite alternate;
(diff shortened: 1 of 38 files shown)


khawkins98:
If we ever get some non-busy time, would be good to sit down with @markboulton  and see if we can hash out a more logical approach to the colours chaos.

sturobson:
I need to fix `vf-color--white` as I somehow removed it from `vf-design-tokens`.

It's more a UI color than a theme colour - so I'm going to throw in a quick edit to `vf-design-tokens` into this PR too.

khawkins98:
Ah, have just seen your comments when I was looking through the code on my local. I think I've gotten most of those fixes.

PR #419, opened 2019-06-19
Update vf-core version for alpha.6

khawkins98:


sturobson:
also in 'non-busy times' - a script that can automate most of the moving parts to create a release.

PR #420, opened 2019-06-19
Fixes/colors again

sturobson:


components/vf-design-tokens/src/maps/vf-ui-colors.yml
```diff
@@ -20,6 +20,12 @@ props:
       friendlyName: UI Grey light
       sassVariable: $vf-ui-colors-map, vf-ui-color-color--grey--light
       CSSCustomProperty: --vf-ui-color--grey--light
+  - name: vf-ui-color--white
```
khawkins98: Not sure if you've mean to leave out `ui-color--off-white`?

sturobson: black and white are now ui colors as they're #FFF and #000 rather than slightly white or black variants for a theme.

khawkins98: What I mean is in `components/vf-design-tokens/src/core.palette.alias.yml` there is: 
```
  ui-color--off-white: 'rgb(250,250,250)'
```

It's referenced in various components, not sure if you're meaning to remove `off-white` or just forgotten to map it.

khawkins98: Oh, wait, there it is now. strange. having a second look.

khawkins98: Ok, figured it out. I had added both of these as UI colours and you added white, which was causing some odd things to happen and also my confusion. Have pushed a commit sorting it.

sturobson: Thanks - I ended up in git hell trying to fix it :sob:

sturobson:
that last update changes `azure` to `blue` and `ruby` to `red`. It's also snagged a couple of loose `-dark` that should be `--dark`.

PR #431, opened 2019-06-24
Fix: docs should soft fail when vf-design-tokens aren't present

khawkins98:
This is a workaround so the doc compilation in fractal will softfail when `vf-design-tokens` aren't present.

A few notes:

1. When the docs don't compile, the fractal components aren't available
2. This is mostly an issue for `vf-eleventy` sites, but fractal should be compiling the vf-core docs anyways 🤷‍♂ 
3. ☝️ That could also be an issue in `vf-core` if/when we move vf-design-tokens as an npm install

To me all that continues to suggest that we still need a more "holistic" way to approach the docs, so this PR is more of an "alpha-level" fix.

### DIFF
--- a/docs/styles/colours.config.js
+++ b/docs/styles/colours.config.js
@@ -2,8 +2,22 @@
 
 const path = require('path');
 
-module.exports = {
-  context: {
-    colors: require(path.join(process.cwd(), 'components/vf-design-tokens/dist/json/vf-colors.ios.json'))
+// Only generate the colour tokens if the vf-design-tokens are installed.
+// That is: if this is a full VF desing system, we probably have design tokens,
+// if not we're likely running as a VF component system and we don't want things
+// to blow up
+try {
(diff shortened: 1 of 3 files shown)


sturobson:
It's bizarre we're hitting this - I wonder if there's a better way. I really want to tie up the `design-system` site this week or next - so I'll be taking another look.

khawkins98:
Yeah, wonder if this sort of `docs` should just live at the design system level?

sturobson:
the reason for the separation at the moment is that `design system` is really only 'high level' so people looking at it shouldn't really care what the CSS custom property is.

Although, we could re-jig bigzing! So that the `design system` is actually where _everything_ lives.

Which gets my vote :)

khawkins98:
It kinda feels like things are tilting that way-ish. Would be a more natural fit if, sya, EMBL and EMBL-EBI "VFs" have slightly different colour palettes.

PR #432, opened 2019-06-24
Chore: don't send backstop files to npm

khawkins98:
Just a cleanup thing I suppose.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -10,3 +10,4 @@ build
 .max-age
 .gitlab-ci.yml
 temp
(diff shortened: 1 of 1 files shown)


sturobson:
Ideally - I think we need to look to Percy.

Taking them out of version control could lead to some false positives and some negative errors(?) as each rendering of the images could be different from machine to machine. 

I can look at Percy a bit more this week.

For now - let's merge this in - with the end game of off-loading this to a CI task

PR #434, opened 2019-06-24
WIP: the one where we add percy.io, remove BackStopJS and get it as part of the CI build

sturobson:
- [x] remove BackStop.js
- [x] install Percy
- [x] setup `.percy.yml` file
- [x] get it working as part of the CI build


sturobson:
I cannot get the `ignore files` bit to work - but it's no real problem. 

The last update _should_ get travis kicking it into touch -- although it might need the auth token :-/

khawkins98:
Trying to run `npx percy snapshot build/components/preview/` i get

 ›   Warning: Skipping visual tests. PERCY_TOKEN was not provided.

I suppose that's expected as this is meant to run at the Travis level, do you need help adding that token?

sturobson:
I need to work on this - to get Percy to test the snapshots against develop - not the branch itself.

Or, @khawkins98 - should we be testing this against master? And should we only set Percy to test between develop and master? I guess that's for the future - as there's a lot of things using the develop CSS live at the moment?

khawkins98:
Given that we'll only get 5,000 snapshots a month, maybe we should only test develop against master.

That would probably mean we'd need to step up our merges to master (and releases?), as once we have a desirable visual change, we'd want to merge it into master to stop false positives in the visual regression testing.

☝️ Which all sounds fine to me.

sturobson:
Cool, I guess we could put BackStopJS back in to the system to work locally. But I think that might be too much - as we'd really only be expecting either a weird breaking change somewhere - or something we have decided to and agreed upon changing.

khawkins98:
I'd be in favour of seeing backstopJS removed from the vf-core main dependencies. It's quite a hefty install.

sturobson:
let's do it as planned then. Remove BackStopJS, and make Percy only test on `master` any changes (I think we decided that)

khawkins98:
👍

PR #438, opened 2019-06-27
enhancement: dont-link-breadcrumbs-without-url

khawkins98:
In embl-breadcrumbs-lookup, the current desire is that when we don't have a working url, to not link the term.


sturobson:
Nasty, couldn't we add this in the template to `if harUrl { }` (pseudo code).

Let's add it in - but in the knowledge this isn't the best solution in Stu's opinion.

khawkins98:
Yeah, I'd really like all crumbs to have a url and I was hoping that the `#hash_links` would push folks on. 

Maybe instead we could link to the in-development https://dev.beta.embl.org/info/coming-soon/

PR #440, opened 2019-06-27
the one where we include the first pass at a card component

sturobson:
there's still behind the scenes stuff to do, plus documentation. But I think we're ready for usage.


khawkins98:
It'd be really nice to have the "correct" njk template, but that probably doesn't make sense yet as this is likely to change a lot. I guess this will eventually feel smooter once we have proper pre-alpha/alpha/beta/live patterns. 

Maybe a task for next week is to update versioning across the components.

sturobson:
Ah - yes -- I completely forgot to separate them. I don't think I'll have capacity to do this until Monday.

PR #441, opened 2019-06-27
the one that fixes quite a lot of small issues with vf-core and vf-wp

sturobson:
tidies up the breadcrumbs so they don't need a grid
spacing stuff
other bits


khawkins98:
Pushed one commit to update where vf-breadcrumbs was used within vf-core.

khawkins98:
Ugh, forgot to push those commits. New PR coming.

PR #444, opened 2019-07-01
enhancement: use-heading--l-alt-for-some-patterns

khawkins98:
Follows up on #443 by subing out `heading--l` for `heading--l-alt` in:

- vf-card
- vf-section-header
- vf-summary
- vf-video-teaser

Basically, where a bump from 30px to 24px was undesirable.

We may have a larger refactor of the naming of some of these tokens coming, but for now this "makes things look right", i think.

### DIFF
--- a/components/vf-card/vf-card.scss
+++ b/components/vf-card/vf-card.scss
@@ -93,7 +93,7 @@
 }
 
 .vf-card__title {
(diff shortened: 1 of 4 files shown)


sturobson:
let's get this one in - and move towards #445

Issue #412, opened 2019-06-14
remove vf-scss-lint task from the gulp watch task of gulp vf-dev


sturobson:
I think we should leave linting the CSS until we wish to push the branch up -- as it can slow down initial development as you're waiting for the linter to finish before the CSS gets injected in the page

khawkins98:
Sounds fair, taking this one.

PR #454, opened 2019-07-02
Enhancement: gulp css building, linting

khawkins98:
Addresses #412 and also address the "spirit" of that ticket, by:

- Making the css linting come after the css recompilation
- Adds a soft-fail css linting task that won't explode the local deve experience
- Fixes `vf-css` to use a better callback approach so that:
    1. You get an accurate idea of the css build time
    2. A real signal that the vf-css pipe stream has completed
    3. No longer get a flash of fractal before the css build has finished

And it should be a bit faster now that fractal and all the other gulp tasks won't try to run while the Sass is still building.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -103,7 +103,7 @@ const sassPaths = [
   path.resolve(__dirname, 'node_modules'),
 ];
 
-gulp.task('vf-css', function() {
+gulp.task('vf-css', function(done) {
   const sassOpts = {
     // Import sass files
     // We'll check to see if the file exists before passing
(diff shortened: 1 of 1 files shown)


khawkins98:
Looks like it drops the travis gulp build time from ~29s to ~25s 🐎

sturobson:
As put in #457 I don't think we need to include style lint when developing as it makes having browser sync pretty much useless … I would rather remove it for development and leave in pre-push

khawkins98:
I don't see any speed difference on my machine with the linting and I wouldn't think it should block the css build. That said I'm a 13 MBP and I think you're on a bit of an old-ish macbook... so could imagine the linting is gumming things up. 

I'll drop the sass lint from the watch command on this PR, but I'd still like to get this one in as it the Sass compilation here runs a lot more smoothly without churning on top of the other tasks.

sturobson:
So, for me:

without linting - the gulp task compiles the changes and then injects them into the page - instant changes when developing components.
with linting - the gulp task compiles the changes, then runs style lint, and if that passes it injects them into the pag - it takes time enough for me to remove it manually at the moment.

sturobson:
we could try `parallel` rather than `series` for this - with whatever you've done to make it 'soft fail'.

That should probably speed it up because the tasks will be done … in parallel and not one after the other (in series).

khawkins98:
Weird. I wonder why the experience is so different. Maybe it's something we can address when we're next in person together.

Really the linting task will be most important for developers new to the VF, so for now the use case we need to work on is "speed for Stu" over "linting for devs that don't get it" 😆

PR #457, opened 2019-07-02
the one where remove stylelint from dev and save lots of time

sturobson:
having `stylelint` process every time a Sass file changes means that it takes an age to see your changes whilst working away on a component or feature.

This PR removes it from the Sass-y `vf-watch` task so it is only tested as a pre-push commit


sturobson:
This is now addressed back in #454 - so closing.

Issue #364, opened 2019-04-19
Using frctl `path` with njk templates


khawkins98:
Now that we're passing data in from the Yaml files, it presents an issue with file assets.

Issue: from the Yaml we can't pass `  imageUrl: "{{ '/assets/vf-figure/assets/figure-example.png' | path }}"` without affecting the Njk template. 

So we're passing fixed paths in #363, that works fine on local but won't work once deployed. 

One way around is to utilise the "real" template in some other way, like the README.md:

```
{% render "@vf-figure", {
  text: "Version, 1982, Adenovirus with 217 dots",
  alttext: "hello alt text",
  imageUrl: '/assets/vf-figure/assets/figure-example.png' | path
} %}
```
However the `| path` filter doesn't properly work there :cry: 

As a temporary work around we could pass in assets deployed on remote URLs, but we'll eventually need to sort this.

sturobson:
> As a temporary work around we could pass in assets deployed on remote URLs, but we'll eventually need to sort this.

I always thought it would be nice to use 'real content' from Bynder for the component library

I knew I'd break things with this

khawkins98:
#365 has addressed usage within the component's `.njk` files, however we still have issues in the README.md. 

For example doing:`{% render '@vf-figure'  %}` fails to link the image, returning a path of `assets/vf-figure/assets/figure-example.png`. 

That's because the request context isn't correctly bubbling up to the Fractal NJK plugin. 

Doing `{{ '/assets/vf-figure/assets/figure-example.png' | path}}` in the README.md also has the same issue. 

I dropped a PR (https://github.com/frctl/nunjucks/pull/13) that would resolve the later of the two problems, but the request context is still getting lost when we `{% render '@vf-figure'  %}`, however a workaround would be to pass the path at the README.md level:

- `{% render '@vf-figure', {imageUrl: '/assets/vf-figure/assets/figure-example.png' | path } %}`

But that would only work if the PR gets merged in.

sturobson:
is `{% render %} supposed to work in README.md files?

I know this - https://fractal.build/guide/documentation/dynamic-docs.html#using-bundled-helpers - is handlebars but I'd've expected the same limited availability.

khawkins98:
When you set the docs to use nunjucks it uses the same for the notes (README.md); 

```
    fractal.docs.set('ext', '.njk');
    fractal.docs.engine(nunj); 
```

sturobson:
[...]

PR #458, opened 2019-07-02
docs: Using frctl `path` with njk templates

khawkins98:
Addresses #364


sturobson:
@khawkins98 I wonder if this failed because cloud flare was down?

Feel free to merge it in if it's safe to do so.

khawkins98:
Re-ran the tasks and it checks out. Must have been cloudfare outage.

PR #464, opened 2019-07-03
Refactor: trim fractal previews

khawkins98:
Removes the `_preview--grids` and `_preview--containers` as they were effectively redundant. 

May be more cleanup we can/should do as discussed in #128

If we get the boilerplates out we can also drop `_preview--fullhtml`


sturobson:
I've got some other thoughts on creating a one size fits all preview that might actually help inform the container grid needed to be used for a set of components (or it might not, but it'll help preview)

Let's get this in and I'll run some ideas around and make a draft PR to look at..

PR #465, opened 2019-07-03
bug: vf-conent-hub-html grid-escape

khawkins98:
The below tweak introduced some layout issues as non-layout affecting content out of the content hub (think: blocks) was being escaped from the grid.

```
.embl-content-hub-html {
  grid-column: 1 / -1;
}
```

This fixes that by only targeting elements that are direct children of `-grid`.

It also adds `.embl-content-hub-html` so we can begin to update classes in the contentHub.


khawkins98:
You can see the issue this would resolve at https://dev.beta.embl.org/topics/drug-discovery/

PR #466, opened 2019-07-04
the one where we update all of the design tokens

sturobson:
This PR updates quite a few of the existing design tokens used. Primarily because `heading--l-alt` changed sizes and was sprinkled in other places. Changing its size to be different to `heading--l` means that it doesn't really make semantic sense.

I'll add comments to the changes depending on what's changed on each commit.

This PR attempts to make sure the changes cascade across all file types within the component library. Including any uses of `vf-utility-classes` in `.hbs` examples


khawkins98:
design machine

sturobson:
The commit that updates all the button, body, and heading sizes includes the following changes. They are changed in the design tokens, which cascades to the Sass maps, variables, CSS custom properties, and utility classes where they're used.

| original token  | new token  |
|---------------|--------------------|
| `heading--xl`  | `text--heading-1` |
| `heading--l`  | `text--heading-2` |
| `heading--l-alt`  | `text--heading-3` |
| `heading--r`  | `text--heading-4` |
| `heading--s`  | `text--heading-5` |
|   |   |
| `body--xl`  | `text--body-1` |
| `body--l`  | `text--body-2` |
| `body--r`  | `text--body-3` |
| `body--s-alt`  | `text--body-4` |
| `body--s`  | `text--body-5` |
| `body--r`  | `text--body-6` |
|   |   |
| `button--r`  | `text--button-1` |
| `button--r`  | `text--button-2` |

sturobson:
At the moment the colour order for the greys is out of whack with the actual order of dark to light.

We change around `vf-color--grey--light` and `vf-color--grey--lightest`.

This cascades from the design tokens to the Sass maps, variables, CSS custom properties, and also the utility classes.

Where the utility classes have been used and currently in place - they will need to be swapped around to make it look the same still.

sturobson:
updated the t-shirt sizing a little bit:

| original token  | new token  |
|---------------|--------------------|
| --s | --sm |
| --r | --md |
| --l | --lg |

sturobson:
well… I think I've fixed everything ¯\_(ツ)_/¯

khawkins98:
Pushed a few minor fixes I found. One question, in `vf-heading` should we change the size names to be consistent? 

Right now:

```
.vf-text--heading-s {
  @include set-type(text-heading--5);
}
.vf-text--heading-r {
  @include set-type(text-heading--4);
}
.vf-text--heading-alt {
  @include set-type(text-heading--3);
}
.vf-text--heading-l {
  @include set-type(text-heading--2);
}
.vf-text--heading-xl {
  @include set-type(text-heading--1);
}
```

sturobson:
I knew I'd miss something -- we should do that I think.

I know it looks 'odd' with 1 being the largest and 5 being the smallest - but it's a convention that others are using for similar things.

khawkins98:
👍

khawkins98:
Just pushed one more commit to update the text/headline njk templates to reference the new number-based system. Button was also similar but I added a new "size" property to the approach as previously the style was both text size and look, which now use different class name approaches (`.vf-button--rounded` vs `.vf-button--pill`)

sturobson:
Gah! 

Let's hope this is (apart from the media queries changes for `beta`) the last time we look at tokens for some time.

Or we look at trying to see how we can mitigate all these misses (by me) in the future :D

khawkins98:
Ok, should be all set. Approving, will leave it to you for one last check and to merge.

PR #468, opened 2019-07-04
Docs: move design token demonstration

khawkins98:
As I work through cleaning up docs and making "things more approachable", it became clear that the design tokens can (of course) change in both how they're used and how much technical detail should be shown.

With that in mind I've moved the existing `/docs/styles/` to be paired directly with `vf-design-tokens`

I think this:

1. Makes it clear that these settings come from `vf-design-tokens`
2. Are where developers might expect the most technical values
3. We can link to from `/docs`
4. Makes it possible to put something more high level in `docs/styles` (think: "how do i use colours/type/spacing with the VF")
5. It also stops the risk of "what happens when there's `docs/styles` but `vf-design-tokens` wasn't installed?


sturobson:
hmmm, I'm not so sure.

I'm thinking that _really_ `vf-design-tokens` should be part of `tools` and they're installed into `vf-core` as a dependency … this does spring up the question of 'should they be in `vf-core` documentation at all'? Which is part of the bigger 'what should be the `vf-core` developer library - and what should be the larger documentation.

I guess we can merge this in and come back to it when we find some … time :D

sturobson:
and I've just realised I need to pull this branch into my feature branch if it gets pulled into `develop`

khawkins98:
Cool, could see that as well. This PR will at least get them inside the component which will help separate concerns and make moving things around a bit easier.

PR #469, opened 2019-07-04
Chore: set percy to run only on develop

khawkins98:
This limits the percy command to only run on commits to develop and compares against master.

We might also want to disable the github integration to stop comparissons on PRs, which has chewed through our alotment of screenshots.

This also set it only do mobile/desktop for a 33% savings :D

But it doesn't really matter until August as we've already gone through our quota!

### DIFF
--- a/.percy.yml
+++ b/.percy.yml
@@ -1,6 +1,7 @@
 version: 1
 snapshot:
-  widths: [375, 800, 1330]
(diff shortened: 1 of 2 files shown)


khawkins98:
🤞 that i've gotten the travis ci config right, awaiting outcome of that.

PR #472, opened 2019-07-08
adds vertical spacing for vf-lede, normalises spacing for vf-intor

### LINKED ISSUE #471: BUG - vf-intro spacing is seemingly wonky
The spacing on `vf-intro` seems off. 

Maybe too much on `vf-intro__heading vf-intro__heading--has-tag` and missing altogether on `vf-lede`.

image

Demo: https://dev.beta.embl.org/guidelines/visual-framework/dev-docs/components/detail/vf-intro.html

sturobson:
should fix #471

### DIFF
--- a/components/vf-lede/vf-lede.scss
+++ b/components/vf-lede/vf-lede.scss
@@ -10,7 +10,7 @@
 @import 'vf-lede.variables.scss';
 
 .vf-lede {
-  @include set-type(body--xl, $custom-margin-bottom: 2rem);
+  @include set-type(body--xl, $custom-margin-bottom: 0);
 
   & > a {
     @include inline-link;
--- a/components/vf-lede/vf-lede.scss
+++ b/components/vf-lede/vf-lede.scss
@@ -10,7 +10,7 @@
 @import 'vf-lede.variables.scss';
 
 .vf-lede {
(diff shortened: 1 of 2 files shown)


components/vf-lede/vf-lede.scss
```diff
@@ -10,7 +10,7 @@
 @import 'vf-lede.variables.scss';
 
 .vf-lede {
-  @include set-type(body--xl, $custom-margin-bottom: 0);
+  @include set-type(body--xl, $custom-margin-bottom: 2rem);
```
khawkins98: I the reason this was set to `0` was for use on the group pages. Does this need to be a mixin, or a new `.vf-intro .vf-lede` added?

image

sturobson: hmmm, overall vertical spacing still needs a look at - perhaps we should move this to the `vf-intro` component and have it add the margin there. 

I'm not sure if it needs to mixin.

Once we get more pages built out - we can start seeing how the vertical spacing of all the things fits (or doesn't).

khawkins98: Moving the tweak to `vf-intro` sounds like a cleaner mid-term approach.

And +1 to revisiting spacing all around.

PR #476, opened 2019-07-08
RELEASES - alpha.7, to master

khawkins98:
And now we go from develop to master. 

Once our visual regression testing is back up, this is when we'd probably want to look for "oh noes! 😭"


khawkins98:
Running with admin privs again for this chore and to save Stu a bit of a hassle.

PR #477, opened 2019-07-09
ENHANCEMENT - ship, enhance component generator

khawkins98:
Does two things:

1. Ships to npm: It will be easier to create local design system components (think: `randomservice-component-idea`) if we ship the component generator along with `vf-core`
2. Picks up the project namespace for making the component  … That is, if you're in an embl environment it gives you an `embl-` option ... EBI gives you `ebi-`, Service X gives `servicex-`

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -4,7 +4,6 @@ bin
 .travis.yml
 CHANGELOG.md
 CONTRIBUTING.md
-tools/component-generator
 public
 build
 .max-age
--- a/tools/component-generator/index.js
+++ b/tools/component-generator/index.js
@@ -5,8 +5,9 @@ var path = require('path');
 var config = require(path.resolve('.','package.json'));
 
 config.vfConfig = config.vfConfig || [];
-global.vfName = config.vfConfig.vfName || "Visual Framework 2.0";
-global.vfNamespace = config.vfConfig.vfNamespace || "vf-";
+vfName = config.vfConfig.vfName || "Visual Framework 2.0";
+vfNamespace = config.vfConfig.vfNamespace || "vf-";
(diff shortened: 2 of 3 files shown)


khawkins98:
I also had the thought that the component generator might become its own component (`vf-component-generator`), as it would:
- make it a modular install
- allow other component generators for bespoke needs (embl-componemt-generator)
- devote it from the vf-core releases

But none of those are pressing

sturobson:
I've a couple of reservations over this.

To me, this is more of a 'tool' or a 'utility' rather than a 'component' (more on this when I finally share a much larger documentation on nomenclature). This is a minor point, and something that can be 'changed' in the future (if we all agree on the new doc.). So it's not a `blocker` for this PR.

It definitely should be a thing on NPM although I'm not convinced we should be advocating for other teams to go off and create their own components 'willy nilly'. Of course, they could do that _anyway_. But I don't think it would be good for `service-x-button` to be a thing that's wildly different than `vf-button` without some communications between both teams to start with. 

We should be offering / doing something [like this](https://coggle.it/diagram/V0hkiP976OIbGpy8/t/vanilla-pattern) for any 'new' components. 


Let's merge this in for now and come back to it.

khawkins98:
Cool. This PR still keeps the component generator as part of vf-core, but agree it might be useful to have it as a separate npm install. 

Ideally we'll get the VF used by parts of the org (or even outside of the org) that have exoctic needs, so I expect there will be things like `ensembl-species-browser` that would be so ensembl-specific they'll never make it back to the vf-core repo.

khawkins98:
aside: that's a nice little governance doc we can reference when things will/won't be added to vf-core.

PR #488, opened 2019-07-17
CHORE: update-footer

khawkins98:
Makes the vf-footer pattern generic. 

Why?

- This also sorts the current situation where we have to update the contentHub and here
- Lets the contentHub be just a template
- De-conflicts the having embl-specific stuff in a generic vf pattern

This change won't break anything.


khawkins98:
I suspect this is going to be hard to use as a programmatic template, needing a variable list length and number of columns. And then perhaps various "other use cases"?

Probably worthwhile to use the yaml as a good best practice/demo, but bigger fish in need of frying and all that.

PR #490, opened 2019-07-17
BUG - vf-global-header site link, misc related issues

khawkins98:
This allows the `vf-global-header` site title to be a link.

Currently if using `vf-link`:

image

With PR:

image

Notes/other things that were bubbling up in vf-global-header:

- Utilises the otherwise unused `$vf-global-header__link-text--color` 
- Logo width was still set for EMBL logo
- Adds `spaceless` to remove some extra whitespace that was in the header

### DIFF
--- a/components/vf-global-header/vf-global-header.config.yml
+++ b/components/vf-global-header/vf-global-header.config.yml
@@ -3,5 +3,5 @@ label: Global Header
 preview: '@preview--nogrid'
 context:
   component-type: block
-
+  href: JavaScript:Void(0);
   text: Visual Framework for Life Science websites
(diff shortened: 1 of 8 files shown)


sturobson:
I think that the logo should hold the text. As well as a potential hover 'issue' that would be the small gap between the logo and the text, a screen reader will be getting duplicates.

`vf-link` imo is becoming more of a burden, not just here.

khawkins98:
So you mean something like a `vf-logo--with-text` variant?

sturobson:
If anything I think we should be able to consolidate `vf-logo` and `vf-logo--responsive`. As it's Scalable we should not have to have 3 different files for the `.svg` file.

Leaving us with `vf-logo` which has a `<span>` inside it. 

This span can have `vf-sr-only` or not depending on the requirements. 

I've done a quick branch of how I see it - https://github.com/visual-framework/vf-core/tree/refactor/vf-logo

khawkins98:
this one is eclipsed by #545, closing.

PR #493, opened 2019-07-18
the one where we remove a dependency to aid portability

sturobson:
this PR removes the normalize node module

we add `margin: 0;` as needed to `vf-body` but, for the time being also to `body` as currently not everything has `vf-body` on the `body`.

For the beta - we need to work on the documentation page for `vf-body`. 

At least this _should_ mean there's one less dependency for people using vf-core and it means that our patterns (except for `vf-body` (up unto this point)) work regardless of any 'normalise' or 'reset' CSS added.


khawkins98:
Looks good to me, but will leave to you to merge as I cleaned up the merge conflicts and removed a couple npm dependencies. Probably not off-piste, but not sure if maybe there were other plans for those modules. 

ski

Issue #489, opened 2019-07-17
BUG - Forms component is broken


khawkins98:
Looks like something has broken on `vf-form`

https://dev.beta.embl.org/guidelines/visual-framework/dev-docs/components/detail/vf-form__core.html

## 1. The label floats, well, don't:

![image](https://user-images.githubusercontent.com/928100/61369690-f58e2000-a890-11e9-9a85-6b2ecf03bd7b.png)

The [float-labels.js](https://pryley.github.io/float-labels.js/) seems to still be there, but is either not being invoked and/or needs some css to control the display. 

Not sure if there's been much progress on CSS-only approaches: https://medium.com/@eugeniolujambio/accessible-css-only-floating-labels-74bf9173bc5f ... if so, IE 11 could probably live with a fallback approach.

## 2. The radio buttons seem to have broke:

![image](https://user-images.githubusercontent.com/928100/61369778-266e5500-a891-11e9-813d-66edd9967c34.png)

khawkins98:
On the radio buttons, looks like it lost settings for border width/style; adding this back seems to fix it:
```css
.vf-form__radio + .vf-form__label::before {
    border-width: 3px;
    border-style: solid;
}
```

![image](https://user-images.githubusercontent.com/928100/61370407-a34dfe80-a892-11e9-8551-ada227bf0991.png)

sturobson:
radio buttons are now fixed in #497 

todo: the JS interaction of things for the input slide up whizz bang

PR #497, opened 2019-07-18
the one where we get the form elements to work

sturobson:
this should address #489

### DIFF
--- a/components/vf-form/vf-form__radio/vf-form__radio.scss
+++ b/components/vf-form/vf-form__radio/vf-form__radio.scss
@@ -14,7 +14,7 @@
   // Box.
   & + .vf-form__label::before {
     background: set-ui-color(vf-ui-color--white);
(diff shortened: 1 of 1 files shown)


khawkins98:
Should we go ahead and merge this so we have radios for alpha.8, and then we can fix the JS in another PR?

sturobson:
I think so -- I managed to get the JS to work locally with a little hack - but it doesn't seem to 'work' as I would expect it to anyhow … so I'm wondering if this either 

a) never got finisied
b) got finished but somewhere we(I) managed to bring in some older 'work in progress' for it.

I'll scour the Git history - but I think for now we can merge in so 'it 'works'

khawkins98:
We probably need to revise however it was working anyways, now that we're favouring only working on things with `data-vf-js-form-labels` (or whatever the attribute would be called).

PR #500, opened 2019-07-20
the one where publications have a short summary component added

sturobson:
This adds the `vf-summary--publication` component as well as a new helper class of `vf-et-al`.

The `vf-summary--publication` can be nested inside a `vf-box` where it takes the `vf-box` colours.

If the `vf-summary__author` list is truncated to a certain number of authors you will need to add `vf-et-al` to the `<p>` - `<p class="vf-summary__author | vf-et-al` - for it to add *et al* to the end of the list.

This _should_ cover all three use cases for publication summaries that are [here](https://gitlab.ebi.ac.uk/emblorg/backlog/issues/225). [here](https://gitlab.ebi.ac.uk/emblorg/backlog/issues/226), and [here](https://gitlab.ebi.ac.uk/emblorg/backlog/issues/227)

### DIFF
--- a/components/vf-summary/vf-summary.scss
+++ b/components/vf-summary/vf-summary.scss
@@ -12,6 +12,30 @@
     margin-bottom: 0;
   }
 
+  // vf-summary can also sit inside a vf-box - so we need to adjust some things to match
+  .vf-box & {
+    .vf-summary__title {
+      color: inherit;
+    }
+    .vf-summary__link {
+      color: inherit;
+    }
+    .vf-summary__date {
+      color: inherit;
+    }
+    .vf-summary__text {
+      color: inherit;
+    }
+    .vf-summary__source {
+      color: inherit;
+    }
+    .vf-summary__author {
+      color: inherit;
+    }
+    .vf-link {
+      color: inherit;
+    }
+  }
 }
 
 .vf-summary__button {
@@ -38,6 +62,13 @@
   color: set-color($vf-summary__date-color--text);
 }
 
+.vf-summary__source {
+  @include set-type($vf-summary__date-typography);
+
+  color: set-color($vf-summary__date-color--text);
+  font-style: italic;
+}
+
 .vf-summary__link {
   @include inline-link();
 
@@ -62,4 +93,11 @@
 
 .vf-summary__title {
   @include set-type($vf-summary__title-typography);
+}
+
+.vf-summary__author {
+  @include set-type($vf-summary__author-typography);
+
+  color: set-color($vf-summary__author-color--text);
+  font-style: italic;
 }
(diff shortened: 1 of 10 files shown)


sturobson:
I'm going to add what i've put in this PR description into the documentation.

gulpfile.js
```diff
@@ -171,7 +171,6 @@ gulp.task('vf-css', function(done) {
             return 'Problem at file: ' + error.message;
           })
         )
-        .pipe(autoprefixer(autoprefixerOptions))
```
khawkins98: Was this intentional? I know we've discussed that dropping autoprefixer would likely be a good optimisation, but it's still in the package.json dependencies.

components/vf-summary/vf-summary.scss
```diff
@@ -38,6 +62,13 @@
   color: set-color($vf-summary__date-color--text);
 }
 
+.vf-summary__source {
+  @include set-type($vf-summary__date-typography);
+
+  color: set-color($vf-summary__date-color--text);
+  font-style: italic;
```
khawkins98: Feels weird using italic as we don't anywhere else, but I suppose that's a comment to direct to the design pattern/Mark.

components/vf-summary/vf-summary.scss
```diff
@@ -12,6 +12,30 @@
     margin-bottom: 0;
   }
 
+  // vf-summary can also sit inside a vf-box - so we need to adjust some things to match
```
khawkins98: Quite a few things might sit inside vf-box and i expect we'll have other things that might sit inside other things with background colours. When we have time, would be nice to have a method to support some kind of `vf-give-it-contrast` method.

components/vf-summary/vf-summary.scss
sturobson: Which I asked already ;)

components/vf-summary/vf-summary.scss
sturobson: Id rather work on something that either fixes this programmatically or adds specificity than sling yet another class onto something.

gulpfile.js
sturobson: No. Unfortunately you can swap a PR to a draft PR. Which this still is.

sturobson: I'll put this back in and push

components/vf-summary/vf-summary.scss
sturobson: I think we can go with this now 'as is' and come back to it -- it'd be nice to look at component 'theming' in more detail for the next beta

components/vf-summary/vf-summary.scss
sturobson: checking this in Firefox using the fonts panel in the dev tools -- this is using the correct IBM Plex Sans Italic font file

sturobson:
@khawkins98 I think we're good to do a first push of this now?

khawkins98:
👍 Approved. Will leave you to merge when ready.

PR #501, opened 2019-07-23
the one that fixes misspellings we haven't caught yet

sturobson:
Whilst working on #497 I noticed several CSS errors in Firefox's console.

This PR rectify's them.

- missing `$` in variables
- declarations that don't work in the ruleset because of the other declarations
- other misspellings


khawkins98:
nice. 

mr. clean

PR #502, opened 2019-07-23
Enhancement/split gulpfile

khawkins98:
Addresses #324. 

I still need to make sure this doesn't blow up vf-eleventy (it shouldn't).

## Also...

Some things I found that should probably be done in a separate PR for beta.1 or beta.2:

1. The vfConfig at the top of the gulpfile.js could probably be a JS module, or even its own npm package
2. I've found several npm packages that don't seem to be in use. 

```js
// todo we might be able to remove these package.json Dependencies,
// they don't seem to be needed
// const replace = require('gulp-replace');
// const postcss     = require('gulp-postcss');
// const reporter    = require('postcss-reporter');
// const syntax_scss     = require('postcss-scss');
// const concat = require('gulp-concat');
// const uglify = require('gulp-uglify');
```

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -24,14 +24,14 @@
 // all settings are optional
 // todo: this could/should become a JS module
 const fs = require('fs');
+const path = require('path');
 const config = JSON.parse(fs.readFileSync('./package.json'));
 config.vfConfig = config.vfConfig || [];
 global.vfName = config.vfConfig.vfName || "Visual Framework 2.0";
 global.vfNamespace = config.vfConfig.vfNamespace || "vf-";
-global.vfComponentPath = config.vfConfig.vfComponentPath || __dirname + '/components';
+global.vfComponentPath = config.vfConfig.vfComponentPath || path.resolve('.', __dirname + '/components');
 global.vfBuildDestination = config.vfConfig.vfBuildDestination || __dirname + '/temp/build-files';
 global.vfThemePath = config.vfConfig.vfThemePath || './tools/vf-frctl-theme';
-const path = require('path');
 const componentPath = path.resolve('.', global.vfComponentPath);
 const buildDestionation = path.resolve('.', global.vfBuildDestination);
 
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -31,6 +31,7 @@ const buildDestionation = path.resolve('.', global.vfBuildDestination);
 // -----------------------------------------------------------------------------
 
 const gulp = require('gulp');
+const watch = require('gulp-watch');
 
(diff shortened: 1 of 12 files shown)


khawkins98:
Looks like there are several other packages in package.json that could go, I'll have go at that tomorrow afternoon -- hopefully after this one's ready/merged. Always satisfying to drop some dependencies.

sturobson:
So - moving things to JS modules -- I've got no idea what they are … Is it worthwhile moving things away from 'standard' gulp file stuff (as I (perhaps wrongly) know it to be)? Could this negate potential new users of the systems in helping improvement in the future?

khawkins98:
Committed a couple more things to get this to play nice with 11ty (and general extensibility) and fix a few edge cases.

---

On modules: There's not really much to JS modules. More or less they're just fancy functions ... receives parameters and returns something. 

Upshot is that instead of redefining each npm package in each child js file, we can just reuse it. Makes the code more efficient and allows passing state -- and avoids having to use `global` all over the place, there are few globals that i've previously used that I can hopefully trim out now. 

It doesn't change the gulp syntax, just adds a wrapper -- so I don't think it will add a terrible amount of friction (beyond the — as you've already noted — "what's a JS module?"). 

Also enables other future things if we need them; a couple of posts on the technique I found quickly: 
- http://macr.ae/article/splitting-gulpfile-multiple-files.html#approach-two
- https://www.sparkbit.pl/building-multi-module-projects-gulp/

This will also be really helpful in using `vf-core` as a dependency in other projects.

sturobson:
> "On modules: There's not really much to JS modules. More or less they're just fancy functions ... receives parameters and returns something."

> "It doesn't change the gulp syntax, just adds a wrapper -- so I don't think it will add a terrible amount of friction (beyond the — as you've already noted — "what's a JS module?")."

👍

> "This will also be really helpful in using vf-core as a dependency in other projects."

I've been 'wondering' about this - and wonder if we should be providing separate gulpfiles per 'project' or have one thing to rule them all -- I guess now we're splitting things out it will make a one-size-fits-all easier to manage as we can separate dev/prod tasks down for what they're actually used for.

khawkins98:
> I've been 'wondering' about this - and wonder if we should be providing separate gulpfiles per 'project' or have one thing to rule them all -- I guess now we're splitting things out it will make a one-size-fits-all easier to manage as we can separate dev/prod tasks down for what they're actually used for.

Yeah, likewise. This approach should already get us a lot closer to being able to just import the gulpfile partial we need for something.

khawkins98:
This one should be ready for more review/merge now.

khawkins98:
chop

PR #503, opened 2019-07-24
the one where we add a moving label

sturobson:
this adds the floating label to form inputs … note the need for additional classes

I think between beta.1 and beta.2 we/me should concentrate on documentation a little


sturobson:
This should close #489

PR #504, opened 2019-07-24
CHORE - vf-forms linting

khawkins98:
A linting, code approach thing to namespace and use data attributes for the vf-forms js functionality.

### DIFF
--- a/components/vf-form/vf-form__input/vf-form__input.scss
+++ b/components/vf-form/vf-form__input/vf-form__input.scss
@@ -154,13 +154,12 @@
   opacity: 0;
   position: absolute;
   top: 0;
-  transition-delay: 100ms;
   transition: opacity 120ms, visibility 120ms;
+  transition-delay: 100ms;
   visibility: hidden;
   white-space: nowrap;
   width: auto;
 }
-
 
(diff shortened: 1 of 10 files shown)


khawkins98:
There will be more to do here on the error messaging, but this does most of what's needed.

components/vf-form/vf-form__input/vf-form__input.scss
```diff
@@ -139,37 +139,11 @@
...
+.vf-form__item.vf-form__floatlabel {
   position: relative;
 }
-.floatLabel {
+.vf-form__floatlabel {
   .vf-form__label {
     position: absolute;
   }
 }
-
```
sturobson: where's this gone?

khawkins98: Dropped this bit as it wasn't actually being used -- or is there a plan to use it?

sturobson: ah, that's true -- liberal copying and pasting happened here :D

khawkins98: hoarding :D

Issue #483, opened 2019-07-11
ENHANCEMENT - move Autoprefixer, minification to build only


sturobson:
Currently Autoprefixer adds:

- about 250ms to build times
- 50kb to the file size of CSS

Most of this is (probably) down to `2 last versions` as that includes Blackberry OS 7.

We should look to either

a) move it to build time only
b) remove it completely and wait for the bug reports

sturobson:
this might be easier now we've got split out gulp tasks?

We'd probably need to pass an `ENV` variable so that if we're in build it pulls in the autoprefixer task

khawkins98:
As per conversation on #481; we can use the same logic to also run the minification on deploy only. 

From #481:
- [ ] Disable minification on local builds (~600ms)
- [ ] Investigate need for/optimising prefixing (~210ms)

PR #509, opened 2019-07-30
the one where we put autoprefixer and cssnano into the build task

sturobson:
This should complete #483.

Rather than get bogged down in working with Node Environments and as what we're wanting to do can be done after Sass has compiled, I have created a new task `vf-css-build` which takes the `styles.css` file, runs `autoprefixer` and then saves it, then runs `cssnano` and saves it with a `.min.css` suffix. 

This seems to have cut down `vf-css` for me by quite some margin (for my crawling 2016 MacBook).

I've also adapted `vf-build` so that we make use of `gulp.parallel` so certain tasks can go off and 'do their thing' alongside others which should speed the processes up a little bit.

### DIFF
--- a/tools/gulp-tasks/_gulp_rollup.js
+++ b/tools/gulp-tasks/_gulp_rollup.js
@@ -31,8 +31,13 @@ module.exports = function(gulp, path, componentPath, buildDestionation) {
   ));
 
   // Build as a static site for CI
-  gulp.task('vf-build', gulp.series(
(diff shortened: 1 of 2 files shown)


khawkins98:
Looks good. Copied some notes from the PR as a little bit of inline docs.

sturobson:
seems I forgot to push the version where I've removed the commented out `autoprefixer` and `cssnano` from `vf-css` :(

Issue #467, opened 2019-07-04
vf-code-example, and vf-font-plex-mono


sturobson:
With the changes to the design tokens #466 there's only a real need for one monospace font-size and weight. 

Currently as part of #466 we have reduced the amount of options there are in the Sass map to use this font.

I think we need to make some changes and below lists them out as a checklist.

- [x] change `text-body--5` to `text-code--1` (currently this is breaking the map, therefore:)
- [ ] change how this token is generated to make a Sass variable rather than a map.
- [ ] look at how to incorporate the specific font file as part of `vf-code-example`.

khawkins98:
Pretty sure we've done this, largely in #511. Being bold and closing.

@sturobson If i'm off, just re-open.

PR #511, opened 2019-07-30
the one where we update our use of plex mono

sturobson:
This PR is related to #467. So far we have:

- [x] renamed the usage of plex in the design tokens and `vf-code-example` from `5` to `1`.
- [x] make it a Sass variable rather than a map
- [x] reduced the amount of font files needed.


sturobson:
regarding: 

> make it a Sass variable rather than a map

Because a font being used in a component comes with `font-size`, `font-weight`, and `line-height`it is probably better to keep it as a Sass map of one thing.

Other wise we would need to refactor the `@mixin` that we're using for typography to cope when there's no map and only variables. I don't think we've really go a use case here that necessitates such a larger refactor.

sturobson:
@khawkins98 as you worked on `vf-font-plex-mono` can you look into removing the typefaces we don't need? 

We only need the `regular`/`400` files now

khawkins98:
Have removed the unneeded mono assests, but ...

The Sass size mapping isn't working as expected due to a bug in [`components/vf-sass-config/mixins/_typography.scss`](https://github.com/visual-framework/vf-core/blob/develop/components/vf-sass-config/mixins/_typography.scss): 

https://github.com/visual-framework/vf-core/blob/b09afd050e393c56cbeeadb86da77602a7e44101/components/vf-sass-config/mixins/_typography.scss#L2-L9

☝️ The font map is hardcoded to `$vf-font--sans-serif`, which means `set-type(text-body--1,$vf-font-family--monospace, $custom-margin-bottom: 0);` results in huge text as it references the sans font map:

image

I didn't spot an easy/clean way to fix this...

sturobson:
I'll take a look at this, to fix - then we should be able to add it into `beta2`

sturobson:
This _should_ be good now.

khawkins98:
Meanwhile inside the sass innards

Aside: can't believe I even manage to find a linotype gif.

Issue #19, opened 2018-11-08
FEATURE - Add z-index map and function


sturobson:
So far we have used z-index in three places:

[vf-form__input](https://github.com/visual-framework/vf-core/blob/develop/components/elements/vf-form/vf-form__input/vf-form__input.scss#L32)
[vf-masthead](https://github.com/visual-framework/vf-core/blob/develop/components/blocks/vf-masthead/vf-masthead.scss#L84)
[vf-data-protecion-banner](https://github.com/visual-framework/vf-core/blob/pattern/data-protection/components/blocks/vf-data-protection-banner/vf-data-protection-banner.scss#L27)

These are based on [my novel way of using z-index](http://www.alwaystwisted.com/articles/2014-07-22-talkin-bout-z-index) but I think we should move towards something that makes use of Sass, it's variables, functions, and maps.

Something [like this](https://www.sitepoint.com/better-solution-managing-z-index-sass/)

So that we call the z-index with a function that references the map of variables.

I guess for now, we only need to look at the ones we currently have, and we can add and adapt as the component library fills out a little more.

Thoughts?

khawkins98:
Might be worth revisiting this in the next few weeks, noticed a first z-index issue on the microsites:

https://wwwdev.ebi.ac.uk/about/teams/protein-function-development/

![zindex](https://user-images.githubusercontent.com/928100/60335452-43082300-999e-11e9-9c14-d44fdb6a2a10.gif)

sturobson:
Cool - it can be something that is a 'design token' - even though this only really effects web things. I could get that into this weeks 'design tokens sprint'.

sturobson:
… although in this css I would suggest that the z-index of the banner needs bumping up - which can be a `hack.scss` addition to `vf-wp`.

sturobson:
regarding [your comment](https://github.com/visual-framework/vf-core/issues/19#issuecomment-506682724) I think this should be something that's added to the 'hack' file for `vf-wp` rather than something that's added in `vf-core` to fix

khawkins98:
This one close-able now?

PR #512, opened 2019-07-30
the one where we tokenise the z-index

sturobson:
this PR addresses most of the things in #19:

- creates a new design token file and Sass map for `z-index`
- creates a `set-layer` Sass function 
- replaces hardcoded `z-index` in component Sass files with the `set-layer` function
- removes unneeded `z-index` declarations that were not doing anything.

### DIFF
--- a/components/vf-summary/vf-summary--profile.scss
+++ b/components/vf-summary/vf-summary--profile.scss
@@ -42,8 +42,6 @@
   --vf-text-margin--bottom: 24px;
 
   @include set-type($vf-summary__body-typography);
-
-  z-index: $vf-summary__link--z-index;
 }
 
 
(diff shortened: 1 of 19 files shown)


components/vf-summary/vf-summary--profile.scss
```diff
@@ -42,8 +42,6 @@
   --vf-text-margin--bottom: 24px;
 
   @include set-type($vf-summary__body-typography);
```
khawkins98: Removing z-index from vf-summary variants and vf-masthead doesn't seem to hurt anything, despite my best efforts using the updated CSS on vf-wp and some pages on dev.beta.embl.org 

¯\_(ツ)_/¯

khawkins98:
Have approved, but will leave you to merge in case there's anything else that's popped to mind.

components/vf-summary/vf-summary--profile.scss
sturobson: I'm wondering if I was being over-zealous in my thinking, or it was a left over of me trying to get the interactions required working.

PR #515, opened 2019-07-30
updates color and spacing for navigation, and color for content

sturobson:
I've fixed this - an erroneous inclusion of the wrong colour from `vf-ui-color-map` and some spacing issues … 
Screenshot 2019-07-30 at 15 12 59

I've made changes to the main navigation so it's centered too … it shouldn't blow anything up.


khawkins98:
Stu, I think you still had a surprise bug to fix here, or is this actually just waiting on me to approve?

PR #516, opened 2019-07-30
BUG - vf-build asset copy task misses js

khawkins98:
This resolves a bug that was exposed by #509, where `gulp vf-fractal:build` was responsible for asset copying. 

That was the wrong spot to do so — and in 509 the bug was exposed by using `parallel` for `vf-fractal:build`, it was resulting the JS assets being missed from deployment. 

Note that https://dev.assets.emblstatic.net/vf/develop/scripts/scripts.js is currently a 403 (missing and denied)

### DIFF
--- a/components/embl-content-hub-loader/embl-content-hub-loader__fetch.js
+++ b/components/embl-content-hub-loader/embl-content-hub-loader__fetch.js
@@ -187,7 +187,7 @@ function emblContentHubFetch() {
     var dateRemainingList = document.querySelector('.'+emblContentHubGenerateID(position)).querySelectorAll('.date-days-remaining');
     var todayDate = new Date();
     if (dateRemainingList.length > 0) {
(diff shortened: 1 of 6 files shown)


components/embl-content-hub-loader/embl-content-hub-loader__fetch.js
```diff
@@ -187,7 +187,7 @@ function emblContentHubFetch() {
     var dateRemainingList = document.querySelector('.'+emblContentHubGenerateID(position)).querySelectorAll('.date-days-remaining');
     var todayDate = new Date();
     if (dateRemainingList.length > 0) {
-      for (dateRemainingIndex = 0; dateRemainingIndex < dateRemainingList.length; dateRemainingIndex++) {
+      for (let dateRemainingIndex = 0; dateRemainingIndex < dateRemainingList.length; dateRemainingIndex++) {
```
khawkins98: Oops, I didn't mean to include this JS fix here, but it currently causes an error too.

PR #517, opened 2019-08-01
Bug: Handle optional js dependencies

khawkins98:
Optional JS dependencies weren't working correctly. The `embl-conditional-edit` isn't required by `embl-content-hub-loader`, however by not including it as a `import`, it wasn't working. And if we did `import` it, it would break on sites when `embl-conditional-edit` wasn't included.

This solution works by:

1. We always list optional dependencies in parent JS as an `import`
2. If the JS isn't found, the gulp task strips out the compiled and failed `import` (which requireJS treats as non-browser compatible `require()`

This also sorts a minor variable name bug and check.


sturobson:
@khawkins98 left this for you to merge :)

khawkins98:
🛒👍

PR #520, opened 2019-08-02
Bug: check for succesful content import was in wrong spot

khawkins98:
Was checking for success before `targetLink` was checked.


khawkins98:
Stu's out today, so using admin rights to get this simple fix in.

Issue #524, opened 2019-08-07
FEATURE - Output core, component versions into CSS


khawkins98:
When looking at the compiled CSS (sample: https://visual-framework.github.io/vf-welcome/css/styles.css) it can be harder to track why something is happening as you have to backtrack to the source project to see the version of components or vf-core in use.

It should be _relatively_ trivial to add the version numbers in, something like:

```css

/*
 *
 * styles.css
 * The Visual Framework kitchen sink of styles.
 * Include this style sheet as a quick way to get *everything*,
 * or follow the guide at github.com/visual-framework/vf-core
 * to learn how to build custom and optimised CSS.
 *
 * This CSS was compiled by vf-component-rollup 1.x.x using vf-core-v2.0.0-beta.2
 *
 */

```

We can split this into two PRs:

To do:
- [x] add to component templates #525
- [x] add to existing components (to be done)

khawkins98:
Have a proof of concept in #525

sturobson:
@khawkins98 liking this. A couple of questions / thoughts.

- if this is still part of Sass we would need  `/*! loud comments */`in case we minimise the `.css` file (which we should be doing).
- I wonder if we should have this comment per component compiled (in the non `.min.css` file?.

khawkins98:
👍 

If I follow, the PR should be doing it in that manner. 

My brain is thoroughly fried now after dealing with an issue on gulp being slow to scan the filesystem 😭

khawkins98:
Once #525 is finalised/merged in we can go through and update existing components as a separate chore PR.

sturobson:
This is what I was meaning with the CSS comments --

https://github.com/visual-framework/vf-core/pull/525/files#r313750421

like I've said in the comment - our gulp settings might be doing the work for us - but we'd need to check that it's doing it for the `.min.css` -- which should be easy to do

PR #525, opened 2019-08-07
Feature: verbose output of component version

khawkins98:
For #524

This:
- Generates a per-component (and .gitignore-d) `package.variables.scss`
- Outputs a comment in compiled CSS:
```css
 * Component: @visual-framework/vf-componenet-rollup
 * Version: 1.0.0-beta.3
 * Location: components/vf-componenet-rollup
```

### DIFF
--- a/components/vf-componenet-rollup/index.scss
+++ b/components/vf-componenet-rollup/index.scss
@@ -6,8 +6,8 @@
  * Include this style sheet as a quick way to get *everything*,
  * or follow the guide at github.com/visual-framework/vf-core
  * to learn how to build custom and optimised CSS.
- *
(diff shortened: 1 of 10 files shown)


components/vf-componenet-rollup/index.scss
```diff
@@ -1,3 +1,4 @@
+@import 'package.variables.scss';
 /*
```
sturobson: when we start to use the `.min.css` this needs to be written as `/!*` for Sass to not remove it on compiling. AFAIK.

Actually this may be negated by gulp settings

khawkins98: Ah, ok follow now. 

We do need the loud comments, pushing a change shortly to:

image

PR #530, opened 2019-08-19
the one where we look again at forms

sturobson:
note: this is a draft PR.

We are adding form validation that gracefully degrades using bouncer.js. This PR also changes some components markup so they're easier to include using nunjucks

todo: 
- [ ] SVG icons (which is a bigger thing, so I'll look to see how I can mitigate code changes when adding whatever here
- [ ] the `bouncer.JS` needs to be added correctly (one I'm leaving for @khawkins98)
- [ ] finish of refactoring the form components so they're more modular and nunjucks-y


khawkins98:
Is it worth merging this one in and then making issue (s) to pick up on the rest?

sturobson:
Yeah - it's the JS for the errors that I'm still not set on - I think it 'works' but we're tied to how a library wants us to include things - so the SVG icon needs to be hard encoded plus other things

I think it's in a better place than it was - but we/me should ratify requirements to get it 'tip top' in an issue to pick up.

khawkins98:
Sounds like a plan.

PR #539, opened 2019-08-30
Bug/load vf-core version directly from `node_modules/vf-core`

khawkins98:
This wasn't directly impacting vf-core, but is a config issue in sites that use vf-core as a dependency. In short, it resulted in the vf-core version being reported as the client version.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -24,13 +24,14 @@ const path = require('path');
 // todo: this could/should become a JS module
 const fs = require('fs');
 const config = JSON.parse(fs.readFileSync('./package.json'));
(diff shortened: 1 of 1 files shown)


sturobson:
Do we need to put this `config` kinda stuff in a separate file away from `gulpfile.js`too? 

Now, soon, somewhen in the future?

khawkins98:
Yeah, def need to make it re-usable/componentised. Still trying to get a feel for how we should best do it; currently debating:

1. Just another gulpfile include
2. Some kind of separate dependency

It's probably #1, I just need to get on it. Would be good to have that done before the demo next week.

PR #540, opened 2019-08-30
Bug: vf-code example formatting

khawkins98:
Two things:

- Bug: fix nesting of code>pre …
   Should be `<pre><code>`: https://html.spec.whatwg.org/multipage/grouping-content.html#the-pre-element
  Doing it the other way around is typically fine, but Chrome tries to flip the elements which leads to odd behaviour.

- Bug: use grid layout on pre tag not code …
  This is partly related to doing `<pre><code>` but in some cases affected the layout of code snippets, breaking each element onto its own line.

### DIFF
--- a/components/vf-code-example/README.md
+++ b/components/vf-code-example/README.md
@@ -2,6 +2,8 @@
 
 ## About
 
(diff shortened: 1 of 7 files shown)


sturobson:
leaving you to merge :)

PR #551, opened 2019-09-05
bug: sass maps in css custom properties

khawkins98:
the sass map wasn't being executed, but there must be a better way to do this?

### DIFF
--- a/components/vf-logo/vf-logo.scss
+++ b/components/vf-logo/vf-logo.scss
@@ -9,24 +9,23 @@
 }
 
 .vf-logo__image {
-  --vf-logo-margin-right: map-get($vf-spacing-map, vf-spacing--sm);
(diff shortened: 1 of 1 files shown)


sturobson:
> "...there must be a better way to do this?"

Perhaps with a function, but I'm conflicted to do this as we're burying away something into Sass that's just a DX issue.

khawkins98:
Works for me, probably doesn't bother overthinking yet.

sturobson:
We've also got some [design tokens as CSS custom properties](https://github.com/visual-framework/vf-core/tree/develop/components/vf-design-tokens/dist/sass/custom-properties) already 

So rather than doing:

`--vf-logo-margin-right: #{map-get($vf-spacing-map, vf-spacing--sm)}` 

we could do:

`--vf-logo-margin-right: var(--vf-spacing--sm);`

PR #545, opened 2019-09-05
Chore/embl logo update

khawkins98:
This uses the latest logos that tabe sent around and turned into a bit more than first anticipated:

- Brings in the suggestions from (BUG - vf-global-header site link, misc related issues #490) and your branch
- Drops the PNG logos (no need for them)

PR #555, opened 2019-09-06
Bug: logo fixes

khawkins98:
Follow on fixes to #545 and #548 around classnames and leaving `vf-logo--responsive` as deprecated to not break things.

### DIFF
--- a/components/embl-logo/embl-logo.scss
+++ b/components/embl-logo/embl-logo.scss
@@ -12,13 +12,13 @@
 }
 
 @media (min-width: 500px) {
-  .embl-logo--responsive {
+  .embl-logo {
     background-image: url('../assets/embl-logo/assets/logo-medium.svg');
     height: 28px;
   }
 }
 @media (min-width: 999px) {
-  .embl-logo--responsive {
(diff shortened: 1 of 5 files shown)


components/embl-logo/embl-logo.scss
```diff
@@ -12,13 +12,13 @@
 }
 
 @media (min-width: 500px) {
-  .embl-logo--responsive {
+  .embl-logo {
     background-image: url('../assets/embl-logo/assets/logo-medium.svg');
```
sturobson: Are we swapping out the SVG because of the rendering of the text?

What about moving the media queries into the SVG? So there's only one file if that's the case?

khawkins98: Different SVGs show a different number of dots, and I think some manual spacing adjustments between dots.

I've not done mediaqueries in an SVG before, but I think that would mostly meen aggregating all three SVGs in one final and then displaying them conditionally ... so shouldn't be hard? My only real concern would be updating any minor tweaks from the design team ...

Maybe we add an issue to do this once it's more "stable"?

Issue #546, opened 2019-09-05
DOCUMENTATION: Sass Utilities & Mixins & Variables


sturobson:
Now that we have exposed the Sass Utilities in the component library (under Containers) and after sitting in on the preview of the workshop yesterday (04/09/19) I think we need do document the all the Sass things so that developers using `vf-core` to create things can see what options there are available.

For example, exposing the global Sass variables for spacing could show developers that we have a spacing system and aren't really keen on deviating from that. This could also apply for when a developer attempts to create a new component and can see that there is a mixin available to create the typography.

khawkins98:
Further addressed in #638 — is it worth keeping this open, or are we pretty aware that we need to keep on with the documentation effort?

sturobson:
let's close it. We know we (me) need to do it

PR #556, opened 2019-09-09
Documentation: Improve vf-sass-config documentation

khawkins98:
This follows up on #550 and improves documentation and resolves an issues with broken links appearing in the Utilities section on the VF Core index of components.

This should mostly address #546

Also kills `_vf-variables.shame.scss` which we seem to no longer need 🎉

### DIFF
--- a/components/vf-componenet-rollup/index.scss
+++ b/components/vf-componenet-rollup/index.scss
@@ -16,7 +16,6 @@
 @import 'vf-global-custom-properties.scss';
 
 @import 'vf-global-variables.scss';
-@import '_vf-variables.shame.scss';
 @import 'vf-global-variables.scss';
 @import 'vf-border-radius.variables.scss';
 @import 'vf-breakpoints.variables.scss';
@@ -163,3 +162,6 @@ html, button {
 // This is a demonstration of vf-core's ability to warn and proceed on missing
 // sass imports
 @import 'vf-somepattern/vf-i-dont-exist.scss';
+
+// If you have any locoal overrides, put them in:
+@import 'vf-local-overrides/vf-local-overrides.scss';
--- a/components/vf-componenet-rollup/index.scss
+++ b/components/vf-componenet-rollup/index.scss
@@ -16,7 +16,6 @@
 @import 'vf-global-custom-properties.scss';
 
 @import 'vf-global-variables.scss';
-@import 'vf-global-variables.scss';
 @import 'vf-border-radius.variables.scss';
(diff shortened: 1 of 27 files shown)


sturobson:
> "Also kills _vf-variables.shame.scss which we seem to no longer need 🎉"

components/vf-componenet-rollup/index.scss
```diff
@@ -16,7 +16,6 @@
 @import 'vf-global-custom-properties.scss';
 
 @import 'vf-global-variables.scss';
-@import '_vf-variables.shame.scss';
 @import 'vf-global-variables.scss';
```
sturobson: I cannot remember why we've got this in twice?

components/vf-sass-config/README.md
```diff
@@ -1,46 +1,168 @@
...
-- set-layer.scss
-- string-replace.scss
-- vf-functions.scss
+### set-color.scss
+
+ `set-color($color-name)`
+
+### map-deep-get.scss
+
+`map-deep-get($map, $keys`
```
sturobson: missing `)` ?

components/vf-sass-config/mixins/_blockquote.scss
```diff
@@ -1,3 +1,6 @@
+// Reusable styling for html blockquote elements
+// @include blockquote;
+
 @mixin blockquote() {
   @include set-type(text-body--2);
   margin: 0;
```
sturobson: could we include the Sass variable in the `set-type` mixin instead of `margin: 0` here?

I know it's slightly out of the PR's remit

components/vf-componenet-rollup/index.scss
khawkins98: Hrm, looks like merge conflict gone awry.

components/vf-sass-config/mixins/_blockquote.scss
khawkins98: Took a stab at this. Further in the weeds: I have an itch that the blockquote mixin should live within the `vf-blockquote` component, but that would make various Sass import difficulties 🙈

sturobson: Yeah, and if we did it for `vf-blockquote` we could be opening up a little can of worms –– "should the button mixin be with buttons‽" etc etc.

I like the separation as it gives the opportunity for people to use `vf-core` as a little mixin library, like you can do with Bootstrap.

PR #566, opened 2019-09-16
FEATURE: Make compiled CSS available as an "asset"

khawkins98:
This PR collectes the per-component `index.scss -> vf-component.css` as an asset to be made available alongside any SVGs, JPGs, etc.

image

Useful for debugging and for edge cases around including one CSS file.

### DIFF
--- a/tools/gulp-tasks/_gulp_rollup.js
+++ b/tools/gulp-tasks/_gulp_rollup.js
@@ -28,7 +28,7 @@ module.exports = function(gulp, path, componentPath, componentDirectories, build
   // -----------------------------------------------------------------------------
 
   gulp.task('vf-dev', gulp.series(
-    'vf-clean', 'vf-component-assets', ['vf-css', 'vf-scripts'], 'vf-fractal:start', ['vf-lint:scss-soft-fail', 'vf-watch']
+    'vf-clean', ['vf-css', 'vf-scripts'], 'vf-component-assets', 'vf-fractal:start', ['vf-lint:scss-soft-fail', 'vf-watch']
   ));
 
   gulp.task('vf-prepush-test', gulp.parallel(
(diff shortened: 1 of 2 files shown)


sturobson:
does this duplicate the CSS generation that's already happening or replace it and move it to `assets`, a problem with replacing is that the `.css` file is hidden from the component view on vf-core - because Fractal can't look inside folders for things :( 

Although in saying that - we're not generating the CSS files for when the component library is built? Or it's no longer exposed in Fractal's viewer?

khawkins98:
It leverages the CSS generation that's already happening; that is:

- First: `vf-css:generate-component-css` runs
    - (this command is only run on deploy/`gulp vf-build`)
- Then: this gulp command copies that compiled css to the build folder

The compiled CSS is also available/deployed at `./build/components/raw/embl-logo/embl-logo.css`, _however_ the components assets aren't available underneath that `./build/components/raw/` path:

image

Which in turn means paths are broken. 

Copying the compiled CSS into assets allows us to see if the "links to things are working". Alternatively we could copy the assets down into the fractal `./build/components/raw/..`.

Thinking out loud, that doesn't seem as good as it would be trickier to achieve, would leave us with two copies of all the assets, and wouldn't be available for debugging on the CDN (the `./assets` directory gets deployed but the `./components` directory doesn't)

khawkins98:
The use case this is scratching:

1. "`vf-logo` isn't showing an image, is that because of how the Sass is being compiled or is the component's path wrong?" This gives a "CSS of truth" on the CDN. 
2. "I just need one component". This might be if someone is asked to include `embl-data-protection-banner` and can just use `embl-data-protection-banner.css`

sturobson:
> ""I just need one component". This might be if someone is asked to include embl-data-protection-banner and can just use embl-data-protection-banner.css" 

Which would save some copying and pasting from a browser into an IDE.

Also - I'm guessing this could be some sort of benefit for React-y webpack-y things too.

Great stuff

PR #568, opened 2019-09-18
the one where we update the vf-button component

sturobson:
This PR works on a few things in the `vf-button` component:

- fixes the secondary colours not coming through
- removes background colour on hover of `vf-button--outline` variants
- adds a 150ms transition on hover state so it's a little smoother
- changes the `vf-text-button--1` to `vf-button--sm` etc as ¯\_(ツ)_/¯ why `vf-text-button--1` was there

ancillary things

- removes the `default` button from the component library view - as it's not really a thing
- adds some margin to the top and bottom of components when using the default preview
- makes `vf-tabs` fullwidth with its border as it looks 'odd' on a page that's not `vf-inlay`ed.

### DIFF
--- a/components/vf-button/vf-button.scss
+++ b/components/vf-button/vf-button.scss
@@ -91,3 +91,47 @@
 
   padding: 16px;
 }
+
+// Begin deprecated classes and styles
+html:not(.vf-disable-deprecated) {
+
+  // Deprecated in 1.0.0-beta.7
+  // Will be removed in vf-button v2.0
+  .vf-text-button--secondary {
+    @include button-link($vf-link--color: set-color(vf-color--green), $vf-link--hover-color: set-color(vf-color--green--dark));
+  
+    box-shadow: 0px 6px 0px 0px set-color(vf-color--green--dark);
+    margin-bottom: 6px; // because we're using box-shadow we need to 'create the space' again
+  
+    &:hover {
+      box-shadow: 0px 2px 0px 0px set-color(vf-color--green--dark);
+    }
+  }
+
+  // Deprecated in 1.0.0-beta.7
+  // Will be removed in vf-button v2.0
+  .vf--button--outline.vf-text--button--secondary {
+    @include button-link--ghost( $vf-link--color: set-color(vf-color--green) );
+  }
+
+  // Deprecated in 1.0.0-beta.7
+  // Will be removed in vf-button v2.0
+  .vf-text-button--2 {
+    @include set-type(text-button--2);
+    padding: 8px;
+  
+    &.vf-text-button--rounded {
+      border-radius: 8px;
+    }
+  }
+
+  // Deprecated in 1.0.0-beta.7
+  // Will be removed in vf-button v2.0
+  .vf-text-button--1 {
+    @include set-type(text-button--1);
+  
+    padding: 16px;
+  }
+  
+}
+// End deprecated
--- a/components/vf-button/vf-button.scss
+++ b/components/vf-button/vf-button.scss
@@ -59,11 +59,11 @@
   }
 
   &.vf-button--secondary {
-    @include button-link--ghost( $vf-link--color: set-color(vf-color--green) )
+    @include button-link--ghost( $vf-link--color: set-color(vf-color--green) );
   }
 
   &.vf-button--tertiary {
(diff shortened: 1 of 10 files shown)


components/vf-tabs/vf-tabs.scss
```diff
@@ -13,6 +13,31 @@ $vf-tabs__list--active-border--color: set-ui-color(vf-ui-color--white);
   border-bottom: 1px solid $vf-tabs__list-border--color;
   margin: 0;
   padding: 0 0 0px 16px;
+  position: relative;
+
+  // @media (min-width: $vf-breakpoint--xl) {
```
khawkins98: From reading your PR description, it sounds like this shouldn't be commented out?

If you did intend to use it, one thing: the tab line could flow into other content. Example: if `vf-tabs` is inside a `.vf-grid__col-2`. I suppose there might be other times, like an EMBL Grid container, but that's the only one that comes to mind; screenshot:

image

khawkins98: Maybe the easiest way around is to do `vf-tabs vf-tabs--full-width` or smthn.

components/vf-button/vf-button.scss
```diff
@@ -55,12 +58,12 @@
     @include button-link--ghost;
   }
 
-  &.vf-text-button--secondary {
-    @include button-link--ghost($vf-link--color: set-color(vf-color--green), $vf-link--hover-color: set-color(vf-color--green--dark));
+  &.vf-button--secondary {
```
khawkins98: Good change, and `vf-button-sm`, obviously will require some changes in consumers of the VF. We should bubble up to the changelog. _That awkward phase when things aren't stable but are in use._

khawkins98: Thinking more on this. Yesterday I felt like the only action we could do would be to stick a note in the CHANGELOG.md, which would be good, but likely go unseen by most. Especially as it's not a major version change. 

I think this is a good chance to use the `vf-deprecated` approach. PR to the PR coming.

components/vf-tabs/vf-tabs.scss
sturobson: I commented this out - as it breaks the overflow of the page. It'd be neat to have - but it's probably more it's own little PR

components/vf-button/vf-button.scss
sturobson: I don't think it's deprecated in the same sense as we've been using 'deprecated'.

I agree a changeling in itself is not enough. I've spoken w/ David about it - and he's not using it anywhere yet.

Who else would be using it?

I have another idea - but I'll raise an issue for it - as it's something that's probably worth commenting on.

sturobson:
see also - https://twitter.com/StuRobson/status/1174590746361499648

components/vf-button/vf-button.scss
khawkins98: The EBI devs might be using it on the EBI training project, but it won't be an issue there until they update their components. 

There might also be a page here or there that's using `vf-text-button--secondary`

The trickier one is the [data protection banner](https://dev.beta.embl.org/api/v1/pattern.html?filter-content-type=article&filter-id=581&pattern=node-body&source=contenthub) that's served from the contentHub. Once the JS runs it renders a `vf-button vf-text-button--secondary` — and after this PR it will doe a `vf-button vf-button--secondary`.

And if a site is using an older version of vf-button the button will look wrong.

That particular issue could be worked around by making vf-banner also use the old class, but that feels messy? and would be a less general fix.

sturobson: how `vf-text-button--secondary` got into this I cannot remember.

But it's really inconsistent, don't you think? 

From searching the codebase, it feels like (to me) a find/replace that was done - and this got missed as a bad outcome.

khawkins98: oh yeah, for sure it must have been some find-replace gone mad. will be very happy to see it go — did you see the PR I made, that keeps your change and keeps `vf-text-button--secondary` for deprecated support.

sturobson: yep, left for you to merge - what needs to be done w/ this PR -- close it?

khawkins98: Cool. Merged my PR into this one, and now this one gets merged into dev.

khawkins98:
You want to merge, just in case anything else comes to mind?

PR #569, opened 2019-09-18
updates to the Fractal / VF Theme

sturobson:
I procrastinated a little after before lunch … here's a little tidy up of the current theme.

Perhaps it just needs tweaks here and there - rather than removing it completely. 

I think we can agree that dogfooding, as you've mentioned, is good - and having this theme can quickly highlight where we're missing something in … design or documentation.

### DIFF
--- a/tools/vf-frctl-theme/views/layouts/skeleton.njk
+++ b/tools/vf-frctl-theme/views/layouts/skeleton.njk
@@ -19,7 +19,7 @@
   </script>
   {% include 'partials/stylesheets.njk' %}
 </head>
-<body>
+<body class="vf-body">
 {% block page %}{% endblock %}
 {% include 'partials/foot.njk' %}
 </body>
--- a/tools/vf-frctl-theme/dist/css/default.css
+++ b/tools/vf-frctl-theme/dist/css/default.css
@@ -9,5 +9,21 @@
   border: 1px solid #b3b3b3;
   padding: 0; }
 .Frame-body {
-  grid-column: 2 / -2;
+  grid-column: 1 / -1;
+}
+
+.Pen { grid-column: 1 / -1; }
+.Pen .vf-page-header__heading { grid-column: 2 / -2; }
+.Pen .component-info { grid-column: 2 / -2; }
+.Pen .vf-tabs { grid-column: 2 / -2; }
+.Preview { grid-column: 1 / -1; }
+.preview-component-list { grid-column: 2 / -2; }
+.Preview-wrapper .vf-figure__caption {
+  max-width: 87.5em;
+  margin: 0 auto;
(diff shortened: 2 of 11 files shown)


tools/vf-frctl-theme/views/layouts/skeleton.njk
```diff
@@ -19,7 +19,7 @@
   </script>
   {% include 'partials/stylesheets.njk' %}
 </head>
-<body>
+<body class="vf-body">
```
khawkins98: Moving `vf-body` up to the body element "traps" some of the grid and container layouts; example: /components/detail/vf-grid-page

Maybe conditionally using `vf-u-grid--reset` somewhere is a way around it.

sturobson: what would you say is in correct about `/components/detail/vf-activity-group` I'm guessing the lack of full-width colour block.

I think this goes back to an old idea that started to get implemented with `vf-inlay`s `vf-masthead` so that we use `:before` and `:after` to do the full width bits. 

Also, I think the grids aren't _really_ components but 'styles' as several other design systems categorise them (along with some our simple element components). With Design Tokens as their own thing.

I'm writing a little proposal about this - so I'll work on getting that to all

khawkins98: Just a note that the other wide components also get impacted, like vf-header vf-masthead and vf-footer

sturobson: Yep :)

sturobson:
see also -- https://twitter.com/StuRobson/status/1174590746361499648

sturobson:
Screen Shot 2019-09-19 at 10 14 15

changes to the HTML and CSS to the theme:

- fixes 'full width component viewer'
- makes the home page look as expected because of Fractals re-use of classes and lack of differential between 'pages' to hook into.

khawkins98:
Feels strange using vf-body multiple times inside itself and then resetting in with app-level css. That said, we've got no shortage of other more important things to sort out.

sturobson:
> "Feels strange using vf-body multiple times"

Yep, in the theme it's now only used once, apart from a component view (for obvious reasons).

sturobson:
well, kinda only once -- I'll take another swing.

There was a reason I plumped for `vf-body` not `vf-container` … it's just unfortunately got put in as and where it's thought to've been needed -- perhaps it should have been an #id :sob:

sturobson:
I'm feeling fruity. so I'm going to merge this in now. 

There's an stale branch on my local with regards to grids that I want to take a look at once this is in.

khawkins98:
froot loops

PR #572, opened 2019-09-19
Feature/verbose output component versions

khawkins98:
Adds verbose output of component version to all remaining components. This follows up on #525 and is the second half of #524.


khawkins98:
Resolved. i'm sure it will be fine. 

cards

PR #574, opened 2019-09-25
Add a "fractal data" build mode

khawkins98:
This repurposes the unused "VRT" mode as a "dataobject" mode. This has Fractal do its bootstrap to generate the javascript object of components but doesn't announce the webserver or generate static HTML files ... it makes for a faster static build time with things liek vf-eleventy

### DIFF
--- a/fractal.js
+++ b/fractal.js
@@ -144,6 +144,7 @@ module.exports = {
       });
     }
 
+    // If you want to build static html files
     if (mode == 'build') {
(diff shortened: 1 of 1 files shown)


khawkins98:
Retrospective now that beta.4 is out: Drops build time in vf-elelvent projects by  considerable amount (depending on the number of components available), but generally saves ~20 seconds: https://travis-ci.org/visual-framework/vf-welcome/builds

PR #578, opened 2019-09-25
Refactor/vf button

sturobson:
this should update the look of buttons and fix all current use cases


components/vf-button/vf-button.scss
```diff
@@ -213,3 +219,16 @@ html:not(.vf-disable-deprecated) {
 
 }
 // End deprecated
+
+
+// HACK
```
sturobson: this is what needs to go into the `vf-wp` CSS file -- although there's a potential issue w/ CSS custom properties not crossing CSS stylesheets.

khawkins98: Should we comment it out here — that way we don't accidently get used to its effects on non vf-wp sites?

khawkins98:
Only other thing I've noted is: you can get a jitter if you hover over the top of the button.

components/vf-button/vf-button.scss
sturobson: if we comment it out here - the border won't work in groups -- because `vf-wp` main.css comes before styles.css … so the custom properties don't cascade.

Unless I'm not understanding your question?

sturobson:
> "Only other thing I've noted is: you can get a jitter if you hover over the top of the button."

I noticed this and put it down to what happens when elements move on the page.

If this gets raised a lot - I could look at it more :)

components/vf-button/vf-button.scss
khawkins98: Then maybe we can just add the custom properties to the vf-wp workaround CSS file?

khawkins98:
> If this gets raised a lot - I could look at it more :)

I think it's more than fair to leave it as something we can bother with once the wheels get squeaky enough.

components/vf-button/vf-button.scss
sturobson: We could do this - and be sure to update whenever needed. 

It's better to make 'fixes' for platforms using the `vf-core` in the platforms own CSS file at the end of the day.

And when we get to any kind of 'theming' in future `vf-wp` -- perhaps getting `vf-wp` to `npm instal`l the design tokens could be a good thing, moving forward (but not for now)

PR #581, opened 2019-09-28
ENHANCEMENT - Require vf-sass-config, vf-design-tokens

khawkins98:
You can't do anything useful with vf-core without these two components, so it saves a couple of `yarn add`s and removes a common miss-configuration step. -- For developing vf-core, this won't mean anything practical, however it does help when using it as a dependency.

Also .npmignore-ing yarn.lock makes vf-core a lighter dependency by reducing the number of duplicate similar-version npm packages install, and this ensures that the required vf components are minimums, that is: default to vf-sass-config@v1.0.1, but newer is ok too

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -10,3 +10,4 @@ build
 .gitlab-ci.yml
 temp
 backstop_data
+yarn.lock
--- a/package.json
+++ b/package.json
@@ -11,6 +11,8 @@
   "dependencies": {
     "@frctl/fractal": "^1.2.0-beta.3",
     "@frctl/nunjucks": "^2.0.1",
+    "@visual-framework/vf-design-tokens": "^1.0.0-beta.6",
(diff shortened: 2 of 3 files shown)


khawkins98:
Gave this a little more thought and realised that it might be short sighted to require `vf-design-tokens` and `vf-sass-config` if an org might need to bring their own versions of colours; which prompts two related things:

1. We still need a way to address "primary" and secondary colours (as discussed previously)
2. One of the main things I was trying to address with this PR is how tedious it can be to add in ~30 components to make a site, maybe we could make "recipes" that would basically be quick lists to add in dependencies (i.e. `recipe-embl-corporate-site` might include some 30 basic components but skip `vf-logo`)

I still think it's useful to .npmignore the lockfile, but I'll make a seperate PR for that.

PR #582, opened 2019-09-30
ENHANCEMENT - Optimise npm dependencies

khawkins98:
As partly discussed in #569, this does two things to make `yarn add @visual-framework` better.

1. It moves several packages from `devDependencies` to `dependencies`. Without this we're also having to manually add in 20 vf-core dependencies in places like vf-eleventy
    - Example of configuration savings with this approach: https://github.com/embl-communications/design-system/pull/19/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2
2. It `.npmignore`s the lock file so that the client site can have more discretion on things like `"gulp-shell": "^0.7.1"` so we'll have a lighter install (common practice to not push the lockfile to npm; https://cdn.jsdelivr.net/npm/@frctl/fractal@1.2.0-beta.3/)


khawkins98:
Sticking a little retrospective on this: Reduces the number of dependencies by some 27 on client projects! https://github.com/visual-framework/vf-welcome/commit/006e6d9626ad1785cf84f2758f2a6cac7348d14c

PR #586, opened 2019-10-02
Fixes/hardcoded links

sturobson:
There were _several_ instances where we had hardcoded the `href` to either be empty `href=""` or contain a `#`. 

This PR intends to fix that by making all links use `.yml` data to populate. 

There's also a small fix to the `vf-pagination` component so that it's more a configurable template.


components/embl-logo/embl-logo.config.yml
```diff
@@ -16,6 +16,7 @@ context:
   component-type: embl-element
   href: http://www.embl.de
```
khawkins98: There's also an `href` in addition to `logo_href`, I'm guessing this accidental?

components/vf-pagination/vf-pagination.config.yml
```diff
@@ -1,8 +1,92 @@
 title: Pagination
 label: Pagination
-status: alpha
-variants:
-  - name: default
-    hidden: true
+status: beta
```
khawkins98: Pushed a commit to set the package.json to also be beta status.

components/embl-logo/embl-logo.config.yml
sturobson: this is 'technically' ok - as the `href` is the `.yml` not actually in the code as `href="#"`.

But this does bring up the lack of consistency with the key naming for this across components.

I think something like `component_link` or `component_href` would be better.

I'll refactor so they're all on the same hymn sheet.

khawkins98: What I was getting at is you've added a `logo_href`, I'm guessing you just forgot to delete the old one? Or is it used somewhere I'm not seeing?

sturobson: hmmm, now why wasn't that used before :sob:

I'll still go through and standardise the naming convention and do another push

khawkins98:
Cool. Probably good to take the time on this one as once we push any change will make David's life hard.

PR #594, opened 2019-10-08
Bug/fix font css import 2

khawkins98:
All about namespacing, we had a clash in varaible names too; follow up fix to #593


khawkins98:
Hrm, could have worked, I think. I'm not 100% on the behaviour of `default` in sass sometimes, it seems

PR #595, opened 2019-10-09
Enhancement: typography mixin

khawkins98:
This isn't done to change behaviour but to make the typography mixin a little easier to follow as it shifted directions a few times over all the alphas and betas and it was a little hard to follow what was happening when I was debugging yesterday.

### DIFF
--- a/components/vf-sass-config/mixins/_typography.scss
+++ b/components/vf-sass-config/mixins/_typography.scss
@@ -1,53 +1,53 @@
 // Generate correct font information when included into an element
 // Recommended to use with typography and sizing maps
-// @include set-type(text-body--3, $global-font-family, $custom-margin-bottom: vf-spacing--lg)
+// @include set-type(text-body--3, $font-family, $custom-margin-bottom: vf-spacing--lg)
 
 // $custom-margin-bottom: variable, auto, disable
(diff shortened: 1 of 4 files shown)


components/vf-sass-config/mixins/_typography.scss
```diff
@@ -1,53 +1,51 @@
 // Generate correct font information when included into an element
 // Recommended to use with typography and sizing maps
-// @include set-type(text-body--3, $global-font-family, $custom-margin-bottom: vf-spacing--lg)
+// @include set-type(text-body--3, $font-family, $custom-margin-bottom: vf-spacing--lg)
 
 // $custom-margin-bottom: variable, auto, disable
-@mixin set-type($font-size, $global-font-family: $global-font-family, $custom-margin-bottom: auto) {
-  @if $global-font-family != $vf-font-family--monospace {
```
sturobson: this was for that time when a team wanted to use a different typeface - but I guess that point is moot

khawkins98: Briefly thought "should we add the font map as an optional parameter to the mixin?" But then I thought it might be overcomplicating things ... at some point when there's time (hah!) would be good to see just how much breaks when someone tries to use their own design tokens ... also get s back to the nightmare refactor when we have a `$colour-primary`.

PR #596, opened 2019-10-11
The One Where We Add Tree Navigation V0.0.1

sturobson:
I'm pretty happy with the markup and naming conventions now.

Todo (but in another PR)
- JavaScript interactions
- CSS updates - as parts of it aren't 'there' yet (not full height, for example)
- documentation

But this should be enough to get the training pages integrating the HTML into WP etc etc


khawkins98:
Just adding a footnote that there's still some thinking about "where this lives" ... is it a `vf-navigation` variant, or is it just a tree. If it is only `vf-tree`, should it be `vf-tree--collapsible`?

_naming things_

PR #597, opened 2019-10-11
Component/vf tree enhancements

khawkins98:
This PR would need to be merged into the `component/vf-tree` branch.

This doesn't really change the behaviour but rather tries to improve consistency in some places, which was probably a byproduct of some of the obstacles in development. 

I've tried to break down the changes into chunks with commit comments to make things easier to follow; here's a quick link out:

1. Add placeholders for `collapsed` … a8fd47f
Although it does nothing, I think this makes the behaviour easier to understand.
1. A bit more of a programmatic naming convention 764beab
1. Respect the expanded state a8763c5
1. Improve demo-ability of YML … ce4c47c
Rather than saying `sublist: true`, add a stub list and set it to not expanded
1. Use only `expanded` … b66e9c1
Probably simpler than using `expanded` in some places and `collapsed` in others.  Maybe should also change `.vf-tree--collapsed`?
1. In place of `currentParent` and `currentChild` use `activeTrail` 5965652
1. Optimise NJK file … 5451c77
This is probably a pre-mature optimisation, but I think it demonstrates how we can make the code tidier with the standardising of the Yaml values.
1. Protodocumentation …8d4db14
Some notes that I had made as I was going through.

### DIFF
--- a/components/vf-tree/vf-tree.njk
+++ b/components/vf-tree/vf-tree.njk
@@ -1,3 +1,23 @@
+{% set vfTreeButton %}
+<button href="{{ tree_example_href }}" class="vf-button vf-tree__button">
+  <svg width="8" height="13" xmlns="http://www.w3.org/2000/svg"><g stroke="" stroke-width="1.6" fill="none" fill-rule="evenodd" stroke-linecap="square"><path d="M1.417 1.275L6 6.326M6 6.105l-4.583 5.61"/></g></svg>
+</button>
+{% endset %}
+
+{% macro vfTreeList(list, currentDepth=1) %}
+  <ul class="vf-tree__list {% if currentDepth > 1 %}vf-tree__list--additional{% endif %} vf-tree__list--{{ currentDepth }} | vf-list" aria-role="{% if currentDepth == 1 %}tree{% else %}group{% endif %}">
+  {% for item in list %}
+    <li class="vf-tree__item {%- if item.artiveTrail == true %} vf-tree__item--selected{% endif -%}{% if item.expanded == true %} | vf-tree__item--expanded{% endif -%}" aria-role="treeitem">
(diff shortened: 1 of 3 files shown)


components/vf-tree/vf-tree.njk
```diff
@@ -1,49 +1,31 @@
-<div class="vf-tree{% if collapsed %} vf-tree--collapsed{% endif %}" {% if collapsed %}data-vf-tree--collapsed="true"{% endif %}>
+{% set vfTreeButton %}
+<button href="{{ tree_example_href }}" class="vf-button vf-tree__button">
+  <svg width="8" height="13" xmlns="http://www.w3.org/2000/svg"><g stroke="" stroke-width="1.6" fill="none" fill-rule="evenodd" stroke-linecap="square"><path d="M1.417 1.275L6 6.326M6 6.105l-4.583 5.61"/></g></svg>
+</button>
+{% endset %}
+
+{% macro vfTreeList(list, currentDepth=1) %}
```
sturobson: I thought we weren't using `macro` - but that could have been the use case I had a while back (moving 'property logic' out of the `vf-button.njk` file :sob:

khawkins98: Is there a downside to using macro?

sturobson: I think it was more me having a separate file to abstract the logic more than using it.

khawkins98: Cool. To me this seems a good place to use a macro, given the repeated amount of code ... and in principle we could now support depths of 20 or more! 🙈

components/vf-tree/vf-tree.njk
```diff
@@ -1,49 +1,31 @@
-<div class="vf-tree{% if collapsed %} vf-tree--collapsed{% endif %}" {% if collapsed %}data-vf-tree--collapsed="true"{% endif %}>
```
sturobson: @khawkins98 where did this end up?

PR #599, opened 2019-10-15
Refactor/vf tree

sturobson:
This updates the HTML and CSS a little to make the component look as intended.

todo:

- JS interactions: (@khawkins98 is this something you can 'quickly' knock together
- Documentation: There's a few different classnames here and there and they need explaining so the they're used correctly so that the style of the component is as intended.

### DIFF
--- a/components/vf-grid-page/vf-grid-page.scss
+++ b/components/vf-grid-page/vf-grid-page.scss
@@ -24,13 +24,16 @@
 
 // Initial set up for browsers that do not support CSS grid layout
 html {
+  height: 100%;
   margin: 0 1.25em;
 }
 
 .vf-body {
   display: block;
+  height: 100%;
   margin: 0 auto;
   max-width: 76.5em;
+  min-height: 100%;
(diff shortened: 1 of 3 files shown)


sturobson:
CSS needs to have whatever 'remove EBI border' code around it.

khawkins98:
Ah, just saw this is a `draft`. Well, it's approved for whenever.

Happy to have a go at the JS .... I don't think it should be too tricky, which means I'm not thinking clearly 😆

sturobson:
> I don't think it should be too tricky

I think it'll ok - it's kinda like tabs, but not… and vertical

sturobson:
@khawkins98 do you want to merge this in then add the JS (and docs) as a secondary/tertiary PR - or do you want to add to this?

Getting this closed would at least have the CSS and updated HTML 'ready' for EBI to use

khawkins98:
Yeah, I'd say let's go with this and add on the rest as we can.

PR #600, opened 2019-10-16
Bug: Unset EBI link style on hover for links

khawkins98:
Fixes an issue as seen on the header here on hover http://ebiwd.gitdocs.ebi.ac.uk/ebi-academy-terrified/

It's a more general issue that's worth getting on all `a:hover`

This might all need a refactor once we make links not underlined 😭

### DIFF
--- a/components/ebi-vf1-integration/ebi-vf1-integration.scss
+++ b/components/ebi-vf1-integration/ebi-vf1-integration.scss
@@ -74,6 +74,7 @@ body.ebi-vf1-integration,
   
   a:hover {
     text-decoration: unset;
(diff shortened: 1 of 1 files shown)


khawkins98:
We've mostly used `unset` in the component so that's why I gravitated to that. This one feels a little strange, but afraid to think to deeply about it for fear of uncovering a rats nest.

sturobson:
here's a neat video on what `revert`,`unset`, and `initial` do - https://www.youtube.com/watch?v=GAjoVRmipcU

sticking with `unset` is probably the way to go for now.

PR #602, opened 2019-10-16
Bug: vf-button default colours

khawkins98:
The default colours weren't being processed by Sass -- meaning `class="vf-button"` was "invisible". 

I've had do interpolation (` #{set-color(vf-color--grey)};`) not sure if it warrants extrapolating into a Sass variable to just pass it to a CSS custom prop ...

Also slips in a fix for vf-tree.scss where `package.variables.scss` weren't referenced.


khawkins98:
Actually, have had thought. 2 secs

khawkins98:
Realised I should just reference the already set custom props: 

```css
:root {
  --default-bg-colour: var(--vf-color--grey);
  --default-bg-colour-dark: var(--vf-color--grey--darkest);
  --default-button--visited-border: var(--vf-color--grey--lightest);
}
```

As an aside, is `default-bg-colour` the right namespacing. Shouldn't this be `default-button--backgroun-colour`?

sturobson:
> "As an aside, is default-bg-colour the right namespacing. Shouldn't this be default-button--backgroun-colour?"

Not really.

> "Realised I should just reference the already set custom props:"

Thought this when reading your first comment

---

Also `vf-button` needs to be get it's own `naked` or similar variant to remove the design - I had to do a big `all: unset` for training because of it.

Buttons -- ugh so much 'work' for such little things.

sturobson:
I'm also not 'happy' that the `vf-button` default is currently (essentially) the same as `vf-button--secondary`.

The original reason for this was to when OldIE get the button they should be just grey/white because we're using css custom properties.

Something I need to look at again (very, _very_ briefly) I think.

khawkins98:
Unless I'm missing something, won't:

```css
:root {
  --default-bg-colour: some-color;
}
```

Become an issue if there's another component, say vf-tab that also has a default background colour? 
```css
:root {
  --default-bg-colour: some-color-for-tabs;
}
```

Makes the defaults unpredictable? Or should it be a global vf-core default? Or maybe that default should be scoped to `vf-button`?

sturobson:
I've just pushed a little update (well, not according to that last push … what the heck).

I've changed the naming of the variables.
```
  --vf-button__color__background--default: var(--vf-color--grey);
  --vf-button__color__background--default-dark: var(--vf-color--grey--darkest);
  --vf-button__color__border--default-visited: var(--vf-color--grey--lightest);
```

As a start for 'precedent' in how we name localised variables.

I've also pushed it inside the `vf-button` ruleset as it doesn't need to be dumped into `:root` in my opinion.

khawkins98:
That looks nice. whenever we get a chance, should document that approach or at least get a sample into the component template.

PR #610, opened 2019-10-22
the one where we update the colours a little bit

sturobson:
this PR does these things

- updates several colours so they're in 'gamut' which is raised [here](https://github.com/embl-communications/design-system/issues/20)
- removes hardcoded RGB from the design tokens as that's generated using JS in the view
- makes hex codes all uppercase


khawkins98:
Looks good. Dropped a couple minor things in #611 that would be merged into this.

sturobson:
Added a small bit of text for a landing page - I'll update as and when we get more of an idea of how the design tokens are to be presented (via recent Slack conversations)

khawkins98:
The only thing that feels weird about the welcome pattern is that it duplicates the role of the README.md ... but at this point more docs is better than less.

sturobson:
I thought this - and I wondered if we could have something in the front-matter to show/hide the additional things.

Perhaps it should have no default here  - and that a welcome page should be something in the overall design system site.

Issue #613, opened 2019-10-23
vf-fractal:build Gulp task error


dbushell:
The `vf-fractal:build` gulp task assumes Fractal config exists:
https://github.com/visual-framework/vf-core/blob/develop/tools/gulp-tasks/vf-fractal.js#L14

In cases like `vf-wp` where Fractal isn't used it errors:

```
UnhandledPromiseRejectionWarning: Error: Cannot find module '[...]/vf-wp/fractal.js'
```

Solution: do a file exists on `fractalConfig` path and resolve the task early? I suppose `vf-wp` could also avoid using `vf-build` and add its own grouped task.

khawkins98:
Yeah, this should be fairly easy to sort, will have a look.

khawkins98:
As noted in #614, I think that will half fix this issue. 

I think what we now need to do is:

> - Option B: We support client projects using `vf-build`, but we need to add some parameter so that the correct Fractal build mode is invoked (or not at all, when it's not needed)

I think to do that it would probably mean adding another config option in the client (in this case `vf-wp`) `package.json` file:

```js
  "vfConfig": {
    "vfName": "VF WordPress",
    "vfNamespace": "vfwp-",
    ...
    "vfFractalMode": "normal|dataObject|none"
  },
```

Where:

- **default** `normal`: standard full build of all static html for components, `gulp.task('vf-fractal:build', `
- `dataObject`: just render the objects to memory for further use, `gulp.task('vf-fractal:dataobject',`
- `none`: don't invoke fractal at all

sturobson:
@khawkins98 can this be closed now?

khawkins98:
Think so.

PR #616, opened 2019-10-24
Feature: add fractalBuildMode

khawkins98:
This realtes to #613, better allowing client projects to run `gulp vf-build` by allowing them to indicate how Fractal should be built.

>   // in vf-build specifices to run fractal as 'normal' full build of static assets,
>  // 'dataobject' to render only to memory or
>  // 'none' to not build at all

This should finish 613.

Next up should be to move the configuration block of text into an npm module and add documentation for it.


khawkins98:
In #617, that would also add `./tools` to lerna, so would be a good first step in decoupling things.

PR #622, opened 2019-10-28
Minor bugs, docs

khawkins98:
I might pick up a few minor fixes and enhancements as I update everything to beta.5, I'll collect the edits here.


khawkins98:
I'll be out tomorrow and then have my brain in conference mode next week, so I probably won't have much more to add to this. Good as time as any to get it in.

PR #624, opened 2019-10-29
Enhancement: improve build process this

khawkins98:
This does two things spurred on by not-quite-right features (https://github.com/visual-framework/vf-wp/pull/76) in beta.5 and a related need @dbushell spotted for `vf-wp` to copy JS assets: 

- Enhancement: Separate out logice of vf-build task … 64aeafc 
    some vf-build tasks should only be done by vf-core, this moves those tasks out of the reach of the `require('../node_modules/@visual-framework/vf-core/tools/gulp-tasks/_gulp_rollup.js')(`
- Enhancement: capture component JS assets … d01bb5c
  A new gulp task `vf-component-assets:js`  to grab JS files from components and deploy them in the `{build}/assets` url

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -35,7 +35,7 @@ build:
   script:
     - npm install
     - npm install -g gulp-cli
-    - gulp vf-build
+    - vf-core:prepare-deploy
   artifacts:
     paths:
       - build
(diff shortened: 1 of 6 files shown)


sturobson:
I'm not sure why the CI is breaking here.

khawkins98:
Ah, i accidently forgot "gulp" from "gulp vf-core:prepare-deploy" in the gitlab.ci.yml

Just committed a fix, should be sorted now.

khawkins98:
Good catch, I don't think I'd have paid any attention and just thought "Gitlab must be being stupid", turns out it is I who be stupid.

Issue #626, opened 2019-10-30
BUG - vf-button :hover/:focus interaction bug


sturobson:
![button](https://user-images.githubusercontent.com/925197/67851729-f32f9900-fb02-11e9-8eba-7c718f102257.gif)

I'm not sure how this regression has crept in. Spotted in Safari on Catalina.

khawkins98:
Also weird in Safari in 10.14.

It looks like adding:

```css
.vf-button {
  transform: translate(0,0);
}
```

sorts it.

Probably some sort of issue around using the gpu for resizing and a stutter+antialias thing, will whip up a quick PR.

sturobson:
hmm, it's commented out in the source file. It doesn't fix it in Catalina for me :-/

khawkins98:
I think this might be a Safari issue as it jitters on small movements of box shadows: https://codepen.io/khawkins98/pen/ZEEaBME

khawkins98:
I think we've done the best we can for now? (pending an upstream safari fix)

Being bold and closing, reopen if it's worth keeping open.

PR #632, opened 2019-10-31
makes the button transition slower to aid the issue in Safari

sturobson:
this tries to slowdown the Safari but we've encountered in #626 

It also removes a load of CSS that had been commented out, because it is no longer required.

We should keep #626 open, in the hope webkit fixes the underlying rendering issues

### DIFF
--- a/components/vf-button/vf-button.scss
+++ b/components/vf-button/vf-button.scss
@@ -36,7 +36,9 @@ $button-hover-fix: -5px -5px rgba(0, 0, 0, 0);
   text-align: center;
   text-decoration: none;
   transform: translate(0, 0);
-  transition: all linear 125ms;
+  transition: all linear 125ms; // Ideally we want this to be 50ms but there's an issue in Webkit/Safari on box-shadow animations
+                                // additionally we can't do a pseude :before/:after as we need to support <button> elements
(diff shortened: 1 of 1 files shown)


khawkins98:
Looks good. 

I pushed a comment to the Sass to note why we've gone with 125ms.

We're of course blocked on fully resolving this until: https://bugs.webkit.org/show_bug.cgi?id=202489

components/vf-button/vf-button.scss
```diff
@@ -16,10 +16,11 @@ $button-hover-fix: -5px -5px rgba(0, 0, 0, 0);
   --vf-button__color__background--default: var(--vf-color--grey);
   --vf-button__color__background--default-dark: var(--vf-color--grey--darkest);
   --vf-button__color__border--default-visited: var(--vf-color--grey--lightest);
-  
+
   @include set-type(text-button--1);
 
   appearance: none;
+  backface-visibility: hidden;
```
khawkins98: Do we need this? It's probably not hurting anything though.

sturobson: I thought it looked 'better' when testing it in webkit browsers with this. Could be my eyes :sob:

PR #638, opened 2019-11-06
Feauture/utility classes

sturobson:
This PR makes some headway with our utility classes.

- deprecates `vf-sass-utilities` as they're now part of the `vf-u-` name space.
- renames the `vf-sass-utilities` so they're a little more specific `vf-no-wrap` becomes `vf-u-text--nowrap`.
- adds `vf-u-float` options.
- creates a couple of 'demo' pages for floats, borders, and backgrounds
- adds a little 'intro' to the README


khawkins98:
Looking good, what else needs to happen here to get this shipped?

sturobson:
I think we could ship this now - I think there's some docs that could be added to it (as always)

khawkins98:
Had a couple thoughts i dropped into #646, mostly around stuff you probably hadn't yet gotten to as you stopped mid-work.

PR #646, opened 2019-11-11
Documentation: utility class docs

khawkins98:
- This PR merges into #638 to uncomment some documentation I think was still "to-be migrated".
- Makes the "default" into "misc" and notes it's a work in progress.
- As a last minute thought I also think "collated" makes sense as the default view.


sturobson:
I'm not convinced on collating all utility classes in one place, so I've removed that.

It'd be like a big Sass variables file - no one would scroll through it to find what they're looking for.

Also we've got things that do styles and things that do things for a11y etc etc. Plus whatever else gets added when time arrives. It could get pretty big.

khawkins98:
👍  I'm a find-in-page type, but ultimately what we'll probably need is some kind of search.

PR #660, opened 2019-11-22
COMPONENT - ebi header footer

khawkins98:
This component provides initial support for sites that need to load the EBI global header and footer. 

The ideal use case are new sites that are built with the VF 2.0 in mind but still need the EBI footer+header but don't want to have to unset all the EBI VF 1.x styles with the `ebi-vf1-integration` component. 

That is:

- The new pages @esanzgar is working on
- The EBI Group pages in WP cold use this and drop `ebi-global.css`

Looks like:

image
 
Usage:

```html
<link rel="stylesheet" href="../raw/ebi-header-footer/ebi-header-footer.css" type="text/css" media="all" />
<link rel="stylesheet" href="https://dev.ebi.emblstatic.net/web_guidelines/EBI-Icon-fonts/v1.3/fonts.css" type="text/css" media="all" />

<header id="masthead-black-bar" class="clearfix masthead-black-bar | ebi-header-footer">
</header>

Your content here.

<footer>
  <div id="global-footer" class="global-footer | ebi-header-footer vf-content">
    <nav id="global-nav-expanded" class="global-nav-expanded row">
      <!-- Footer will be automatically inserted by footer.js -->
    </nav>
    <section id="ebi-footer-meta" class="ebi-footer-meta row">
      <!-- Footer meta will be automatically inserted by footer.js -->
    </section>
  </div>
</footer>

<script defer="defer" src="//ebi.emblstatic.net/web_guidelines/EBI-Framework/v1.3/js/script.js"></script>



```

This currently isn't included in the vf-component-rollup and would need to be explicitly included as it contains quite a bit of legacy CSS. 

This is a starting point and over time we can use it to better adapt the EBI header and footer to v2.0. It already makes a start (using the VF 2.0 custom props for colours and `vf-content` for some type consistency)


sturobson:
I recall a discussion about including or improving things from (essentially) vf1.x and I thought we decided it was a 'no go'.

khawkins98:
My recollection of it is fuzzy, but I think that was about a wider "all the vf1.x things". Here if we include (and begin to port) the EBI header and footer it will effectively end a lot of pain on the group sites.

Or, perhaps, it also starts to show that EBI should have its own set of components for its needs -- but I don't think we're yet at a critical mass for that?

sturobson:
ok, a few thoughts …

- perhaps we should name-space anything here that isn't (`global-footer`) with `ebi-` as well as `global-footer` so it's 'ready' for the move
- make things like the background of the header more 'embl-y' with the design tokens (there's an override for this in group pages already).
- remove that horrible 'hover ux' for the header if it's there.
- maybe some other things when I take a fresh look on Monday.

khawkins98:
Most of that makes sense. There are some things we can't do as the content
still Congress from the legacy JS. Eventually that will change when "truth"
is the content Hub.

Might be easier to catch up on Monday's standup call.

khawkins98:
Made a few more edits over the weekend and this now uses `vf-content` so it can hand a lot more of the look and CSS:

image

There's only so far we can go as we still need to load the HTML from the existing EBI JS to avoid maintaining things in two places (EBI JS and contentHub), eventually that will switch; added in a "roadmap" to the `READEME.md`:

> ## Roadmap
> -  Version  1: Uses the existing `//ebi.emblstatic.net/web_guidelines/EBI-Framework/v1.3/js/script.js` to load the HTML for the header and footer. As such the component will use mostly legacy CSS from the existing v1.3 EBI VF.
> - Version 2: Rewrite the EBI header and footer in v2.0 HTML and pull from the contentHub, then we can drop the legacy CSS and use "pure" v2.0 CSS. 
>     - We can't yet do this as v1.3 is still the stable and supported solution for EBI sites, we will need to move a majority of sites to VF 2.0 before we can drop `v1.3/js/script.js`
> 
>

esanzgar:
The order of the menu items are different than in VF 1.3.

It should be, from left to right: EMBL-EBI, Services, Research, Training, and About us

khawkins98:
I'll have a look at this and the mobile issue. Not sure what happened, I know they were sorted at one point

PR #663, opened 2019-11-25
Bug: update usage of class names for vf-utility-classes

khawkins98:
Bug: the vf-utility-classes picked up the now-deprecated vf-sass-utilities and changed the class names, but this wasn't reflected in the docs and other component templates.


khawkins98:
We'll also need to update the contentHub templates in a couple places, but that's less urgent.

PR #664, opened 2019-11-26
the one where we introduce the start of vf-hero

sturobson:
I'm going to go out on a limb and say, even though this is a PR we need to get in, it's 'still not done'.

Adding global and component theming as well as levels of design sparkle has made the whole way we're authoring components change a little.e And we/I still need to document the theming and sparkle ideas I have to get on top of it all.

I think, however, that this PR will at least get David off the ground with having the component in `vf-wp` actually styled now. 

But I'm expecting various things that will still need to be done. 

An odd, but necessary PR.


components/vf-hero/vf-hero--example.njk
```diff
@@ -0,0 +1,48 @@
+<style>
+  :root {
+    --vf-hero-bg-image: url('{{ '../../assets/vf-hero/assets/vf-intro-group.png' | path }}');
```
khawkins98: Should this be pass from the config yml?

components/vf-hero/vf-hero--example.njk
```diff
@@ -0,0 +1,48 @@
...
+  }
+</style>
+
+
+<section class="vf-hero vf-hero--easy">
```
khawkins98: I'm assuming these would be variants but just haven't been broken out for the first iteration?

components/vf-hero/vf-hero--example.njk
sturobson: Yeah - I initially had issues w/ `.yml` and `--custom-properties` so threw it here.

We need to think about how we can consolidate all things in `:root`, it it's needed, if it's possible, if we can stop duplication. etc.

components/vf-hero/vf-hero--example.njk
sturobson: Yep.

components/vf-hero/vf-hero.scss
```diff
@@ -0,0 +1,154 @@
...
+
+  left: 12%;
+  left: calc( (0 + var(--vf-hero-content-position--inline-start) ) * 1% );
+
+  padding: calc( (0 + var(--vf-hero__padding) ) * .1em );
+}
+
+
+.vf-hero__heading {
+  background-color: rgb(calc( (0 + var(--vf-hero-heading-bg-color) ) * 255), calc( (0 + var(--vf-hero-heading-bg-color) ) * 255), calc( (0 + var(--vf-hero-heading-bg-color) ) * 255));
```
khawkins98: What's happening here, is this a contrast checking calculation?

components/vf-hero/vf-hero--example.njk
khawkins98: Yeah, have had a similar feeling -- but as of yet no good suggestions 🤔

components/vf-hero/vf-hero.scss
sturobson: ah, it's for 'on/off' of black and white. It can probably be a) improved b) documented a little better.

khawkins98:
I wonder if we should call it `embl-lens`

sturobson:
As it’s used in the training site(?). I think it’s more generic(?)

khawkins98:
Seems lines it's purely aesthetic, would anyone outside EMBL be expected to use it? Or maybe it's also functional? -- I'm assuming it's just a pattern for the masthead type areas.

khawkins98:
And by aesthetic, I suppose I mean "a visual element tied to the EMBL visual identity"

sturobson:
I mean, if we could get an actual SVG without any embedded raster images - we wouldn’t need this at all - and it could still be theming. 

Also - because there’s 3 rings _and_ the variations in layout that are possible we can’t tie the circles to the parent `vf-hero`. Hence this. 

Perhaps it’s a stop gap.

But from a) it’s usage in training and b) it’s usage in group pages (which aren’t ‘just EMBL’) I think this aesthetic is more global.

Perhaps this is one for Monday ;)

sturobson:
I’ve been thinking about this and the whole ‘adding an empty div’. Really a single SVG of actual SVG code that does it is what’s needed with the background colour being themable.

For a 0.0.1 release I think we should:

- remove the ‘lens’ completely for now
- have it so there’s a choice to have 
  - a background colour
  - an image
  - nothing (a white background colour)
- get Design to give us the correct/proper asset required.

With a proper SVG we’ll have something that’s more easily updated if the ‘lens’ changes too.

… again, we could rattle through this on Monday.

Issue #637, opened 2019-11-05
ENHANCEMENT - vf-figure--float


khawkins98:
We need something like a `vf-figure--float`  that can be used for your classic bit of image+caption to:

1. [x] have it float right or left within a block of text
2. [x] Default width should be image width (now works with ` vf-u-float__right` and  ` vf-u-float__left`)
3. [ ] Some way to control sizing 1/3, 1/4, 1/5, etc
4. [ ] stack inline (centred?) on smaller screens

Useful things:

- [existing vf-figure component](https://visual-framework.github.io/vf-core/components/detail/vf-figure.html)
- [bare bones codepen of a situation where we'd like to use it](https://codepen.io/khawkins98/pen/WNNzoEO?editors=1100)

khawkins98:
And we could probably do with some utility `vf-u-float--right` classes for ad hoc stuff.

sturobson:
ideally these would need to be hooked into WordPress' defaults somehow - as I think they're not working when set in the WYSIYG.

Also - does it need to be tied to `figure` or do we need some abstracted utility classes for this - so that they can be used be a.n.other platform as needed?

sturobson:
I'm happy to grab this one and expand on it with 'other' utility classes we could do with.

I have some other thoughts on utility classes which would be best to go into the documentation somewhere too.

khawkins98:
All good thoughts ... i kinda tried to go with the most narrow/actionable approach by focusing on vf-figure (as that's the squeaky wheel at the moment) but, yeah, agree. 

If you're up for taking this one, go for it. I won't have much time this week anyways.

khawkins98:
We've probably done enough here for the current needs.

PR #668, opened 2019-12-02
component: vf-u-width

khawkins98:
component: vf-u-width for general purpose building. 

Relates to #637, this adds responsive `vf-u-width__50` in 5 percentage point increments.

image

image

### DIFF
--- a/components/vf-utility-classes/vf-u-width.scss
+++ b/components/vf-utility-classes/vf-u-width.scss
@@ -3,35 +3,35 @@
 // .vf-width__10--sm { width: 10%; }
 
 @for $i from 1 through 20 {
-  .vf-u-width__#{$i*5} { width: 5%*$i; }
+  .vf-u-width__#{$i * 5} { width: $i * 5%; }
 }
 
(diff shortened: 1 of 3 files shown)


components/vf-utility-classes/vf-u-width.scss
```diff
@@ -0,0 +1,37 @@
+// Creates responsive sizes like:
+// .vf-width__10 { width: 10%; }
+// .vf-width__10--sm { width: 10%; }
+
+@for $i from 1 through 20 {
+  .vf-u-width__#{$i*5} { width: 5%*$i; }
```
sturobson: As you're doing `#{$i*5]` can we do `$i * 5%`? whitespace is your friend ;)

khawkins98: Makes sense. Pushing a change.

sturobson:
We could probably do with more `vf-grid__placement` classes too I guess. Something for 'when it's needed rather than doing it now 'just in case'.

PR #669, opened 2019-12-02
the one where we update vf-hero to work with full width and inlay

sturobson:
This updates quite a few things for integration

- option for including inline SVG which turns things to related colours if/when primary/secondary things are chosen
- adds full-width background w/ centred content that also works w/ absolute position text
- changed the background-position for `center top` as that might work better
- makes `vf-link` colours fit with theming
- makes `vf-hero__lens | vf-lens` a thing - rather than need `vf-lens`
- removes `vf-lens`


khawkins98:
Looks good. 

One comment: Noticed on the demo group site the z-index was conflicting with the data protection banner, should we add it to `components\vf-design-tokens\src\maps\vf-zindex.yml` so it's 1 under the banner?

khawkins98 (approved):

sturobson:
> Looks good.
> 
> One comment: Noticed on the demo group site the z-index was conflicting with the data protection banner, should we add it to `components\vf-design-tokens\src\maps\vf-zindex.yml` so it's 1 under the banner?

There's probably still a few issues here and there. Lots of room for improvement on the sprint that this facilitates overall

PR #671, opened 2019-12-03
the one where we update the navigation and start adding global things

sturobson:
This PR

- updates the navigation (main) to match new designs
- starts the addition of `vf-global-theme--` and `vf-global-layout--` classes for the 'design sparkle'

Needs to be in `vf-wp` for Thursday so I can tidy up before.

### DIFF
--- a/components/vf-inlay/vf-inlay.njk
+++ b/components/vf-inlay/vf-inlay.njk
@@ -1,66 +1,26 @@
+<div class="vf-body">
+  <div class="vf-grid">
+    <div class="vf-box">
 
-<div class="vf-global-layout--easy">
-<section class="vf-inlay">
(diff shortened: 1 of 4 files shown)


khawkins98 (approved):
Looks good. Can we do some stub docs on the vf-global-layout-- and vf-global-theme-- classes? I guess that would go upstream in something like https://github.com/visual-framework/vf-welcome/blob/master/src/site/developing/guidelines/css.njk ?

sturobson:
I've got a document I'm working on that discusses the whole global and local layout and theming things. Hopefully that should be ready soon to add.

khawkins98:
I'll pop that in an issue so we don't lose track with all the other things happening.

PR #680, opened 2019-12-12
CHORE: Refresh component versions

khawkins98:
chore: Make component version status in line with where we actually are

With this PR I intend to go through each component and:

1. make sure the .yml and package.json version status are matching (a `beta` yaml component should have an equivalent  `x.x.x-beta.x` version number)
2. fix component do not yet have an alpha/beta/rc version status
3. see if any components are `alpha` but should be `beta` (and similar)

This is partly a maintenance task but also to help us treat components accordingly (we should feel more free to change an alpha component but do more outreach/docs for a beta component)

### DIFF
--- a/components/ebi-header-footer/package.json
+++ b/components/ebi-header-footer/package.json
@@ -1,5 +1,5 @@
 {
-  "version": "0.0.2-alpha.0",
+  "version": "1.0.0-beta.0",
   "name": "@visual-framework/ebi-header-footer",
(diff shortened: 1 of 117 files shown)


khawkins98:
This should now be ready. 

Many components are now RC, which is an accurate reflection of ho we're using them in (near)production sites.

However, this PR stops short of taking any components to stable releases, which should be a follow on task/discussion.

PR #684, opened 2019-12-17
CHORE - Push some components to v1.0 stable 

khawkins98:
Following on with #680 (CHORE: Refresh component versions), we want to start treating some components as "stable" and should make them v1.0 releases. 

This will help us better handle regressions/breaking changes (requiring full version bumps) and better handle improvements and features by doing "real" semantic versioning.

There are a number that should be eldigble; including:

- vf-content
- vf-link
- vf-grid
- etc.

Anything that goes to v1.0 stable should also satisfy the guide https://visual-framework.github.io/vf-welcome/developing/components/what-is-a-component/


khawkins98:
There are still many that won't be bumped to stable:

- activity list
- discussion 
- figure
- form
- header
- hero
- masthead 
- news-container
- pagination
- search
- tree
- video

Though when we have time (hah!) we should consider what's up with them/how can we address why they can't go stable yet.

sturobson:
I think we need to nail down some sort of 'official' PR process to get us (me) up-together with things that need to be released. One of those is to make sure changes are added to the `changed.md` file.

khawkins98:
👍 Maybe that's: 

1. Updating the auto-comment bot to have "Did you update the changelog"
2. Not ignoring auto-comment bot (#guilty)

khawkins98:
Though, it's probably only more important now that we're going stable. When they we're pre-/alpha, it was all wild anyways.

PR #686, opened 2019-12-18
the one where we try and remove another excess <div> element for content-hub-html

sturobson:
Looking through the HTML on the static-html-pages for the new site I noticed that `vf-content-hub-html` element that houses the `vf-banner--phase` was wrapped inside a `vf-grid` element so that it didn't blow up the page.

We already have CSS for `vf-grid >` and `embl-grid >` so I think adding `vf-body >` for the `vf-content-hub-html` and `embl-content-hub-html` will benefit in these cases, and make the markup a little bit more sane.


khawkins98:
This seems logical. I worry if there are times when contentHub html shouldn't go in the main column -- but I can't think of anything concrete. Might be something we can deal with if/when it arises.

Issue #395, opened 2019-06-05
CHORE - Add a bunch of fundamental components (tables, cards, etc.)


sturobson:

khawkins98:
Just adding a note for later reference; List of patterns from audit is at: https://docs.google.com/spreadsheets/d/1unIfejFmxAS1ZpVE3pPN5PrGH4kdtCe-4d7iqwQggqE/edit#gid=0

khawkins98:
Spotted [this in Abstract](https://share.goabstract.com/7db0f437-aaf1-46fb-a6ad-d02e6f1d3245), where Mark had done some designs for me any of these:

![More patterns](https://user-images.githubusercontent.com/928100/77881740-e05ef300-725f-11ea-90fd-de052e512e93.png)

khawkins98:
I think most of these have been implemented or are in the backlog

Issue #314, opened 2019-03-21
COMPONENT - Explainers (`vf-explainer`)


khawkins98:
Life science websites love a bit of jargon and complicated concepts, so similar to form help or a news story background, we could use "explainers" for things like: ORCID, transcript isoform, and lymphoblastoid cell lines. -- Many of our users may know most of these, but never will all users know all. 

## Standard questions

- In a few words, what does this pattern look like?
  Largely TBD. But I imagine it will have two iterations:
  1. Near term: some sort of modal when a term is interacted with
  2. Longer term: these "explainers" could come from a repository of terms that when found on a page could be available in a "utility drawer". Also in that drawer could be a data resource explorer, EMBL exploration, related people, etc. Lots of dependencies there.

- In a few words, what does this pattern do?
  See a thing they don't know about, engage with it, learn more. Possibility to click a second link to learn even more.

- Would a rebrand change the structure or layout of this pattern?
  Nope. 

- Can other websites use this pattern? Or is it only useful in your organisation or website? 
  Yep.

- Do you have any designs or concepts?
  Nothing specific, but I do have an HTML concept. That's below.

- Have you done any user testing already? 
  no.

- How often do you expect to make use of this pattern?
  fairly often, not on every page, but we expect many EMBL teams and some news articles could use this

- Can you list out major elements in this pattern?
  term to explained, explanation, definition (`<dfn>`), further information link, "jargon score"

- Will all elements of this pattern be used every time?
  Generally yes

- Does this pattern contain other patterns?
  vf-link

## Any further thoughts?

Idea coalesced in response to #312 (updates ORCID to be an abbr); extracting meaty parts of that conversation:

> 3. it's representative of a general need to do better with technical terms and acronyms at EMBL
> 4. an opportunity to lead a bit for the wider life science community; that is: exactly what the VF is about. 
> ...
> - Better than an acronym we could offer a nice concise explainer, for example, as Europe PMC [puts it](https://europepmc.org): 
>  > ORCID is a unique identifier for researchers which distinguishes you from every other researcher, and makes it easier to find your work.
>
> ... we could instead use something like `aria-describedby` + `<dfn>`; pseudo code:
>
 ```html
<span aria-describedby="whatis-orcid">ORCID</span>

...

<div id="whatis-orcid" class="vf-explainer">
  <dfn class="vf-explainer--dfn">ORCID</dfn> is a unique identifier for researchers to distinguish themselves, and make it easier to find specific work; 
[...]

Issue #275, opened 2019-03-04
COMPONENT - Accordion/collapse/details


khawkins98:
Accordion patterns are useful in a variety of utilitarian situation, and particularly for services (I think there isn't much debate we'll need one?)

## Standard questions

- In a few words, what does this pattern look like?
  A general purpose accordion, a la https://getbootstrap.com/docs/4.1/components/collapse/

- In a few words, what does this pattern do?
  Fold outs for more information, details.

- Would a rebrand change the structure or layout of this pattern?
  No

- Can other websites use this pattern? Or is it only useful in your organisation or website? 
  Others could use

- Do you have any designs or concepts?
  Not for look, but we should use the `<detail>` HTML element with some sort of fallback for IE/Edge 
    - Docs: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details
    - Demo: https://codepen.io/jdsteinbach/pen/GYmQjo?editors=1100
    - Polyfill options: 
         - https://github.com/rstacruz/details-polyfill
         - https://github.com/javan/details-element-polyfill
         - https://www.npmjs.com/package/details-element-polyfill
         - https://codepen.io/Petrosh/pen/bpZavx

- Have you done any user testing already? 
  no

- How often do you expect to make use of this pattern?
  Perhaps most likely on application spaces (jobs, services, etc.)

- Can you list out major elements in this pattern?
  `vf-content` is likely to be inside

- Will all elements of this pattern be used every time?
  Generally yes

- Does this pattern contain other patterns?
  Don't think so.

## Any further thoughts?

Perhaps we could use tabs instead? But maybe that's too awkward.

What to call it? 

- `vf-accordion`
- `vf-collapse`
- `vf-details`

khawkins98:
Added to the backlog

Issue #235, opened 2019-02-12
COMPONENT - vf-table


khawkins98:
Capturing some notes on how we should approach a `vf-table` pattern:

From slack:

> David Bushell [4:30 PM]
is there a pattern for `<table>`?
ideally a single class that uses selectors like `.vf-table > td` etc :slightly_smiling_face:
> 
> Stuart Robson [4:32 PM]
not yet.
Something, again that could be ‘knocked’ up and informed by existing patterns.
> 
> Ken Hawkins [4:40 PM]
> It’s a good point on how we want to address the use of styling on html elements for tables … given that tables are largely a “functional” thing like `.vf-content`, @David Bushell’s suggestion `.vf-table child-elements` sounds good
> 
> Stuart Robson [4:41 PM]
yep, don’t want to have a class on every table cell… not really :sob:

khawkins98:
And there's probably a few variants that we'll want:

- vf-table
- vf-table__compact (minimal padding)
- vf-table__minimal (no strokes)
- vf-table__stack (stack in responsive mode)

Maybe we could draw code for these from bootstrap or foundation

sturobson:
They should be a double dash rather than an underscore because they’re modifiers?

khawkins98:
Good catch. I always get twisted with `__` and `--` when I'm outside of code land ... well I get twisted in code land, too.

khawkins98:
Added to the backlog

PR #693, opened 2019-12-20
component: flag in the sand

khawkins98:
These are "pre-components" that:

- Make progress on some long-standing to-do issues
- Show that we're committed to making these components
- Are a flag in the sand
- Are a "pull requests" wanted indicator
- Allow use to clean up the issue queue (see: #235 #275 #314 #395)


sturobson:
As mentioned on our call this morning. I think this would be a good PR to create:

 - a component for each, using the component generator
- keep them visible so that people can see that they are things we're looking at including.

In that last item, I wonder if there's (yet) another tag we could have for 'not even started on, but we know we need them' components?

khawkins98:
I think this one's ready for review; things:

1. Are there any other "We know we need to do it" components that we can add?
2. I think the next step after this is to move the related issues to the development backlog.

Issue #653, opened 2019-11-13
EMBL grid sidebar variant CSS


dbushell:
The `.embl-grid--has-sidebar` variant has a thin grey line separating the last two columns.

https://github.com/visual-framework/vf-core/blob/develop/components/embl-grid/embl-grid.scss#L85

The current position is using pixel units:

```css
left: -18px;
```

however it would make sense to use the `--page-grid-gap` custom prop:

```css
left: calc(var(--page-grid-gap) / -2);
```

PR #694, opened 2019-12-20
chore: embl-grid--sidebar gap

khawkins98:
Addresses #653 by using `--page-grid-gap`


components/embl-grid/embl-grid.scss
```diff
@@ -82,7 +82,7 @@
       background-color: set-ui-color(vf-ui-color--grey--light);
       content: '';
       height: 100%;
-      left: -18px;
+      left: calc(var(--page-grid-gap) / -2);
```
sturobson: if anything this should be a design token (in Sass) - let's merge it in for now

PR #702, opened 2020-01-07
FEATURE: Create github workflow

khawkins98:
Seeing what we can do with github workflows.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -1,0 +1,26 @@
+name: Node CI
+
+on: [push]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [10.x, 12.x]
+
+    steps:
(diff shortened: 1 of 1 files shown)


khawkins98:
Notes:
- It could probably replace our Travis integration and add a nicer UI.
- Makes vf-core more GitHub centric, i guess that's OK?
- Over time it would probably help us with release workflows and npm publishing.

Assuming there's nothing objectionable above, we could go ahead and merge this branch as a stub. It doesn't do anything Travis isn't already doing (Travis test _and_ publishes to `gh_pages`).

sturobson:
I'm happy with this … it would be great to see what Lerna things can get automated with this - maybe not _right away_ … But I think any confirmed merges to `develop` could then set Lerna off to publish things to `npm` automagically.

khawkins98:
Yeah, I think we're getting close to "full automation" -- and this is probably the path that gets us there quicker. It will at least lower the friction to do it.

PR #708, opened 2020-01-08
the one where we update the footer design

sturobson:
With the new site home page, there has been a few changes in the footer design. This PR reflects that:

- `text-decoration` on links removed.
- `font-size` of secondary links made the same size as category links
- `margin-bottom` removed from links
- `background-colour` is darker grey
- legalese text is now same colour as category heading

- a weird on where we had `.vf-grid__col-1 > * + * {…}` that added a top border, commented out for now.

With the darker background the text color for legalese and category headings are passing WCAG AA for colour contrast.

### DIFF
--- a/.github/auto-comment.yml
+++ b/.github/auto-comment.yml
@@ -16,6 +16,8 @@ pullRequestOpened: >
 
   Before your PR can be merged in, it will need to be reviewed.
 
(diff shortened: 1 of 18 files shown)


khawkins98 (approved):

sturobson:
this should now be good to go

PR #709, opened 2020-01-09
the one where we refactor the card component

sturobson:
As the design has matured, the card component has had more variants added to it.

This PR moves the card away from 'content type' as a classname (`vf-card--topic`, `vf-card--navigation`) and towards the 'levels of fanciness' with the Mortal Kombat classes (as adopted with `vf-hereo`). 

There's also a tiny fix where if the card had a link in the `.yml` it wasn't changing the tag from a `div` to an `a`.

todo:

- I wonder if these should be something more `HTML5`-y than a `div`
- Additional classes for the resize on `:hover` need to be added (but not crucial to this PR/MVP)


khawkins98:
Some of the layers of fanciness (medium) seem to mostly (only?) set colours? if that's right, what's the relation to the themes (`.vf-card--secondary`)? 

What I'm wondering: 

1. if there's not a substantive difference in structure for a level of faciness, would these be better as just themes (`.vf-card--inverted`), and 
2. should the themes be limited by level of fanciness? as, it might be that `.vf-card--secondary` might not work for an inverted colour scheme level of fanciness.


☝️ I hope I'm making sense, if not we can jump on zoom.

sturobson:
You're probably right with `vf-card--medium` as it's only the theming class that does anything.

1. I think we should leave 'theme' to do just that - provide the background and foreground colours. If we look at `--normal` there's an added bottom border, it doesn't change the mark up, but it changes the overall appearance, the design.
2. When David was implementing the `vf-hero` into WP he managed to find every variation - to which this has been fixed in the CSS. I don't think we can easily make it so that people won't pick a 'variant' and a 'theme' that don't work together

note: I'm still working my way through the who, what, where, when, how, and why (Jonny Ball Rivals All (TV Show)) of this, so some things might still be in the ol' noggin'. 

Do keep poking me with questions though - it can only help.

khawkins98 (approved):

PR #710, opened 2020-01-09
the one where the card get a container

sturobson:
The cards don't really have a container for when they're used. This fixes that.

It's more needed because of the resize/zoom on `:hover` effect. But the container design pattern does have a heading, so this adds that.

todo:

- see if we need a 'level of fanciness' classname for the resize/zoom on `:hover` effect


khawkins98 (approved):

khawkins98:
Hah, just saw it's draft.

components/vf-card/vf-card.config.yml
```diff
@@ -1,5 +1,5 @@
-# The title shown on the component page
-title: Card
+# The card_title shown on the component page
+card_title: Card
```
khawkins98: I think the `title` replacement on line 1, 2 here are find-replace typos.

khawkins98:
I think this branch might be out of sync with develop, going to try and pull develop into it.

components/vf-card/vf-card.config.yml
sturobson: how embarrassing

sturobson: fixed in latest commit

khawkins98:
That seems to have gone well. Changed files are now 17 and not 52.

khawkins98:
Added a little @todo on animation. 

But looks good otherwise

PR #711, opened 2020-01-09
the one where jobs get a blue button

sturobson:
just a classname change - this needs to be updated in the ContentHub too


khawkins98:
blue white zoolander trot

khawkins98 (approved):

PR #712, opened 2020-01-09
The one where we let our section headers be a link if needs be.

sturobson:


sturobson:
closing as I've made a new branch with better CSS

PR #715, opened 2020-01-10
the one where we add a parallax scroll variant for the hero component

sturobson:
This adds a simple (hah!) parallax scroll effect variant of the `vf-hero` component. 

todo: 

- tidy up JS (@khawkins98 as per slack, can you put things where the need to be to work?)
- add a `preferes-reduce-motion` bit of CSS so there's no scroll
- amend how padding is implemented on the text elements
- see what I can do about the `a` tag colours


khawkins98:
At the risk of over-investing in something that should (as I’ve understood so far) not be used outside of the EMBL.org homepage, can we skip some of these and just make it a local component in `static-html-pages` project?

sturobson:
@khawkins98 this one needs your magic with the JavaScript. 

And I need to work out how to make the CSS more … manageable.

khawkins98:
Where does this one stand -- are there still pending ideas for it?

sturobson:
I think it's safe to call this one 'dead' and delete the PR as I've another branch w/ this and a little bit more that includes the updated mortal kombat levels.

khawkins98:
👍

PR #723, opened 2020-01-16
Bug - babel js presets

khawkins98:
Gulp better rollup does not appear to be working fully with Babel anymore.
As a workaround we run babel on the roll-d up JS.

This re-enables ES2015 support for things like IE 11.

@todo: investigate this further


sturobson (approved):

khawkins98:
We've got a workaround for the new page, so will wait a little before I merge -- just incase there's a better fix that I'm too hazy to see right now.

khawkins98:
Sitting on this for the last week, I still think this is the best approach. Proceeding

PR #733, opened 2020-01-22
the one where we re-introduce a responsive logo

sturobson:
this PR brings in a responsive logo for use in places like the embl.org landing page.

It tidies up things quite a bit as we're only using the single SVG and not having simpler or more complex logos depending on viewport size.

### DIFF
--- a/components/embl-logo/embl-logo.config.yml
+++ b/components/embl-logo/embl-logo.config.yml
@@ -7,10 +7,10 @@ status: live
 #
 # Per-variant options
 variants:
+  - name: default
   - name: responsive
-    hidden: true
     context:
-      deprecated_text: "'@embl-logo--responsive' is deprecated, use '@embl-logo'"
+      classname: embl-logo--responsive
 context:
   # custom-values: passed as {{custom-values}}
   component-type: embl-element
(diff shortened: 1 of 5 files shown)


khawkins98:
I wonder if we should call the variant `embl-logo--hero` or `embl-logo--extreme`, as otherwise it seems like it's just adding responsiveness, and you might add the class thinking "a sprinkle of responsive is good" and not expect a big logo.

components/embl-logo/embl-logo.config.yml
```diff
@@ -7,10 +7,10 @@ status: live
 #
 # Per-variant options
 variants:
+  - name: default
   - name: responsive
```
khawkins98: Should we also name the variant `extreme`?

khawkins98: That comment was meant for line 11.

sturobson: Ah, yes. Will update.

sturobson:
actually - should the changes to `vf-logo` be a separate issue/PR?

khawkins98:
I guess it depends? As long as it's a similar edit, feels right by me ...

sturobson:
Actually, if we're going down this route - do we need the differentia betwixt `vf-logo` and `embl-logo`

sturobson:
> Actually, if we're going down this route - do we need the differentia betwixt `vf-logo` and `embl-logo`

yes, we do. For now.

sturobson:
@khawkins98 ready for re-review.

khawkins98 (approved):

PR #735, opened 2020-01-22
the one where we update the section header again

sturobson:
This section header -

- adds ability to have a link
- adds ability to have a subheading

### DIFF
--- a/components/vf-section-header/vf-section-header.scss
+++ b/components/vf-section-header/vf-section-header.scss
@@ -12,9 +12,10 @@
 .vf-section-header__heading {
   @include set-type(text-heading--3, $custom-margin-bottom: 24px);
 }
-a.vf-section-header__heading {
+.vf-section-header__heading--is-link {
   color: set-color(vf-color--blue);
   cursor: pointer;
+  display: block;
 
   .vf-section-header__icon {
     fill: currentColor;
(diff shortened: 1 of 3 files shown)


components/vf-section-header/vf-section-header.scss
```diff
@@ -10,6 +10,28 @@
  */
 
 .vf-section-header__heading {
-  @include set-type(text-heading--3);
-  margin: 0;
+  @include set-type(text-heading--3, $custom-margin-bottom: 24px);
+}
+a.vf-section-header__heading {
```
khawkins98: why not `.vf-section-header__heading--is-link`? or something similar?

sturobson: I was thinking this when writing it. 

I guess, as we should be moving things on embl.org to using the templates rather than having it as copy/pasted HTML like it currently is - we can do this.

I wonder if `.is-link` would be better and more 'atomic' but then I have a feeling we will be hit by 'edge cases' anyhow.

Will update to `vf-section-header__heading--is-link`and push -- although that is quite the mouthful :sob:

sturobson:
I've updated to change the `a` to a class. I've also added other variants too

khawkins98 (approved):

PR #738, opened 2020-01-24
Bug/component hotfixes

khawkins98:
Fixes a few components in the last release. 

- Fix vf-card context data … 27e6a43
  - Makes vf-card have individual variants, vf-card--container invokes those variants, passing context to vf-card
  - This might also resolve issues around needing "@include"

- vf-section-header wasn't including href attribute … e25ef01
  - Also adds `  @include inline-link;`, otherwise link was getting underlined.

### DIFF
--- a/components/vf-card-container/vf-card-container.njk
+++ b/components/vf-card-container/vf-card-container.njk
@@ -1,12 +1,11 @@
 <section class="vf-card-container{% if modifier %} | {{modifier}}{% endif %}">
   <div class="vf-card-container__inner">
-
     {% include '@vf-section-header' %}
 
     {% asyncEach card in vf_cards %}
-      {% include '@vf-card' %}
+       {# prefix cardtype #}
+       {% set card_type = '@vf-card--'+card.card_type %}
+       {% render card_type, { 'context': card } %}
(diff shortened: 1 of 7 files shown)


components/vf-card/vf-card.scss
```diff
@@ -70,9 +70,8 @@
 }
 
 // vf-card levels of design
-
 .vf-card--easy {
```
sturobson, line 73: isn't this card-normal 

card easy being not real styling (as yet) . Well, that's how I had it

components/vf-card-container/vf-card-container.njk
```diff
@@ -1,12 +1,11 @@
 <section class="vf-card-container{% if modifier %} | {{modifier}}{% endif %}">
   <div class="vf-card-container__inner">
-
     {% include '@vf-section-header' %}
 
     {% asyncEach card in vf_cards %}
-      {% include '@vf-card' %}
+       {# prefix cardtype #}
```
sturobson, line 6: ugh, so this was all working fine for me in `vf-core` was the issue in 11ty?

sturobson (approved):

PR #739, opened 2020-01-24
Enhancement: use example com

khawkins98:
One for dicussion, but suggest we use the generic example.com domain for links over javascript:void as it makes things a bit more "real".

### DIFF
--- a/components/_previews/_preview--fullhtml.njk
+++ b/components/_previews/_preview--fullhtml.njk
@@ -1,1 +1,3 @@
 {{ yield | safe }}
+  <!-- This script is needed only for the VF component library -->
+  <script src="{{ '/themes/vf-frctl-theme/js/vf-frctl-theme.js' | path }}"></script>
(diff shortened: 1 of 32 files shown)


sturobson:
the reasoning behind not using a real url is when you click it, you get the visited state. Which might look 'odd' and not help further design and development of vf-core.

khawkins98:
That's what bugs me a bout `js:void` it makes it harder to test visited states.

khawkins98:
At the risk of over engineering, it would be nice if there was a random has on each page load.

khawkins98:
It would "probably" be easy enough to make a vf-core only bit of js that found all the URLs to example.com and appended a `example.com/?randmom=32423432`

khawkins98:
Couldn't resist adding random hashes.

khawkins98:
Had one more thought over the weekend and have revised to work like this:

1. `javascript:void(0);` is the default template-passed URL
2. When JS loads, they're swapped with a hashed URL to example.com

This way we ensure we get non-visited links on page load and then can click links to view visited status.

PR #750, opened 2020-02-05
Make the right sidebar of vf-inlay responsive

kasprzyk-sz:
So the sidebar content will be two columns by default of mobile.

As noticed in the draft News site but also visible in the vf-inlay pattern as it is now.

Written by Ken.

See also: https://github.com/visual-framework/vf-wp/pull/137

### DIFF
--- a/components/vf-inlay/vf-inlay.scss
+++ b/components/vf-inlay/vf-inlay.scss
@@ -73,6 +73,8 @@
     ;
   }
   /* stylelint-enable */
+
+  padding-bottom: calc(var(--page-grid-gap)/2);
 }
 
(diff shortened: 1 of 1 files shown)


components/vf-inlay/vf-inlay.scss
```diff
@@ -82,18 +84,27 @@
 
 .vf-inlay__content--main {
   grid-column: 2 / -2;
-  padding-right: var(--page-grid-gap);
 
-  @media (min-width: 1100px) {
```
sturobson, line 87: breakpoints are only to come from the list of breakpoints available.

sturobson (approved):

Issue #727, opened 2020-01-17
proposal -  make `vf-links--summary` linkable


sturobson:

khawkins98:
Footnote: there was some related thinking about summary images at #759 (The one where we refactor the summary component)

khawkins98:
Resurrecting this one which is a follow up to the EMBL.org homepage launch: 

![image](https://user-images.githubusercontent.com/928100/81543073-e4f5eb80-9375-11ea-9371-0495dddd31d0.png)

https://stable.visual-framework.dev/components/vf-summary/#vf-summary--news-has-image

sturobson:
similar to #910 … which I'm working on/

sturobson:
looking at this, along with a similar change to `vf-card`.

notes and further questions
- there will be a need for a breaking change
- would there ever be a link in the text that might link somewhere else that needs to work
- are we only wanting the image and text to be an url, rather than 'the whole thing'
- are dates linkable to see search results of other things with that date?

khawkins98:
> - would there ever be a link in the text that might link somewhere else that needs to work

Quite possibly.

> - are we only wanting the image and text to be an url, rather than 'the whole thing'

Yeah, I think that's best/most expected/leaves open other links inside the teaser text.

> - are dates linkable to see search results of other things with that date?

As long as we're refactoring, would be good to consider this ... this could also be an area for an author byline or category.

sturobson:
These changes along with the changes to #910 would create breaking changes for both.

Do you want them in before or after the vf2.0.1 release?

khawkins98:
I suppose this is now redundant with the wider vf-summary refactor?

PR #759, opened 2020-02-11
The one where we refactor the summary component

sturobson:
This PR was initially to fix the link colour issue. It does that, but a little more.

- updates the link colour issue so it's not white text on a white background
- makes use of the nunjucks template more and reduces the amount of files
- makes all design variants use .yml for content
- reworks the yaml key/value pair so they're consistently named


sturobson:
I've added the 'breaking change' tag. 

`vf-summary--has-image`is on its way to being deprecated, currently for `vf-summary--media` although, `vf--media-object` might be more fitting for the design pattern so it's more generic.

For now, I've removed `vf-summary--has-image` as a `.njk` file and replaced it but kept the CSS.

khawkins98:
Smattering of thing/notes -- apologies as it's more than i expected it to be!

- For the new `vf-summary--media` we'll need something like `vf-summary--media__image-thumbnail` so the media list won't always be 180px wide (Assorted uses but current one is for data resource list on group sites)
- Since we're removing `vf-summary--has-image` from the component library list, why not go ahead and formally deprecate it? Otherwise I'm sure we'll forget later.
- Profile image is no longer a circle, that intended?
- `--media-no-image` now sounds confusing, maybe we don't need it?
- similarly, do we need "news" and "news has image", maybe "news" with an image is enough and we can trust that the pattern works without an image? 
- profile large/regular/small should note that they are for profiles, they are only for profiles, yeah? And the large profile is smaller than "profile default"? Maybe the profile default should be removed.
- As it's breaking, I guess we should update this to `v2.0.0-alpha.1`?

If that's all a bit confusing, can make a go of them in a sub-PR.

sturobson:
> For the new vf-summary--media we'll need something like vf-summary--media__image-thumbnail so the media list won't always be 180px wide (Assorted uses but current one is for data resource list on group sites)

`.vf-summary--media` replaces `vf-summary-has-image` https://visual-framework.github.io/vf-core/components/detail/vf-summary--has-image.html which is different to `vf-summary--news` which I think you're referring to.

> Since we're removing vf-summary--has-image from the component library list, why not go ahead and formally deprecate it? Otherwise I'm sure we'll forget later.

👍 

> similarly, do we need "news" and "news has image", maybe "news" with an image is enough and we can trust that the pattern works without an image?

This uses CSS grid to work out if there's an image or not. Unlike `--has-image` there is the design decision that news without an image doesn't need to have the 'space' where an image would be

> As it's breaking, I guess we should update this to v2.0.0-alpha.1

Hmmm, I guess so

sturobson:
I'm working on trying to finish this today … and I've added this comment just to see if gitify on MacOS works

components/vf-summary/vf-summary--profile.scss
```diff
@@ -30,7 +30,7 @@
   @include inline-link;
 }
 .vf-summary__link--secondary {
-  @include inline-link($vf-link---secondary-color, $vf-link---secondary-color--hover, $vf-link---secondary-color--visited, $vf-include-normalisations: true);
+  @include inline-link(set-color(vf-color--grey--darkest), set-ui-color(vf-ui-color--black), set-color(vf-color--purple), $vf-include-normalisations: true);
```
khawkins98, line 33: How much is left to get this PR ready? It'd be really nice to get in as this CSS will fix the issue around white on white links.

khawkins98: Or, is what is already done usable, and we can move the remaining work to a new issue?

sturobson:
I think we can close this one now, and come back to it if needed later,

khawkins98:
Adding a footnote from slack conversations so we can remember why we closed this:

1. The bug-ish behaviour of white links is resolved in #825
1. There is still a related issue to pick back up on some of this work at  #727 (proposal - vf-links--summary image is a link)
1. There were (probably) some wrong turns in njk template complexity, from slack:

> Ok … looking at vf-summary with ‘fresh eyes’ I’m thinking I went too far here - and we should probably look at seeing if the --variants actually are their own component or not and move all of this (code below) in separate .njk files. I think going the way I did will be a maintenance nightmare in the future without excess documentation and inline comments (which won’t be needed as much if every --variant was a separate file again.

PR #767, opened 2020-02-18
Suggestion: Hide is-a-link and use in some examples

khawkins98:
Since it's a utility-class-like variant, do we need to show it? And might be good to demonstrate it's use.


sturobson:
I thought being explicit would make more sense as not all boxes would be a link (well, for now). 

I'll approve anyways. Perhaps the docs could be tightened up for this somehow.

sturobson (approved):

khawkins98:
Probably again gets to the idea of how we present the variants. Feels like you either got a simple linked variant or "designed" one.

PR #769, opened 2020-02-18
the one where we update the mortal kombat levels of design

sturobson:
After discussions:

- remove `--medium` as it feels like it could be focused with `--normal`.
- add `--very-easy` to match the actual Mortal Kombat levels
- pull back any levels of design fanciness as required to match the new levels of design.

### DIFF
--- a/components/vf-hero/CHANGELOG.md
+++ b/components/vf-hero/CHANGELOG.md
@@ -1,15 +1,5 @@
-# 0.1.0
+## 1.2.0
 
-### Added
-- added this
-- added that
-
-### Changes
-
-- changed this
-- changed that
-
-### Removed
-
-- deleted this
-- deleted that
+* Removes `--medium` variants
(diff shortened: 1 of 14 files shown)


components/vf-hero/CHANGELOG.md
```diff
@@ -1,15 +1,5 @@
-# 0.1.0
+## 1.2.0
```
khawkins98, line 1: Why the jump to 1.2.0? To make it instep with vf-card?

sturobson: yah, pretty much - as it's a sweeping change - I thought bumping it all up made sense.

khawkins98:
Looks good. 

Would be nice to capture some of this in the documentation this week while it's still fresh in mind -- and also as some notes in the component generator's `_component.scss` and `_component.config.yml`

khawkins98 (approved):

PR #777, opened 2020-02-25
The one where the vf-section-header goes all blue

sturobson:
Because of Safari not respecting `currentColor` for an SVG that's inside a `:visited` `<a>` tag and because it looks 'odd' for components. We have made a stylistic change to the `:visited` state. Plus some other little nice-to-haves.

- :visited styles so they're blue
- removes cursor: pointer and display: block from --is-link
- moves transform out of todo
- fixes transition-property to animation works
- adds last-of-type to vf-section-header__text to remove any margin

### DIFF
--- a/components/vf-sass-config/CHANGELOG.md
+++ b/components/vf-sass-config/CHANGELOG.md
@@ -3,6 +3,15 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+# 1.1.0
+
+* adds ability for section header to have sub-heading and text.
+* :visited styles so they're blue.
+* removes cursor: pointer and display: block from --is-link.
+* moves transform out of todo.
+* fixes transition-property to animation works.
+* adds last-of-type to vf-section-header__text to remove any margin.
+
 # 1.0.1 (2020-01-24)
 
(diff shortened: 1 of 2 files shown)


khawkins98 (approved):
Looks good. 

Small issue in `:hover` + `:visited` you don't get the darker blue. I'm sure it would be months until anyone else notices.

sturobson:
> Looks good.
> 
> Small issue in `:hover` + `:visited` you don't get the darker blue. I'm sure it would be months until anyone else notices.

I've fixed this in the last commit, and it works in Safari (which I was surprised at).

components/vf-sass-config/CHANGELOG.md
```diff
@@ -3,6 +3,15 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+# 1.1.0
```
khawkins98, line 6: I think the wrong changelog got updated.

sturobson: Oh damn! I'll make a PR to fix the bug

khawkins98: Ah, have just whipped one up, hopefully you saw already: https://github.com/visual-framework/vf-core/pull/778/files

PR #778, opened 2020-02-25
chore: update-vf-section-header-changelog

khawkins98:
Would correct so the right changelog gets udpated; see https://github.com/visual-framework/vf-core/pull/777/files#r383837870


sturobson (approved):
👍

PR #781, opened 2020-02-27
Clean up ebi-header-footer

khawkins98:
Changes to other vf components need to be considered in the ebi-header-footer compatibility code. Mostly changes around base font size and the removal of underlines on fonts.

Currently:

image

Revised:

image

Ping @esanzgar


sturobson (approved):

esanzgar:
Thanks Ken!

Issue #751, opened 2020-02-05
Enhancement: insert svg icons via JS


khawkins98:
For `vf-section-header--link` why not use an actual SVG file via CSS? 

Would make updates easier if/when the graphic changes.

https://github.com/visual-framework/vf-core/blob/cc8ac4a153af2f492515e7dd0ccdc6f523455853/components/vf-section-header/vf-section-header.njk#L15

sturobson:
So, there's the HTTP request for one icon (or even the small set of icons we currently use if made into a sprite) that kinda swayed the decision here. 

There's also the few requirements, like it needs to be easily controlled via CSS.

I'd expect, more often than not, that the nunjucks/WP template should be used (as a best practice), rather than copying and pasting. Which means inlining is fine.

khawkins98:
Keep in mind that 11ty/WP will largely be used only for the corporate portion EMBL.org/EBI.ac.uk sites. Service teams will be using _other_ solutions, so it will make changes to the icons hard to manage/enforce without inspecting source code.

khawkins98:
What sort of CSS control do you mean? Positioning effects/transitions?

khawkins98:
Thinking ~aloud~ typing: depending on the needs of that CSS control, we could make the inline an option when using the NJK template. So if you use the template you get the inline, if not you get the HTML fetch.

khawkins98:
Picking up on https://github.com/visual-framework/vf-core/pull/785#discussion_r387543577

> The 'on this page' list needs an SVG icon -- We need to work out if inline is the best approach, or if we should be creating a sprite. We'd need to be able to style the SVG w/ CSS (at least)

Another option would be to use JS, where we need SVGs we could:

1. Add `data-vf-add-this-icon-here`
2. Component `vf-icons-whatevername` picks up on that data attribute
3. Injects the SVG inline

This would:

1. keep the HTML clean
2. not result in extra HTTP requests (the SVG inline code could be in the minified JS)
3. still allow CSS cascade styling
4. no faffing about with sprites

I think it would be a fair use of JS, as if/when it fails you don't get icons, which shouldn't be critical to your content. (right?)

khawkins98:
We don't have the capacity to action this, it is being moved to the backlog #446

PR #785, opened 2020-03-04
the one where we update vf-links-list a little bit

sturobson:
This brings in a couple of variations on the links lists available. 

For the new 'in this section' list we can add `vf-links__list--very-easy`

For the 'on this page' list we can use `vf-links__list--easy` 

Note: 
- There's also some additional classnames needed for the 'in this section' component to be 'tight' and 'small' -- I may refactor this at some point
- The 'on this page' list needs an SVG icon -- We need to work out if inline is the best approach, or if we should be creating a sprite. We'd need to be able to style the SVG w/ CSS (at least)


components/vf-link-list/vf-link-list--easy.njk
```diff
@@ -0,0 +1,25 @@
+<div class="vf-links {%- if component_modifier %} {{ component_modifier }}{% endif %}">
```
khawkins98: Does this need to be a separate template? (Recalling that for wordpress one template file is better)

components/vf-link-list/vf-link-list--easy.njk
```diff
@@ -0,0 +1,25 @@
...
+        {% if item.image %}
+          {% if item.image == "blank" %}
+            <span><!-- no image --></span>
+            {% else %}
+            {{ item.image }}
+          {% endif %}
+        {% endif %}
+        {{ item.text }}
+
+        <svg class="vf-icon vf-icon__arrown--down | vf-list__icon" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24"><title>arrow-button-down</title><path d="M.249,7.207,11.233,19.678h0a1.066,1.066,0,0,0,1.539,0L23.751,7.207a.987.987,0,0,0-.107-1.414l-1.85-1.557a1.028,1.028,0,0,0-1.438.111L12.191,13.8a.25.25,0,0,1-.379,0L3.644,4.346A1.021,1.021,0,0,0,2.948,4a1,1,0,0,0-.741.238L.356,5.793A.988.988,0,0,0,0,6.478.978.978,0,0,0,.249,7.207Z"/></svg>
```
khawkins98: The entropy of inline SVGs for the HTML template examples still makes me worry, but I get that's where we're at for the now.

components/vf-link-list/vf-link-list--easy.njk
```diff
@@ -0,0 +1,25 @@
...
+      </a>
+      {% if item.badge %}{% render '@vf-badge' %}{% endif %}
```
khawkins98: Tried using item.bade and item.meta, but it came out weird? Is this meant to be used? 

image

khawkins98 (commented):
In the README.md (or somewhere) would be good to have some info on when Very Easy/Easy are appropriate to be used, especially versus "Tight".

components/vf-link-list/vf-link-list--easy.njk
khawkins98: > The 'on this page' list needs an SVG icon -- We need to work out if inline is the best approach, or if we should be creating a sprite. We'd need to be able to style the SVG w/ CSS (at least)

Missed that

khawkins98: But, yes, very much so :)

components/vf-link-list/vf-link-list--easy.njk
sturobson: Yeah. It does really - I'll amend, update, and push

components/vf-link-list/vf-link-list--easy.njk
sturobson: I have in the past used `use` with the SVG pulled inline at the top of the page with a `display: none` attached. Which worked. Following several conversation on SVGs over the years it seems that 'best practice' is to inline (even with the issue of maintainability)

At least we have a confirmed pool of icons to go from - so the chance of their being a drastic change is slightly reduced.

components/vf-link-list/vf-link-list--easy.njk
sturobson: Hmmm, not in the current schema of things - but I can see that it might be a good thing. Especially if it's directing to some content on a page that is 'in beta' -- but then wouldn't the whole page be 'in beta' because of that, negating the need for an individual label?

I'll work on seeing what can be done - the problem here is that the meta and badge are 'outside' the `<a>` where the SVG is inside so that it matches state styles.

components/vf-link-list/vf-link-list--easy.njk
khawkins98: Incase it gets lost in the notifications: https://github.com/visual-framework/vf-core/issues/751#issuecomment-594417259

components/vf-link-list/vf-link-list--easy.njk
khawkins98: Use cases seem appealing ... just seemed off and noted it wasn't in the example, thought maybe it was in-process code that got left behind. Little additional text things could really be nice. Probably a docs thing in the end.

sturobson: Screenshot 2020-03-04 at 09 48 54

We could do this with a little CSS grid, it would place the SVG outside of the `<a>` tag however - which incurs additional CSS for the states 

The idea is to keep the SVG next to the title, so in this example short text links will have a massive gap before the arrow.

sturobson: Screenshot 2020-03-04 at 09 52 29

Which we could turn to something like this


But does it make sense with the arrow to have the potential badge and meta underneath still?

khawkins98: I wonder if the badge is helpful? I mean it should go next to the content that is alpha/beta, right?

I do like the text snippets, but do we _have_ to have it right now? Suppose it could wind up becoming mini paragraphs quite easily, which might be good, might be terrible.

sturobson: > It could wind up becoming mini paragraphs quite easily.

Yeah, and possibly duplicate the content it's trying to direct the user too. 

Perhaps, for now it's a 'do not use it like this' thing

components/vf-link-list/vf-link-list--easy.njk
sturobson: With the comment in #751 I'll resolve this conversation

sturobson:
> In the README.md (or somewhere) would be good to have some info on when Very Easy/Easy are appropriate to be used, especially versus "Tight".

Thinking about it - I'm not sure if we have deployed anything that uses the 'tight' variant.

G'ah! I'm wondering if it's appropriate to have it at all? 

So many 'variables' and variants and options.

Let's leave it 'as is for now'

khawkins98 (approved):

PR #786, opened 2020-03-04
Component: embl-notifications

khawkins98:
Now that EMBL.org and the contentHub are live, it would be good to begin using them for status updates, notifications, emergency information. 

This component will fetch notifications from the EMBL contentHub and display them, if a matching URL is found.

1. [x] Fetch json from contentHub
1. [x] get current page url
1. [x]  compare to `notification_ulrs`
1. [x] If matching URL invoke vf-banner
   - using options from `notification_position` `notification_presentation` `body` `title` `notification_link`


khawkins98:
Now have a feed: https://dev.content.embl.org/api/v1/notifications?_format=json

khawkins98:
This one should be ready now.

sturobson (approved):
You probably know requirements a lot more than me here, so these question could be null and void:

- is being EMBL specific good for this? 
- is having additional HTML in the `.njk` the way to go?

re: the last question - could we have the default `.njk` for this component only have that and then make use of something like `--example.njk` and `.yml` to show developer/user what it actually does?

Approving anyways, as there's probably a slight sense of urgency to have this 'ready, if needed'.

khawkins98:
> is being EMBL specific good for this?

I think so, potentially another org could pick up the pattern and provide their own JSON feed, but I think that's "out of scope" for now. 

> make use of something like --example.njk

👍

khawkins98:
Merging into dev. Maybe more ideas come up later, but enough to get on development. methinks. 

methinks

PR #793, opened 2020-03-09
bug: vf-figure path handling

khawkins98:
This removes the ` | path` from the njk template which made it non-usable outside of fractal (it now follow the docs made sense this pattern was last updated).

Also eats extra white space.

### DIFF
--- a/components/vf-figure/CHANGELOG.md
+++ b/components/vf-figure/CHANGELOG.md
@@ -1,16 +1,5 @@
 # Change Log
 
-All notable changes to this project will be documented in this file.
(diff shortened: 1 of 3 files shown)


sturobson:
Does this still work when we run the build? I think that's why I started to add `| path`

khawkins98:
Still works, I updated the image path. 

Best way is this, which we figured out after we made vf-figure way back when:

https://github.com/visual-framework/vf-core/blob/5d1ab0920177563e92dd1d0b81f41653b232051d/tools/component-generator/templates/_component.config.yml#L19-L21

sturobson (approved):

PR #796, opened 2020-03-09
the one that adds a full bleed utility class

sturobson:


sturobson:
There's some things here I still need to look at. As 'via the twitters' `100vw` breaks a little if the browser has visible scrollbars. 

There's a couple of options:

1. 10 lines of JS that add a class to the body and creates a custom property to take off the width
2. add `overflow-x: hidden;` to the body element.

The latter goes against the 'only use classes' and _could_ break pages that use VF1.x components and CSS

The former requires JavaScript.

khawkins98:
Pretty sure you're still tinkering away at this, but shout whenever you're waiting on me for something.

sturobson:
@khawkins98 ready for review now :)

components/_previews/_preview--fullhtml.njk
```diff
@@ -1 +1,28 @@
+<!DOCTYPE html>
```
khawkins98: I think this was an accidental commit?

sturobson: ah, yeah - we don't need that if the issue I had with the GDPR disclaimer and other notification showing up is now fixed. Will remove and push

sturobson: hmm, which bit was the accident? 

I don't think we need preview--nojs which I added now

khawkins98: `_preview--fullhtml` is used by other components, so we shouldn't update that one (well not without considering impacts). -- could you use the default preview, still seems to work ok?

sturobson: I _think_ the only change is whitespace? Will check, fix, and update

sturobson: updated the PR with another push of some codes.

khawkins98:
👍

khawkins98 (approved):

PR #802, opened 2020-03-16
fixes issue with vf-grid in production

sturobson:
there's a little bug in places where `vf-grid` causes the `vf-body` component to not be the correct sizes expected.

removing this line from `vf-grid` seems to fix this and not break in places it is used.

```
@media (max-width: 1299px) {
  .vf-grid {
    grid-column: 1 / -1;
  }
}
```

however, this breaks if the `vf-grid` is inside an `embl-grid` like it is on the news homepage - https://wwwdev.embl.org/news/

so the CSS is commented out, and then repeated underneath but scored to `embl-grid`.

This shows that we need to take a look at the original ideas of the grid systems and work towards making a cleaner initial 'page layout' which doesn't rely on us adding a classname or adding CSS to an element.

### DIFF
--- a/components/vf-grid/README.md
+++ b/components/vf-grid/README.md
@@ -12,6 +12,8 @@ You can define the number of columns with a modifier class.
 
 ***Do Not Use*** with any component that uses the `<table>` HTML element as this breaks built-in browser accessibility for screen readers etc.
 
+Currently (16/03/20) The `vf-grid` is expecting to be a parent of `vf-body`. It does, however, work inside `embl-grid` now as we have added CSS to make it respect the boundaries so that it doesn't break. 
+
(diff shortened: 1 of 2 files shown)


khawkins98:
So it looks like for `.embl-grid>.vf-grid` it's intended as a sort of grid reset to make things go full width? 

If so, add that to the docs (Sass comment) for future reference?

sturobson:
> So it looks like for .embl-grid>.vf-grid it's intended as a sort of grid reset to make things go full width?

For now, as I think an overhaul of the grids so that they're a little more robust are needed.

Will add this to change to the docs and push an update.

khawkins98 (approved):

PR #805, opened 2020-03-19
makes changes so njk works in 11ty or WP

### LINKED ISSUE #804: Bug/enhancement: Be more portable with NJK templates
When using the Nunjucks templates, different environments handle `context` differently.

So far we've seen difference between Eleventy and WP's Gutenberg (React):

https://github.com/visual-framework/vf-core/blob/2ab60bb0e5abbf7c2b8e1d1c52ac9128864bce60/components/vf-box/vf-box.njk#L3-L6

This causes issues in WP as there's no easy way to set `context` or pass `field: { box: field_name }` 

An improvement could be to "flatten" the context object in the above code:

```njk
 {# if we're being passed "context" from a parent template, use it as "box" #} 
 {%- if context %} 
   {% set box_title = context.box_title %} 
   {% set box_body = context.box_body %} 
   ...
 {% endif %} 
```

In theory this would then suit Eleventy and Gutenberg.

sturobson:
This should make things a little more portable when we have texted components or sub components that rely on the parent to determine what the data could be.

We check to see if the context knows it is a sub-component and if so sets the keys to be prefixed to the correct component type.

I've tested this in the `vf-core` where one component has nested data and the other does not and it seems to work.

👆 it's not easy explaining this :sob:

This should close #804

### DIFF
--- a/components/vf-box/vf-box.njk
+++ b/components/vf-box/vf-box.njk
@@ -1,11 +1,22 @@
-{% if box.deprecated_text %}<!-- {{ box.deprecated_text }} -->{% endif -%}
-
 {# if we're being passed "context" from a parent template, use it as "box" #}
 {%- if context %}
   {% set box = context %}
 {% endif %}
 
-{% if box.box_href %}
+{%- if box %}
+  {% set deprecated_text = box.deprecated_text %}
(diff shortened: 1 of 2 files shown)


khawkins98 (approved):
👍

PR #807, opened 2020-03-20
adds simple CSS for inline links in vf-box__text

sturobson:
As we could be swapping the WP [fact box] for `vf-box` we need to incorporate the need for links in the `vf-box__text`. 

This does this by making the `<a>` inherit the `currentColor` from the `<p class="vf-box__text">`.

Screenshot 2020-03-20 at 10 37 18

### DIFF
--- a/components/vf-box/README.md
+++ b/components/vf-box/README.md
@@ -6,6 +6,12 @@
 
 The `vf-box` container is an element to be used when ...
 
+## Usage
+
+A `vf-box` can be used in all grid systems and also inside of `vf-content`.
+
+For now `vf-box` is only designed to accept a heading and text (both have classes). The text node – `vf-box__text` can also accept a link which will inherit the text colour. 
+
 ## Options
 
 ### Is Link
--- a/components/vf-box/vf-box.scss
+++ b/components/vf-box/vf-box.scss
@@ -51,6 +51,13 @@
 
 .vf-box__text {
   @include set-type(text-body--3);
+
+  // some times there's a link inside the text in old WP posts.
+  a {
+    border-bottom: 1px solid currentColor;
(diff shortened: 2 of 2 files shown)


khawkins98:
As this is pretty WP specific, should it instead be some local CSS for vf-wp?

khawkins98:
image

WP will probably also need some spacing tweaks on that heading too

sturobson:
I think we can incorporate it in `vf-core` as this does then allow factoids or other 'text in `vf-box`' to have a link - as I'm sure that will definitely be a requirement at some point.

sturobson:
> WP will probably also need some spacing tweaks on that heading too 

I'm going to go out on a limb and say that _that_ is `vf-wp` specific.

khawkins98:
> I think we can incorporate it in vf-core as this does then allow factoids or other 'text in vf-box' to have a link - as I'm sure that will definitely be a requirement at some point.

In this case it might be more future proof if we do:

```css

.vf-box.vf-content a, 
and/or .vf-content .vf-box a {}
```

sturobson:
Yes! It's probably a good idea to add `.vf-content .vf-box a {}`.

I wouldn't expect people to include `vf-content` inside a `vf-box` (unless they're hand coding things (which, they probably shouldn't be doing _really_) … so perhaps I should mention that in the README.md

khawkins98:
👍

sturobson:
I've updated the CSS so that a `vf-box` will now work as a child component insdie `vf-content`. I've also added some simple docs to this effect. I've also gone back and reinstated `text-decoration: underline` so it matches the rest of the system.

khawkins98 (approved):

PR #814, opened 2020-03-23
Fix vf-content blockquote spacing

khawkins98:
- Fix vf-content blockquote spacing on last p tag 
- Also adds support for styling of figcaption and cite

These were most visible on the WP News site, but apply everywhere vf-content is used with these elements.

### DIFF
--- a/components/vf-content/vf-content.scss
+++ b/components/vf-content/vf-content.scss
@@ -122,6 +122,18 @@
 
   blockquote:not([class*='vf-']) {
     @include blockquote;
+
+    p:last-of-type:not([class*='vf-']) {
+      margin-bottom: 0;
+    }
+  }
+
+  figcaption:not([class*='vf-']),
+  cite:not([class*='vf-']) {
+    // matches @mixin figure, but that is not directly usable here as it requires css classes
+    @include set-type(text-body--5);
+    color: map-get($vf-colors-map, vf-color--grey);
+    font-style: italic;
   }
 }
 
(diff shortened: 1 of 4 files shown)


components/vf-content/vf-content.scss
```diff
@@ -122,6 +122,18 @@
...
+      margin-bottom: 0;
+    }
+  }
+
+  figcaption:not([class*='vf-']),
+  cite:not([class*='vf-']) {
+    // matches @mixin figure, but that is not directly usable here as it requires css classes
+    @include set-type(text-body--5);
+    color: map-get($vf-colors-map, vf-color--grey);
+    font-style: italic;
```
sturobson, line 136: has me wondering if this should be part of the typographical 'tokens'

components/vf-content/vf-content.scss
```diff
@@ -122,6 +122,18 @@
...
+    p:last-of-type:not([class*='vf-']) {
+      margin-bottom: 0;
+    }
+  }
+
+  figcaption:not([class*='vf-']),
+  cite:not([class*='vf-']) {
+    // matches @mixin figure, but that is not directly usable here as it requires css classes
+    @include set-type(text-body--5);
+    color: map-get($vf-colors-map, vf-color--grey);
```
sturobson, line 135: be nice to get this as part of a more global 'theme' so we're not calling out the specific colours (and do it for _everything_)

components/vf-content/vf-content.scss
khawkins98: 👍

sturobson (approved):

sturobson:
The 'conversations' above are more for the `vf-roadmap` really. But they made me think about them here.

PR #815, opened 2020-03-24
adds bottom margin to video in vf-content

sturobson:


components/vf-content/vf-content.scss
```diff
@@ -135,6 +135,11 @@
     color: map-get($vf-colors-map, vf-color--grey);
     font-style: italic;
   }
+
+  .vf-video {
```
khawkins98, line 128: Might be overengineering, but should we do `.vf-video + p` as this is to be used with vf-content where we wouldn't want extra spacing at the end of section ... or before and H2/H3, we wouldn't want space there, right?

Oh, and could you drop something in the changelog about this?

sturobson: ah, you've hit on something -- I guess we should cover H2/3/4/5/6 too an lists and alter accordingly.

I'll have a play and see what's a nice solution to resolve it.

sturobson: … because headings have top margin as well as bottom margin (if I recall correctly)

khawkins98: 👍

khawkins98:
Approving. I'm indifferent on if we should tackle the spacing now or as part of a more holistic thing.

khawkins98 (approved):

sturobson:
I have added a little bit of 'tidying up' too.

This PR also now:

- removes `display: block` from `H1-6` and `p` elements as that is their browser default
- moved `margin-bottom` changes per element to the `@include set-type` using `$custom-margin-bottom`

PR #819, opened 2020-03-30
Enhancement: optimise vf-form labels

khawkins98:
Following the call with Gabby, this removes needed for vf-form__float-label to improve UX.

It updates the default component demo to use an `--example`

image

Eventually, we should also consider if these should have usable NJK templates, but that's probably more work than would be helpful at the current time.


khawkins98:
Picked up in #821

Issue #820, opened 2020-03-30
vf-form refactor


sturobson:
Thins that I'd like to include in a refactor of `vf-form--` elements, this list is likely to see additions 

- [x] remove floating label JavaScript
- [x] update form label style
- [x] create `vf-fieldset` similar to [this issue](https://github.com/visual-framework/vf-core/issues/762) 
- [x] create -vf-legend` similar to [this issue](https://github.com/visual-framework/vf-core/issues/762) 
- [x] spacing
- [x] colours and colour contrast
- [x] error states
- [x] required label
- [ ] documentation

sturobson:
Can be tracked here – #821

PR #821, opened 2020-03-30
the one time we refactored all the form things

sturobson:
issue here - https://github.com/visual-framework/vf-core/issues/820


components/vf-form/vf-form__core/assets/vf-form__float-labels.js
```diff
@@ -1,127 +0,0 @@
-// vf-form__float-labels
```
khawkins98, line 1: Rather than delete all the JS, it would be better if we left the function with a console.log or .warn, noting that your need to update your scripts.js rollup. That way a component update won't break the site's JS and alert the dev they need to update.

sturobson: added to this PR now

components/vf-form/vf-form__fieldset/README.md
```diff
@@ -0,0 +1,59 @@
+👋 Hi component authour, you've read the docs, right?
```
khawkins98: These first 12 lines of boilerplate got left in. I should probably add a note to delete this.

components/vf-form/vf-form__helper/vf-form__helper.scss
```diff
@@ -13,10 +13,10 @@
...
   color: set-color(vf-color--grey);
+}
+
 
-  .vf-form__floatlabel & {
-    left: 24px;
-    position: relative;
-    top: 12px;
-  }
+.vf-form__helper--error {
```
khawkins98, line 19: Should we add any CSS for form element `:invalid` states ... which I guess would be turning the border red or something? (perhaps similar to `vf-form__input:disabled`)

khawkins98 (commented):
A couple of other misc:

- we should get the changleogs updated (happy to help)
- vf-search also needs updating as it has `vf-form__label vf-u-sr-only | vf-search__label`

components/vf-form/vf-form__helper/vf-form__helper.scss
sturobson: good idea - will add

sturobson:
added required/`:invliad` styles and example

Screenshot 2020-03-31 at 14 09 37

components/vf-form/vf-form__fieldset/README.md
sturobson: or it's a pre-commit to check if it's there and ask to delete it in the cli?

sturobson: because search has the action in the button text - I think we can do without having the label visible. 

I've snuck in a slight CSS change for the inline version so the button lines up nicely.

khawkins98 (approved):

sturobson:
I've merged this in now so it's ready to be used after our meeting with EBI nG developers today.

Issue #726, opened 2020-01-17
proposal - vf-links colours


sturobson:
With the launch of the new embl.org landing page - we need to take another look at (generic) link colours.

PR #822, opened 2020-03-31
The one that updates article meta information

sturobson:
This PR updates the meta information for articles component. 

- removes any usage of `vf-text--` classnames and applies component specific typography as needed
- moves around the content order a little
- makes use of the `vf-links-list` `--tight` variant

todo (when `vf-summary` gets merged and updated) - replace the avatar with `vf-summary--profile` smaller variant.

### DIFF
--- a/components/vf-article-meta-information/vf-article-meta-information.scss
+++ b/components/vf-article-meta-information/vf-article-meta-information.scss
@@ -29,9 +29,9 @@
 
   .vf-author__name {
     @include set-type(text-body--3, $custom-margin-bottom: 0);
-    line-height: 1.3;
 
     align-self: center;
+    line-height: 1.3;
     margin: 0 0 0 8px;
     word-spacing: 100vw; // so first name and surname split
   }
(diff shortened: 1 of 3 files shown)


components/vf-article-meta-information/vf-article-meta-information.scss
```diff
@@ -107,9 +106,8 @@
 
 
 .vf-meta__comments {
-
-  .vf-text-body--6 {
+  p {
```
khawkins98: does this have to be a vanilla `p` tag, can we do a class or change the html structure?

khawkins98:
We'll probably get bitten by `:visited` links looking a bit silly here, but I suppose that's _really_ an issue with needing to only have purple visited links in long-form content areas. I think that's what #726 is about

components/vf-article-meta-information/vf-article-meta-information.scss
sturobson: resolved that in last commit - b6fc7cf

sturobson:
re: link colours - yes - more than likely there'll be some oddities until this gets resolved

components/vf-article-meta-information/vf-article-meta-information.scss
khawkins98: 👍

khawkins98 (approved):

PR #825, opened 2020-04-02
the quick one that fixes colour problems with summary links

sturobson:
since updating how the `inline-links` Sass mixin is used to create link styles - we had an issue in how this effected the profile summaries. This fixes that.


khawkins98 (approved):
Looks good. Just needs the changelog bit, happy to add that.

PR #827, opened 2020-04-03
the one that adds some more banners

sturobson:
We add 3 new 'alert banners' that can be used with a button to dismiss them.


sturobson:
@khawkins98 I don't know if you had time/want to add anything regarding JS here? I have left this as a 'BYOJS' to avoid any possible conflicts.

components/vf-banner/vf-banner--states.scss
```diff
@@ -0,0 +1,43 @@
...
+    color: currentColor;
+  }
+}
+
+.vf-banner--success {
+  background-color: set-color(vf-color--green--light);
+  color: set-ui-color(vf-ui-color--black);
+}
+
+.vf-banner--alert {
```
khawkins98: As .vf-banner--alert is the "general" class, it should come first in the Sass or else there are specificity issues:

image

components/vf-banner/vf-banner--states.scss
```diff
@@ -0,0 +1,43 @@
...
+.vf-banner__button {
+  margin-left: auto;
+
+  svg {
+    fill: currentColor;
+    height: 14px;
+    transform-origin: center center;
+    transition: transform 250ms linear;
+  }
+  &:hover {
```
khawkins98: fun spin :)

khawkins98: Suggestion, we also have some stroke/shadow on hover to indicate activity. 

Also probably a `:focus` state.

Maybe the above is general enough that it should go on `. vf-button--icon`?

components/vf-banner/vf-banner--states.scss
sturobson: kinda related(?): I was wondering if we should phase out (don't laugh) `vf-banner--phase` and instead have it as `vf-banner--alert vf-banner--info` or something.

sturobson: because we need a generic class for the padding (which `vf-banner--phase` has) but I think it's confusing to have `vf-banner--phase vf-banner--success`

khawkins98 (commented):
Left a couple of comments. Also a thought, are these mixin classes, in that we should be able to add `vf-banner--alert` to  a `vf-banner--top`?

components/vf-banner/vf-banner--states.scss
sturobson: I've duplicated the hover spin for now.

components/vf-banner/vf-banner--states.scss
sturobson: I've included a `vf-banner--info` variant. Changed `vf-banner--alert vf-banner--alert` duplication so the red one is `vf-banner--danger`

sturobson:
Perhaps we should 

- deprecate `vf-banner--phase` in place for `vf-banner--alert vf-banner--info`
- make use of the yaml to create the alert variants (colours and with/without buttons (I guess the option to not dismiss is only really option on the `--info` alert.

components/vf-banner/vf-banner--states.scss
khawkins98: That makes sense. 

On the `:hover` `:focus` I was thinking more accessibility if you miss the animation while tabbing or hovering, something like:

(tab first, then hover)

Screen Shot 2020-04-03 at 21 48 48

```css
.vf-banner__button:focus {
    border: 1px solid white;
}

.vf-banner__button:hover {
    border-bottom: 1px solid white;
}
```

sturobson: ok, so maybe for `:focus` but that border on `:hover` is a little EBI/VF1.x—y

Also - I’ll look at adding the text-crop to `vf-core` so we can make the box actually square for `:focus` because it looks a little ugly.

If we _really_ need something like a border bottom on `:hover` perhaps we need it to look more like a button than it does now (it has no styling from researching several design systms alerts styles).

I’ll try and do both - the text-crop so the square outline is actually _square_ and also an example of it being a button, perhaps something that’s inset - but that kinda goes against current aesthetics.

sturobson: I mean, in saying some of that - the hover effect isn’t really part of the system either :sob: 

I’m sure there’s a good compromise out there :)

components/vf-banner/vf-banner--states.scss
sturobson: Resolving this as we’ve moved it to another conversation in this PR

components/vf-banner/vf-banner--states.scss
khawkins98: Thinking about this, why do we _have_ to use an icon? Just as well or better to use "Close" or "Hide". 

For example, opening www.gov.uk in an incognito. It's a longer notification, but (as usual) their use of text works better most other systems.

The text is only a few characters longer (plus more explicit) and will bypass issues discussed above -- it will also reinforce a principle of avoiding icons wherever possible.

sturobson: I'm sure a banner will take be larger on a small viewport, and there would be less text for 'hide' rather than a `x` icon. 

re: hover/focus.

As it's a button there's the cursor change on hover (although touch devices will have to guess).

I've tidied up the `:focus` state so that it's a nice square box.

sturobson:
> Left a couple of comments. Also a thought, are these mixin classes, in that we should be able to add `vf-banner--alert` to a `vf-banner--top`?

We _should_ be able to. `vf-banner--alert` is only adding padding to `vf-banner__content` (which could be 'over engineering' (we could look at doing the padding as a default with custom properties)).

The colours from the alert type should work too. 

Things like this make me want to refactor everything to use CSS custom properties :sob:

sturobson:
I'll try and tidy this link `:visited` colouring too … but this is part of a bigger thing.

Screenshot 2020-04-06 at 09 19 02

khawkins98:
> I'll try and tidy this link :visited colouring too … but this is part of a bigger thing.

Part of me wonders if we should just go ahead and make the default for visited be a less obvious grey, adding purple visited links for `vf-content` 

That would probably require less overrides?

components/vf-banner/vf-banner--states.scss
khawkins98: Probably more than fair for now. As we don't even have a real use case for dismissing notifications (beyond the we-probably-need-it from HCA), we're at risk of premature optimisation.

sturobson:
re: the `--phase` banner currently in place for www.embl.org - the HTML needs a tidy up really. 

```
<div class="vf-banner vf-banner--phase">
  <div class="vf-banner__content">
    <p class="vf-text-body--3"><a href="https://www.embl.org/news/lab-matters/corona-updates/">EMBL’s response to the coronavirus outbreak</a></p>
  </div>
</div>
```
should become –
```
<div class="vf-banner vf-banner--alert vf-banner--info">
  <div class="vf-banner__content">
    <p class="vf-banner__text">
      <a href="https://www.embl.org/news/lab-matters/corona-updates/">EMBL’s response to the coronavirus outbreak</a>
    </p>
  </div>
</div>
```

when this PR is merged and the code included in the www.embl.org repo.

sturobson:
> Part of me wonders if we should just go ahead and make the default for visited be a less obvious grey, adding purple visited links for vf-content

this is done in #835

sturobson:
I think we can merge this, `--phase` CSS is still in place so it won't break the existing uses.

khawkins98:
> re: the --phase banner currently in place for www.embl.org - the HTML needs a tidy up really.

We'll get that for free as the notification banners will get updated as we move to the embl-notifcations.

khawkins98 (approved):

sturobson:
> We'll get that for free as the notification banners will get updated as we move to the embl-notifcations.

Excellent.

excellent

PR #830, opened 2020-04-03
the one that adds tables to the component library

sturobson:
initial commit includes:
- nunjucks and yaml logic setup
- styles added for bordered, striped, loose, and tight tables

todo (not an exhaustive list):

- [ ] documentation
- [ ] what to do with numerical data styling
- [ ] sort fixed header styles
- [ ] `vf-table__footer` element
- [ ] better example data in the `.yml` file.

### DIFF
--- a/components/vf-table/vf-table.config.yml
+++ b/components/vf-table/vf-table.config.yml
@@ -3,23 +3,53 @@ title: Table
 # Label shown on index pages
 label: Table
 status: alpha
-# The template used from /components/_previews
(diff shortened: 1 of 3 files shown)


khawkins98:
This is a good start – we could go with it and move the remaining tickboxes to an issue. 

What you've got here seems very usable ... the whole, "don't let great stand in the way of good".

khawkins98 (approved):

PR #832, opened 2020-04-04
Bug: vf-form__core package missing component references

khawkins98:
vf-form__core/vf-form__core.scss requires `__fieldset` and `__legend` but they are missing from the package.json


sturobson (approved):
Oops - again - a pre-commit or pre-push checker would be good to stop me missing this.

PR #835, opened 2020-04-06
the one where we look at link colours, again.

sturobson:
More often than not the `:visited` style for links that is designed makes the page look a little 'off' as links in the page UI can have sprinkles of purple. Here we remove keep the `:visited` styles the same age `:link` for the generic mixin `@include inline-link` and alter as needed (for `--secondary`) including having `:visited` links for long form content in `vf-content`.

- new link colours for UI, or components that use the default `@include inline-link` mixin.

Screen Shot 2020-04-06 at 10 37 27

- an example of a visited link in long form content.

Screenshot 2020-04-06 at 10 38 11

Because components are generally using `@include inline-link` there should be nothing that dramatically explodes.

### DIFF
--- a/components/vf-content/README.md
+++ b/components/vf-content/README.md
@@ -4,24 +4,15 @@
 
 ## About
 
-Use this container to get simple support for narrative content where it is not
-practical to assign classes, such as Markdown or WYSIWYG text.
+This container adds support for long form content where it may not be practical to assign classes, such as Markdown or WYSIWYG text.
+
+It also makes some adjustments for longer form text, such as vertical spacing and making visited links purple.
 
 This container adds basic support for `p`, `ul`, `hr`, `a` and other core
-html elements. Some components may also add specific support for `.vf-content`
+html elements.
 
-### Demonstration
+Some components may also add specific support for `.vf-content`
 
-<div class="vf-content">
-{{ "#### Sub-header" | marked }}
-
-{% markdown %}
-- list items
-- list items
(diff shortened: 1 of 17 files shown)


khawkins98 (commented):
Looks good, a few thoughts:

1. Suggest in `vf-content` we add a `.vf-content .vf-link` so if a link does have the vf-link class, it still gets the purple visited state.
1. Should vf-content be reclassified from a `utility` to a `container`? As it's not going beyond a "I can't apply the classes" to a "this is long-form content and should behave differently".
1. We should probably update the vf-content readme, something like:

> This container adds support for long form content where it may not be practical to 
assign classes, such as Markdown or WYSIWYG text. It also makes some adjustments for 
longer form text, such as vertical spacing and making visited links purple.

sturobson:
> 1. Suggest in vf-content we add a .vf-content .vf-link so if a link does have the vf-link class, it still gets the purple visited state.

Yes, added.

> 2. Should vf-content be reclassified from a utility to a container? As it's not going beyond a "I can't apply the classes" to a "this is long-form content and should behave differently".

I think so, this would lead well into the separation of 'component' and 'content' in regards to the typography that I want to get to with the text-crop tool.

3. We should probably update the vf-content readme, something like:

Yes, will revise.

I'll ping you when they're ready.

khawkins98 (approved):

khawkins98:
Looks good. Merge when ready :)

PR #836, opened 2020-04-07
the one that makes header components a little better

sturobson:
We want to remove the reliance on CSS grid to define the size of the `<header>` components as we cannot always rely on a direct parent/child relationship. 

Here we move to a better supported, simpler (more verbose) solution using –

```
  box-sizing: border-box;
  margin: 0 auto;
  max-width: 78.5em;
  padding: 0 1rem;
  width: 100%;
```

### DIFF
--- a/components/vf-global-header/vf-global-header.scss
+++ b/components/vf-global-header/vf-global-header.scss
@@ -15,18 +15,17 @@ $vf-global-header__link-text--color: set-ui-color(vf-ui-color--black);
 .vf-global-header {
   align-items: center;
   background-color: $vf-global-header-background--color;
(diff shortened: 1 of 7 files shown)


khawkins98 (approved):

khawkins98:
👍 Merge when ready.

sturobson:
I'm going to merge this one now and then get around to a new PR that does the similar code switches for anything else that was relying on `grid-column: 2 / -2` or `grid-column: main` to use –

```
  box-sizing: border-box;
  margin: 0 auto;
  max-width: 78.5em;
  padding: 0 1rem;
  width: 100%;
```

as needed

PR #837, opened 2020-04-08
the one where there was a disco bug on the vf-tree's show/hide button

sturobson:
At some point the `vf-tree` components buttons had some crazy problems with the `vf-button` styles leaking into the 'unstyled' buttons.

This fixes that.

It's probably not possible now but I'd love to have some simple reset on `vf-button` and we layer on top the styles as needed.


khawkins98 (approved):

khawkins98:
👍 Merge when ready.

PR #838, opened 2020-04-08
Bug: EMBL favicon and breadcrumbs

khawkins98:
Resolves a few bugs I've seen while using these njk templates in production

- embl-favicon missing a link to favicon.ico (some browsers seem to want it)
- embl-breadcrumb-lookup.njk was loading demo `embl-content-meta-properties` that were only needed for an example

Also adds some better defaults to the site.webmanifest

PR #839, opened 2020-04-08
Bug/vf logo homepage

khawkins98:
Accidently based off my prior PR, but updates the logo on the component library homepage is currently busted. We should use the CDN asset when deployed.

If we merge after #838, all will be fine.

image


sturobson (approved):

sturobson:
👍 you may merge, when ready.


charming to the last

PR #842, opened 2020-04-10
The one that fixes text inputs on iOS

### LINKED ISSUE #831: BUG: vf-form components get some additional styles in iOS/iPad
I was just looking at the component library on my iPad and have noticed:

`vf-input—text` has slightly rounded corners and an inner shadow.
`vf-input—textarea` has slightly rounded corners and an inner shadow

sturobson:
iOS has a 'fun quirk' in that it adds additional CSS styles to forms because… why not‽ 

We had managed to catch most of these except for a couple of things on `vf-form__input` and `vf-form__textarea`. 

This pull request adds the below CSS rules to both components.

```
appearance: none;
border-radius: 0;
```

note: `appearance` needs to be prefixed in (at least) iOS with `-webkit-`. This is done on `build` but not `dev` tasks for local development.

This will close #831

### DIFF
--- a/components/vf-form/vf-form__input/vf-form__input.scss
+++ b/components/vf-form/vf-form__input/vf-form__input.scss
@@ -15,8 +15,8 @@
   appearance: none;
   border-color: set-ui-color(vf-ui-color--grey);
   border-style: solid;
-  border-width: 3px;
   border-radius: 0;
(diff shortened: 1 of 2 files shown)


khawkins98 (approved):

khawkins98:
👍

PR #843, opened 2020-04-10
The one where we make our CSS grid-gaps gaps

### LINKED ISSUE #826: ENHANCEMENT - replace grid/grid-row/grid-column-gap with gap/column-gap/row-gap
It appears that `grid-gap`, `grid-column-gap`, and `grid-row-gap` are [being deprecated (see note)](https://developer.mozilla.org/en-US/docs/Web/CSS/gap) for just 'gap` and variations there of.

This is also inline with when `display: flex` has `gap` supported in all browsers (currently only in Firefox). 

I think this is a good 1st PR for someone or a quick task to fill the void when time arises.

- [ ] replace `grid-gap` with `gap`
- [ ] replace `grid-column-gap` with `column-gap`
- [ ] replace `grid-row-gap` with `row-gap`

sturobson:
Whilst reading an article on grid I cam across [this comment from Rachel Andrew](https://css-tricks.com/4-css-grid-properties-and-one-value-for-most-of-your-layout-needs/#comment-1755417) about how we should be able to mov away from the, now deprecated, `grid-column/row-gap` and use `gap` without any prefixes. 

This PR does this, tested in Safari, Firefox, and Chrome

closes #826

### DIFF
--- a/components/vf-sass-config/variables/vf-global-variables.scss
+++ b/components/vf-sass-config/variables/vf-global-variables.scss
@@ -8,7 +8,7 @@ $global-font-family: $vf-font-family--sans-serif !default;
 $vf-text-margin--bottom: 16px !default;
 
 // grid variables
-$global-column-gap: 1em !default;
+$global-grid-column-gap: 1em !default;
 $global-page-max-width: 76.5em !default;
 
 // deprecation variables
--- a/components/vf-sass-config/variables/vf-global-variables.scss
+++ b/components/vf-sass-config/variables/vf-global-variables.scss
@@ -8,7 +8,7 @@ $global-font-family: $vf-font-family--sans-serif !default;
 $vf-text-margin--bottom: 16px !default;
 
 // grid variables
-$global-grid-column-gap: 1em !default;
+$global-column-gap: 1em !default;
 $global-page-max-width: 76.5em !default;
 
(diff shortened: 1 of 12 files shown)


components/vf-sass-config/variables/vf-global-variables.scss
```diff
@@ -8,7 +8,7 @@ $global-font-family: $vf-font-family--sans-serif !default;
 $vf-text-margin--bottom: 16px !default;
 
 // grid variables
-$global-grid-column-gap: 1em !default;
+$global-column-gap: 1em !default;
```
khawkins98: I don't think most use cases would require it but also add 

```
$global-grid-column-gap: 1em !default; // legacy support, will be removed in v1.some release
```

Will give us insurance if an older component needs to be used for some reason

sturobson: Dang, I was trying not to effect the Sass variables at all. I’m not a fan of how the CSSWG has moved towards a single rule that effects both grid and flex box, I would have much preferred grid-gap and flex-gap so we can distinguish them - so my hope was to keep Sass variables explicit in what the gap is for … 

So I’ll change this back to `global-grid-column-gap` again

khawkins98 (approved):
👍

PR #844, opened 2020-04-10
the one that adds the max-width to images in components

### LINKED ISSUE #816: vf-figure__image missing max-width: 100%
since removing a a CSS reset it looks like there's the odd bug that's passed through the crack.

- [ ] investigate all `__image` elements to see if they need this CSS rule
- [ ] add the rules, as needed
- [ ] ship it

sturobson:
relating to and will close #816

This PR adds the below, as needed, to components:

```
height: auto;
max-width: 100%;
width: 100%
```

note: tested in the `vf-core` UI across browsers.

### DIFF
--- a/components/vf-figure/vf-figure.scss
+++ b/components/vf-figure/vf-figure.scss
@@ -15,5 +15,6 @@
 
 .vf-figure__image {
   height: auto;
+  max-width: 100%;
   width: 100%;
 }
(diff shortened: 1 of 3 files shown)


khawkins98 (approved):
Looks good. 

Is it over-engineering to make this into some kind of `vf-image-normalise` mixin? It probably is.

sturobson:
> Is it over-engineering to make this into some kind of vf-image-normalise mixin? It probably is.

It would at least mean that we are making sure that any image has the correct CSS rather than hoping I remember to add `max-width: 100%;` :D

PR #847, opened 2020-04-10
the one that removes a parent class for the screen reader only class

### LINKED ISSUE #665: .vf-sr-only displays content
Example:
https://visual-framework.github.io/vf-core/components/detail/vf-pagination.html

The 'page' here is displayed:
<span class="vf-sr-only"> page</span>

sturobson:
related to and will close #665 

I cannot recall why the class for hiding text for screen readers - `vf-u-sr-only` - was a child of the class `vf-js` and I cannot think of a good reason now. 

This removes the need for there to be JS working or not for `vf-u-sr-only` to work.


khawkins98 (approved):
Yeah, agree. I think there was some logic at work, but this seems the better route with hindsight.

PR #853, opened 2020-04-14
Chore: captilisation of "In this page"

khawkins98:
Stylistically we want  "In this page".

Also gets a typo in the placeholder class of `vf-icon__arrown--down`

Also assigning @filipe-dias so he can see what some of our PRs look like,


sturobson:
> Stylistically we want "In this page"

I think we should add "include written style guide to documentation" as well as guides for micro content like this. 

When we have … time.

khawkins98:
Yeah, there's a load of little things around text styling to be addressed from a documentation need. Hopefully the EMBL content style guide isn't too far away and we can leverage that ...

sturobson (approved):

PR #855, opened 2020-04-15
the one that puts link list squares inside the box model

### LINKED ISSUE #854: CHORE: Bullet wrapping should be "outside" on Links List: Very Easy
Noticed on https://wwwdev.embl.org/partnerships/ the [Links List: Very Easy](https://visual-framework.github.io/vf-core/components/detail/vf-link-list--very-easy.html) bullets wrap under:

image

Would be better as:

image

Achieved in the above as:

```
.vf-links__list--s .vf-list__item {
  list-style-position: outside;
  margin-left: 14px;
}
```

sturobson:
we make the list style type inside, and use margin to push the whole list 1em from the left to make it not break

This will close #854

### DIFF
--- a/components/vf-link-list/vf-link-list.scss
+++ b/components/vf-link-list/vf-link-list.scss
@@ -101,8 +101,9 @@
 
 
 .vf-links__list--very-easy .vf-links__list {
-  list-style-position: inside;
+  list-style-position: outside;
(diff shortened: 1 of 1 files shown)


khawkins98 (approved):
👍

PR #860, opened 2020-04-16
the one that adds lots of more table things

sturobson:
Now we have the `vf-table` there is a need for some more advanced options so the component can be used in more places. 

We include the HTML and CSS so that you can:

- use selectable table rows.
- add actions for a row or as a toolbar for a selection of rows.
- adds a footer.
- adds 'sortable' styles with a (currently) inline SVG of icons.
- better examples of 'one variant' doing 'one thing' and then a 'kitchen sink'.
- a small start to the documentation.

This PR does not include any JS functionality for the tables.


khawkins98:
Your nagging reminder for the changelog 😅

khawkins98 (approved):

PR #861, opened 2020-04-16
the one that fixes the hover colours in tables

sturobson:


sturobson:
oh, I branched from the enhancement/vf-tables branch - which is the git history here is a little messy.

khawkins98 (approved):
👍

But it looks like you've gotten into some merge conflict fun?

PR #864, opened 2020-04-17
changes required astericks colour

sturobson:
I noticed for some project mockups that a blue asterisk was use - I think this is nicer than the red currently used for required form inputs. 

Red is like 'you are wrong, fill this out' before you've even started.

### DIFF
--- a/components/vf-form/vf-form__label/CHANGELOG.md
+++ b/components/vf-form/vf-form__label/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change Log
 
+## 1.1.0
+
+- changes the required asterick colour from red to blue, as red is more of an 'error' colour. 
+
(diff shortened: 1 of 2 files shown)


khawkins98:
Nice thought. 

Aside: when it comes to more visual changes, is there guidance on what should be a .0.1 bump in version Vs a .1.0 bump?

khawkins98 (approved):

sturobson:
> is there guidance on what should be a .0.1 bump in version Vs a .1.0 bump?

Not _really_ 

I’ve wanted us to follow semver - but with all the breaking changes before we got to 2.0.0 we would probably be on VFCore v48 :sob:

Perhaps this is something to investigate and ratify in the docs? I’ll start investigating over the weekend.

PR #867, opened 2020-04-20
fix: embl-breadcrumbs-inferences

### LINKED ISSUE #862: BUG - embl-breadcrumbs-lookup inference uses wrong label
When one of the three who/what/where ancestors is left off, `embl-breadcrumbs-lookup` tries to infer. It seems to be finding the right ancestor but uses the wrong label:

Example:

```html
<meta name="embl:where" content="EMBL Heidelberg" uuid="f36a2275-0292-4ddf-9784-9f9f681915f8">
<meta name="embl:what" content="Cuylen Group" uuid="f902c031-b698-4479-957c-644ff57a782e">
<meta name="embl:active" content="what">
```

Results in a `who` crumb of:

image

Medium priority as this only occurs when an omission is present.

khawkins98:
Closes #862

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,8 +21,8 @@ jobs:
       env:
         - PERCY_TARGET_BRANCH=develop
       script:
-        - ls -la build/components/preview/*.html
-        - npx percy snapshot build/components/preview/
+        - ls -la ./build/components/preview/*.html
+        - npx percy snapshot ./build/components/preview/
       branches:
         except:
         - develop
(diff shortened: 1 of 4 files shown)


sturobson (approved):

khawkins98:
Still in progress but ticking ready for review to test percy.

khawkins98:
Finally got my head wrapped around this simple fix.

The taxonomy lookup could probably use a refactor to make the code more streamlined now that we're better understanding the ontology.

PR #872, opened 2020-04-20
Testing: percy branches

khawkins98:
Right now it's only comparing develop branch to develop, this will compare all branches (the expected behaviour)


khawkins98:
(we might want to move to Github's CI testing to do this, but for now this would probably be faster)

sturobson (approved):

PR #873, opened 2020-04-20
the one that makes section header links pass some #a11y rules

sturobson:
I ran https://www.covid19dataportal.org/ through Firefox's accessibility panel and found a glaring issue with the `vf-section-header` when it is a link. The SVG was being seen by assistive technologies and was reporting a missing label.

As the SVG is a visual queue only I have added `aria-role="hidden"` to the SVG in the component to fix it.

### DIFF
--- a/components/vf-section-header/CHANGELOG.md
+++ b/components/vf-section-header/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Change Log
 
+## 1.1.1
+
+* SVG is a visual queue: adds aria-role="hidden" to the SVG in the component so screen readers don't announce it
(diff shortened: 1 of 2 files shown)


khawkins98:
I'm guessing you already checked, but to be sure: the other places we've used SVG in a decorative way, have we also added `aria-role="hidden"`?

khawkins98 (approved):

sturobson:
I can do a ‘full sweep’ tomorrow on this. There’s a chance that they have not been included. I felt this one was pertinent enough to do by itself because of it’s usage.

PR #879, opened 2020-04-22
BUG - Add underline to links in vf-banner

khawkins98:
Links in `vf-banner` were appearing without distinction from the surrounding text.

image

Also adds not that `vf-text` usage in `vf-banner` is deprecated (already noted elsewhere in the Sass)

### DIFF
--- a/components/vf-banner/vf-banner.scss
+++ b/components/vf-banner/vf-banner.scss
@@ -21,7 +21,7 @@
 }
 
 // all vf-text links should be underlined in banners
-.vf-banner__text .vf-link,
+.vf-banner__text .vf-banner__link,
 .vf-banner [class*='vf-text'] .vf-link { // as of 1.0.4 use of vf-text is not recommended and is subject to future removal
   text-decoration: underline;
 }
@@ -77,6 +77,7 @@
     margin-bottom: 24px;
     max-width: 64em;
 
+    & .vf-banner__link,
     & .vf-link {
       color: $vf-banner-color--link;
     }
--- a/components/vf-banner/vf-banner.scss
+++ b/components/vf-banner/vf-banner.scss
@@ -21,7 +21,8 @@
 }
 
 // all vf-text links should be underlined in banners
-.vf-banner [class*='vf-text'] .vf-link {
+.vf-banner__text .vf-link,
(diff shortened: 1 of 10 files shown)


sturobson:
hmmm, perhaps this is a documentation/implementation issue. The examples currently in `vf-core`'s documentation don't use `vf-link`

`vf-banner` text currently has it so that [the `<a>` tag](https://github.com/visual-framework/vf-core/blob/develop/components/vf-banner/vf-banner--alerts.scss#L9) gets [the correct styling](https://visual-framework.github.io/vf-core/components/detail/vf-banner--success.html) so that we don't rely on `vf-link` here. 

Perhaps we could do `vf-banner__link | vf-link` and get it styled

I didn't originally use `vf-link` because of the overrides that could be needed as the link text colour would/should be the same as the rest of the text. 

Or we could get more a11y issues from it:

Screenshot 2020-04-22 at 13 30 35

khawkins98:
With the way vf-link has evolved `vf-banner__link` might make better sense, and allow us to not have to rely on styling raw HTML elements.

khawkins98:
In the meantime, I can sprinkle some `vf-text` classes to embl-notifcations so links are visible.

sturobson:
I wonder if `.vf-banner__link { color: currentColor; }` would trump whatever is in `vf-link`.

The import order we use for `vf-core` has `vf-banner` being imported after `vf-link` so we could 'get away' with just the above. But I think we shouldn't rely on others using the same manifest so we would need to scope it to `.vf-banner vf-banner__link { }`. I think.

sturobson:
noting that Percy only runs when the PR is first created and not when subsequent changes are made.

components/vf-banner/vf-banner.scss
```diff
@@ -21,7 +21,8 @@
 }
 
 // all vf-text links should be underlined in banners
-.vf-banner [class*='vf-text'] .vf-link {
+.vf-banner__text .vf-link,
```
sturobson: could we not do `vf-banner__link` here too so we don't have to get back to it?

sturobson (approved):

khawkins98:
In 92c0b64 and 6edfe07 I've updated to use vf-banner__link and also snuck in an IE11 fix on assigning classes.

PR #880, opened 2020-04-22
Workaround: add vf-text to embl-notifications so links are visible

khawkins98:
See: https://github.com/visual-framework/vf-core/pull/879#issuecomment-617752913


sturobson (approved):

khawkins98:
With the updates to #879, I don't think this will be needed.

PR #882, opened 2020-04-22
the one where we get IE11 working better, again.

### LINKED ISSUE #881: BUG - IE11 does not support calc() inside of the flex property.
In `vf-grid` we make use of `calc()` to determine the size of components for browsers that do not support CSS grid (oh, hello IE11).

We need to 'hard code' the values for when we want 2, 3, 4, 5, or 6 column grids.

sturobson:
This should close #881 

IE11 does not support `calc()` inside of `flex` because, why not‽ 

This fixes the issue that was reported where the size of grid items was being ignored because `calc()` was not working.

Here we have 'hard coded' the percentage values.

Testing on IE11 before moving from Draft.

### DIFF
--- a/components/vf-grid/vf-grid.scss
+++ b/components/vf-grid/vf-grid.scss
@@ -18,7 +18,7 @@
   margin: 10px 0 10px 1.6339%;
 }
 
-.vf-grid >*:first-child {
+.vf-grid > *:first-child {
   margin-left: 0;
 }
 
--- a/components/vf-grid/vf-grid.scss
+++ b/components/vf-grid/vf-grid.scss
@@ -26,6 +26,9 @@
 [class*='vf-grid__'] {
   flex-wrap: wrap;
 }
+// Originally the calculations in the flex declarations below were created using CSS calc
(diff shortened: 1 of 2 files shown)


khawkins98:
Looks good, two thoughts:

- changelog 😅
- would be good to have a comment in the Sass that this has been done without calc as another ie11 measure

components/vf-grid/vf-grid.scss
```diff
@@ -15,25 +15,31 @@
 }
 .vf-grid > * {
   flex: 1;
-  margin: 10px;
+  margin: 10px 0 10px 1.6339%;
```
sturobson, line 18: I’m thinking that we should remove the vertical margins here and move towards parent elements dictating spacing.

sturobson: Especially as we have most of this getting reset when inside the `@supports` for grid

sturobson: I've left this 'as is' for now. To hopefully get this into develop to fix the current issue at hand.

khawkins98 (approved):

PR #888, opened 2020-04-28
the one that adds a simple social links component

sturobson:
todo:

- [x] changelog
- [x] documentation


components/vf-social-links/vf-social-links.njk
```diff
@@ -0,0 +1,50 @@
+<svg display="none" class="vf-icon-collection vf-icon-collection--social">
```
khawkins98: Do we need `title` attributes on these for screen readers?

sturobson: I'll have a look - the SVG is only used for the other SVGs in the links. 

And writing that they probably need aria-role="hidden"

components/vf-social-links/vf-social-links.scss
```diff
@@ -0,0 +1,120 @@
...
+ */
+
+@import 'vf-social-links.variables.scss';
+
+
+.vf-social-links__heading {
+  @include set-type(text-heading--4);
+
+  color: set-ui-color(vf-ui-color--black);
+  text-transform: capitalize;
```
khawkins98: EMBL's (undocumented) style is only for sentence caps, which I don't think there is a `text-transform` for.

components/vf-social-links/README.md
```diff
@@ -0,0 +1,51 @@
...
+[![npm version](https://badge.fury.io/js/%40visual-framework%2Fvf-social-links.svg)](https://badge.fury.io/js/%40visual-framework%2Fvf-social-links)
+
+## Usage
+
+### Variants Available
+
+- `vf-social-links--outline` this removes the background and gives a border around each icon square.
+- `dark-mode` if this is insdie a component that has `dark-mode` theeme then the colours swap.
+
+The dos and don'ts of using this component.
```
khawkins98: I think this line was a leftover of the boilerplate. 

Also a though that it would be good to make clear that for the icons to show you need to add the HTML snippet to the page:

```html
<svg display="none" class="vf-icon-collection vf-icon-collection--social">
  <defs>
```

components/vf-social-links/vf-social-links.scss
sturobson: I'll change this in the actual text, remove it from the CSS and hope any other team uses it follows policy.

I guess we could do something like `p:first-letter { text-transform: capitalize; }` - but at source is best here then

components/vf-social-links/README.md
khawkins98: Or, rather, it needs to be present if you can't use the NJK template.

sturobson: > it would be good to make clear that for the icons to show you need to add the HTML snippet to the page

👍 

in lieu of working out a strategy for new vf/embl icons for this stuffs :D

khawkins98: Yeah, really want to work on that one ... but then all those other things to do 😭

khawkins98 (approved):
Good stuff. merge when ready.

PR #889, opened 2020-04-28
the one that adds some basic styling to tables for vf-content

### LINKED ISSUE #887: ENHANCEMENT - Add vf-table support to vf-content
This came up last week, capturing it here too. 

- [ ] We need a table default when using `.vf-content table`, probably the striped look a that's best for "general-purpose text"
- [ ] That look should be overridable; example: "I have a table inside vf-content I want to be compact" `.vf-content .vf-table.vf-table--compact`

Random notes/thoughts:

-  I think the above pretty consistent with the `.vf-content table:not([class*='vf-'])` approach ... but, you know, tables are always hiding surprises
- This need will probably come up during the WP group site sprint
- Should the code `.vf-content table` code live in `vf-content` or `vf-table`? For bigger components where is it more sustainable to support vf-content? That might belong in a bigger issue.

sturobson:
I think we can release this and I can look at responsive tables as something _much bigger_

This will close #887

### DIFF
--- a/components/vf-content/CHANGELOG.md
+++ b/components/vf-content/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change Log
 
+## 1.1.1
+
+- adds table styles to match default vf-table with striped rows.
+
 ## 1.1.0
 
 * `vf-content` becomes less of a utility and a general purpose for long-form content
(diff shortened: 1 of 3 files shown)


khawkins98 (approved):
👍 

Should cover all of https://github.com/visual-framework/vf-core/issues/887, I think.

PR #892, opened 2020-04-29
the one that adds simple breakpoint documentation

### LINKED ISSUE #756: DOCUMENTATION - Breakpoints missing from design token docs
The [vf-breakpoints.yml](https://github.com/visual-framework/vf-core/blob/develop/components/vf-design-tokens/src/variables/vf-breakpoints.yml) design tokens aren't in the component docs.

https://visual-framework.github.io/vf-core/components/detail/vf-design-tokens--spacing.html

Use case: need a way to see what the breakpoints are without consulting the source.

sturobson:
This adds basic documentation for the breakpoint design tokens.

It exposes the value and the Sass variable available. 

This will close #756

### DIFF
--- a/components/vf-design-tokens/dist/json/vf-breakpoints.ios.json
+++ b/components/vf-design-tokens/dist/json/vf-breakpoints.ios.json
@@ -1,0 +1,64 @@
+{
+  "properties": [
+    {
+      "type": "size",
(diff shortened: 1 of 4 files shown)


khawkins98 (approved):
point break

PR #894, opened 2020-04-29
the one that tidies up vf-masthead

### LINKED ISSUE #758: BUG - vf-masthead colour options are partly broken
With the changes to the vf-wp-group themes the options to change the masthead colours was, i suppose, deprecated (#672).

However we still have some code around using the colour keying and it's still possible to partly colourise the `vf-masthead--default`: https://codepen.io/khawkins98/pen/abOOZdq?editors=1100

In the above, the coloured background still works but text colour does not. 

Confusingly the docs still say this is supported: https://visual-framework.github.io/vf-core/components/detail/vf-masthead--default.html

I think there will still be a need to colourise the backgrounds, however the requirements to do so aren't clear at the moment.

So I think we have three options:

1. Remove all the legacy documentation and code around colourising the masthead
2. Reimplement the colour options as best we can 
3. Make some sort of vf-masthead--colourised variant

My inclination would be to do Option 1 and add a footnote in the README.md that we plan to implement a colourised masthead, and point back to this issue.

---

(This may have been discussed before, but [I didn't find any issues](https://github.com/visual-framework/vf-core/issues?utf8=%E2%9C%93&q=is%3Aissue+vf-masthead+))

sturobson:
At some point changes to `vf-masthead` left things a little broken. 

This addresses those problems in setting good defaults and turning towards the theming of the component locally or globally, 

I've also tidied up the variants so there's on nunjucks file

This will close #758


khawkins98:
Looks good. How much does this risk blowing up places still using `1.0.0-rc.2` (last version)?

sturobson:
Currently - for any sites that are relying on the `develop` compiled CSS it'd break - which is what it currently does.

I'll take another look at the CSS custom properties - as I think we could revert back to component-name specific ones here

sturobson:
@khawkins98 this reverts to the former naming convention for custom properties so should not break. It also now includes JavaScript to check for the 'kitchen sink' of has an image and is a title block to determine if the text colour needs to be changed or not.

khawkins98 (approved):

PR #899, opened 2020-04-30
the one that updates the vf-embed component

### LINKED ISSUE #898: some edits for vf-embed to make it more adaptable
As asked for in #893 

There's some feedback for the `vf-embed` component that would improve it's usability in WordPress (and probably everywhere else, as needed, too.

from @dbushell 

>Hey so I noticed the latest vf-embed component no longer has a generic style property, only AspectRatio and maxWidth 
> which breaks this block somewhat because it was calculating custom ratios
> so - is it possible to support custom ratios via a CSS prop like --vf-embed-ratio-width and --vf-embed-ratio-height ?

vf-embed-settings

> also the iframe width height are now hard-coded as 560 / 315 is it possible to allow these to be specified again?
> but ideally, a ratioWidth  and ratioHeight that sets a css prop like maxWidth would be perfect
> the block uses the nunjucks template  - i don’t want to be writing any css/html myself!

As a component library vf-core completely agrees with the sentiment "i don’t want to be writing any css/html myself!"

My reply to this thread in Slack included: 

> To keep with consistency it might be something like
```
.vf-embed--custom-ratio {
  /* CSS that pulls in the custom property */
}
```

sturobson:
Here we:

- creates a custom ratio variant.
- updates to use 'props' for determining what class selectors are used and if CSS custom properties are available. 
- adds documentation.
- changes the changeling.

this will closes #898

### DIFF
--- a/components/vf-embed/CHANGELOG.md
+++ b/components/vf-embed/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Change Log
 
+## 1.1.0
+
+- component now allows for custom ratios.
+- documentation is updated.
+- defining variants is changed to use 'props' in nunjucks and yaml.
+- adds demonstration variants for custom ratio, and custom ratio with a maximum width.
+
 ## 1.0.0
 
 - inital commit of a new vf-embed component.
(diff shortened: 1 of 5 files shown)


khawkins98 (approved):
Snuck in a couple of human-friendly things for the component directory.

PR #900, opened 2020-05-01
the one that finally removes calc from vf-grid fallback

sturobson:
Although we fixed the flex sizes for grid columns in #882 I missed the 'heavy handed' decision to make all grid items 50% on smaller viewports. 

This should fix that as it is not using calc anymore


khawkins98 (approved):
Looks good. added a changelog.md entry.

PR #904, opened 2020-05-04
the one that changes the vf-figure component for the better

### LINKED ISSUE #903: vf-figure alignment and sizing tweaks
As mentioned in [#85](https://github.com/visual-framework/vf-wp/issues/85) on the `vf-wp` repo there's a need for the `vf-figure` to not only `float` to `inline-start/end` but also be centred. 

There's also an issue with `vf-figure` defaulting to a width of 100%.

sturobson:
This will close #903

- We changed the use of `float` in classnames to `align`. 
- We then added a centered variant. 
- We then changed the nunjucks file to include 'props' for the variants. 
- We also added a css custom property for with. 
- We wrote some documentation.
- We updated the changeling.

### DIFF
--- a/components/vf-figure/CHANGELOG.md
+++ b/components/vf-figure/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Change Log
 
+## 1.2.0
+
+- Changes structure of nunjuck to add the ability to make use of props.
+- Changes class selector naming from `float` to `align`.
+- Adds a `align-centered` variant.
+- Adds `--vf-figure__width` as a CSS Custom Property that can be overriden.
+- Adds some documentation
+
 ## 1.1.0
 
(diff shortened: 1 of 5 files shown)


khawkins98:
Looks ok. Can't checkout on my local due to some weird clash with `develop`, hopefully that's only on my local. 

I wonder if we should make this 2.0.0 as we're removing class names? Though I guess 1.1.0 was only used on vf-wp, so not really a "real" problem.

khawkins98 (approved):

khawkins98:
"semantic versioning #905" Jinx.

sturobson:
Yeah - I went with 1.2.0 as it's only in `vf-wp` at the moment - probably wrong - but things are still moving so there's change we could get to a double figured version of a component pretty quickly :sob:

PR #921, opened 2020-05-13
the one that makes a tiny change to vf-banner

### LINKED ISSUE #920: refactor vf-banner template
From @dbushell in Slack

> is it possible to add a vf-banner Nunjucks (and precompiled) template that is just vf-banner__text and an option for info/warning/danger/success class?
> I’d like to add a Gutenberg block but vf-banner.njk is empty and it’s going to be a pain using separate templates for each type

sturobson:
This will close #920 

- makes the default `.njk` file take on the various `vf-banner--alert` variants that are available so `vf-wp` can make use of it more easily as a block.
- removes now unrequited hard-coded variants
- fixes a little hover bug on the dismiss icon in `vf-banner__button`

We have only a slight breaking change for the `vf-banner--alert` variants where we've changed the nunjucks variable from `alert_message` to `banner__message` to match semantics.

This has not been changed where used in other banners.

This update might need testing with what contentHub does with text

### DIFF
--- a/components/vf-banner/CHANGELOG.md
+++ b/components/vf-banner/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Change Log
 
+## 1.2.0
+
+* Makes the alert style banner the default .njk
+* puts the style of alert banner as a prop from yaml/nunjucks
+* removes hardcoded variants of alert banners
+* fixes hover colour of `vf-banner__button`.
+
 ## 1.1.0
 
 * Adds option to determine button theme using data attributes.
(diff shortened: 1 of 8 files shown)


sturobson:
If changing `alert_message` to `banner__message` is going to be problematic we could do

```
<p class="vf-banner__text">{{banner__message}}{{alert__message}}</p>
```

khawkins98:
I think we should be all good here from a contentHub perspective 👍

khawkins98 (approved):

PR #923, opened 2020-05-15
the one that makes the global header responsive-r

sturobson:
Screen Shot 2020-05-15 at 11 59 48
Screen Shot 2020-05-15 at 12 01 27
Screen Shot 2020-05-15 at 12 00 01
Screen Shot 2020-05-15 at 11 59 54

### DIFF
--- a/components/vf-global-header/vf-global-header.scss
+++ b/components/vf-global-header/vf-global-header.scss
@@ -23,14 +23,15 @@ $vf-global-header__link-text--color: set-ui-color(vf-ui-color--black);
   padding: .5rem;
   width: 100%;
 
-  .vf-logo {
+  .vf-logo,
+  .embl-logo {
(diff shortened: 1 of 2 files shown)


khawkins98 (approved):

components/vf-global-header/vf-global-header.scss
```diff
@@ -20,25 +20,43 @@ $vf-global-header__link-text--color: set-ui-color(vf-ui-color--black);
   height: auto;
   margin: 0 auto;
   max-width: 78.5em;
-  padding: .5rem 1rem;
+  padding: .5rem;
   width: 100%;
+
+  .vf-logo {
```
khawkins98: Does it need embl-logo too?

sturobson: Yes, I think it would

PR #927, opened 2020-05-15
the one that adds a load of vf-grid utility classes

### LINKED ISSUE #913: vf-grid enhancements
add:

- row utility classes
- add `row-gap` spacing

sturobson:
this will close #913 when approved.

I think the documentation could do with adding to - not sure how we can automate it - rather than typing out all the options by hand.

We could merge this in now - so folks can use it.


khawkins98:
👍  Looks good.

+your changelog nag 😛

khawkins98 (approved):

PR #929, opened 2020-05-18
the one where badges in a links_list with images works

sturobson:
Screenshot 2020-05-18 at 10 41 10

### DIFF
--- a/components/vf-banner/package.json
+++ b/components/vf-banner/package.json
@@ -1,5 +1,5 @@
 {
-  "version": "1.1.0",
+  "version": "1.2.0",
   "name": "@visual-framework/vf-banner",
   "description": "vf-banner component",
   "homepage": "https://visual-framework.github.io/vf-core/",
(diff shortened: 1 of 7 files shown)


sturobson:
might need an overhaul so it looks nicer if there's lots of text and you're on a mobile device.

I think the only way we can do that is with a breaking change in the markup.

Screenshot 2020-05-18 at 11 01 38

components/vf-banner/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "1.1.0",
+  "version": "1.2.0",
```
khawkins98, line 2: Are the package.json files accidental commits? They might mess with lerna?

khawkins98 (approved):
Aside from my worry on the package.json files, looks good.

components/vf-banner/package.json
sturobson: I seemed to have published it to NPM before it got merged 

https://www.npmjs.com/package/@visual-framework/vf-banner

PR #937, opened 2020-05-19
the one that removes padding and reduces margin from breadcrumbs

### LINKED ISSUE #933: vf-breadcrumbs - remove l/r padding

sturobson:
as we are using padding on the parent we do not need it here … this should also enforce good structure and use of classnames on parent elements for it to look correct

"the parent dictates the spacing of the child, or lack thereof"

This will close #933


khawkins98 (approved):
👍 

Changelog for this in #938: https://github.com/visual-framework/vf-core/pull/938/files/deb99b9eb1f6a67a8e6eaf0c27b9ee00c277125d#diff-90819e0658f11c5aa55d95a4bfbe0115

PR #938, opened 2020-05-19
the one that removes horizontal padding from the global-header

### LINKED ISSUE #934: vf-global-header remove l/r padding

sturobson:
as we are using padding on the parent we do not need it here … this should also enforce good structure and use of classnames on parent elements for it to look correct

"the parent dictates the spacing of the child, or lack thereof"

This will close #934

### DIFF
--- a/components/vf-breadcrumbs/CHANGELOG.md
+++ b/components/vf-breadcrumbs/CHANGELOG.md
@@ -3,10 +3,6 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
-# 1.0.1
-
-* removes left and right padding so we rely on the parent for horizontal spacing for better alignment
-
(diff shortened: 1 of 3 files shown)


components/vf-breadcrumbs/CHANGELOG.md
```diff
@@ -3,6 +3,10 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+# 1.0.1
```
khawkins98: 1.0.0?

khawkins98: I think this commit was meant to in #937, but I don't suppose it really matters.

> Don't ask how the sausage is made.

khawkins98:
> "the parent dictates the spacing of the child, or lack thereof"

👍  Would be good to get that in the docs.

khawkins98 (approved):

PR #939, opened 2020-05-19
the one that centres content with less code

### LINKED ISSUE #930: remove CSS grid from `vf-body`
remove all of the CSS grid code from `vf-body` and adding left/right padding

sturobson:
originally we used CSS grid to create a centred column with left and right grid items. But this is more cumbersome than centering with `margin: 0 auto;` especially now we have `vf-u-fullbleed` for things.

This closes #930

### DIFF
--- a/components/vf-grid-page/CHANGELOG.md
+++ b/components/vf-grid-page/CHANGELOG.md
@@ -3,9 +3,11 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
-# 1.0.1
+# 2.0.0
 
 * removes the CSS grid that generated a central column for content, now it's working across all browsers with less code.
+* this makes vf-u-grid-reset redundant.
+* you might need to add vf-u-fullbleed to certain containers
 
 # 1.0.0 (2019-12-17)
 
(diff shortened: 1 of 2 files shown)


components/vf-grid-page/CHANGELOG.md
```diff
@@ -3,6 +3,10 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+# 1.0.1
```
khawkins98: I think we should make this a 2.0.0 release as it will require sprinkling vf-u-fullbleed around or require updating other components, and perhaps might cause other layout breaks.

For example, `vf-divider` is no longer full-width -- or is there a separate PR for what to do with that?

sturobson: it would also require the removal of `vf-u-grid-reset` as it should no longer be needed

sturobson: I think it comes down to some good documentation -- "you need to make sure vf-body is on the body element" -- as best practice.

sturobson:
> For example, vf-divider is no longer full-width -- or is there a separate PR for what to do with that?

I'll investigate in a further PR

khawkins98 (approved):

khawkins98:
+1 Approved.

PR #940, opened 2020-05-19
the one that makes the hero snappy

### LINKED ISSUE #932: vf-hero ::before to be max-width until large viewport
width: calc(100% + 2em)

padding: 0 1em;

sturobson:
This PR makes the vf-hero component 'stick' to the left/right boundaries of the viewport until the largest breakpoint.

It looks better than having left/right padding pushing the content in.

This closes #932


khawkins98:
The positioning on the `--extreme` variant seems a bit wonky?

image

Less so, but `--easy` also looks a bit strange as the hero heading box is perhaps wider than needed: 

image

Responsiveness is also a bit broken on `--easy`:

image

I suspect these will also get change requests once "real content" is used as .... "it's covering my face, can the box go right? or can it go below the image?"

sturobson:
I've fixed the positioning for `--extreme` and `--easy`

sturobson:
@khawkins98 I think this one is ready for merging?

khawkins98:
Locally I merged in the changes from `develop` (so the `.vf-body` changes) and I noticed the heroes don't full-bleed anymore:

image

So I guess not an issue with this PR, but not sure if a change is needed here? 

--- 

Different topic: still mixed on the Jarallax, but I get that having it here makes testing easier. Can you slap some sort of "Use of this is only intended for EMBL.org landing pages" on there in big letters or something? 

That said, I'm getting used to it, maybe it just becomes a thing now.

sturobson:
>  Can you slap some sort of "Use of this is only intended for EMBL.org landing pages" on there in big letters or something?

Yes!

sturobson:
> I noticed the heroes don't full-bleed anymore 

wil take another look

sturobson:
@khawkins98 I've managed to have a look at this - I think you might have a broken version of the branch locally - as it's full-width for me.

khawkins98:
There may be some shenanigans on our machines' git branches. 

We can adjust anything as needed in a following release. 

Would be good to still get that "do not use label" but going ahead and "approving".

khawkins98 (approved):

sturobson:
@khawkins98 this should be working now? It's working fine locally - and I've changed the CSS for `vf-hero` to work with or without the grid of `vf-body` (which should be 'no grid' moving forwards.

I've also spotted that we had an erroneous use of `vf-summary__title`. This has been replaced with `vf-hero__subheading` 

note: I've introduced `[ classname ]` to denote CSS classes that will eventually disappear or shouldn't be used inside a component -- one to be documented.

PR #941, opened 2020-05-19
the one that adds a border to a vf-footer with vf-u-fullbleed

### LINKED ISSUE #931: vf-footer and vf-u-fullbleed border
add border to `vf-u-fullbleed` when it's used with `vf-footer`

sturobson:
As we have removed the CSS grid from vf-body we need to have the `vf-footer` component make use of `vf-u-fullbleed`.

This is fine and should be an easy addition for teams to update but we also need to accommodate for the `vf-footer` border. This CSS does that.

This will close #931

### DIFF
--- a/components/vf-footer/vf-footer.scss
+++ b/components/vf-footer/vf-footer.scss
@@ -17,6 +17,30 @@
   grid-column: 1 / -1;
   padding-bottom: 60px;
   padding-top: 24px;
+  position: relative;
+
+  &::before {
+    background-color: inherit;
+    content: '';
+    grid-column: 1 / -1;
+    height: 100%;
+    margin-left: calc(50% - 50vw);
+    position: absolute;
+    top: 0;
+    width: 100vw;
+    z-index: set-layer(vf-z-index--negative);
+  }
+
+  &::after {
+    background-color: set-color(vf-color--green);
+    content: '';
+    height: 8px;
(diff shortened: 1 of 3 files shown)


khawkins98:
If I follow this is needed because of #939 -- shouldn't the `vf-footer.njk` be: `class="vf-footer vf-u-fullbleed"`?

Also: I should be able to `vf-u-fullbleed` to sites that haven't updated `vf-body` and `vf-footer` without breaking their footer.

khawkins98:
to do: test updating the contentHub global embl footer with `vf-u-fullbleed` https://content.embl.org/node/569

sturobson:
> Also: I should be able to vf-u-fullbleed to sites that haven't updated vf-body and vf-footer without breaking their footer.

`vf-u-fullbleed` adds a `::before` pseudo element that lays behind the parent (in this case the `vf-footer`) so this shouldn't 'break' anything

sturobson:
> If I follow this is needed because of #939 -- shouldn't the vf-footer.njk be: class="vf-footer vf-u-fullbleed"?

Can we always guarantee people will want a fullwidth footer with centered footer content? 

If not do we need `vf-footer--fullwidth`?

sturobson:
@khawkins98 as discussed - I've moved the code for `vf-u-fullbleed` inside of `vf-footer`

khawkins98 (approved):

PR #944, opened 2020-05-20
the one that makes headings in vf-content stick to the correct font-weight

sturobson:
I took a look at this - https://www.embl.org/news/lab-matters/welcome-georgia-rapti/ - today and thought the headings looked a little 'off' in the content.

Inspecting I noticed that the headings had a wrapper of b, or strong around the text making them not have the correct font-weight.

This fixes that


khawkins98 (approved):
no i'm stronger

PR #948, opened 2020-06-02
should fix the image handling error, plus some other little fixes

sturobson:
in the last update we had `hero_bg` for images which is a breaking change to far.

This PR resorts to the original way of using yaml for images as the background image. 

Plus som other little tidy ups


khawkins98 (approved):
Looks good 👍

PR #949, opened 2020-06-03
vf_hero_link should be vf_hero_href

khawkins98:
Looks like an artefact of changes.

Also, might be good to drop vf-intro-group.png and use https://acxngcvroo.cloudimg.io/v7/https://www.embl.org/files/wp-content/uploads/vf-hero-intense.png

As the image doesn't show in the component library (https://visual-framework.github.io/vf-core/components/detail/vf-hero--tertiary.html). I think this might be due to the local vs github.io paths — and I don't think we _need_ to include this PNG asset as a default?

### DIFF
--- a/components/vf-hero/vf-hero.config.yml
+++ b/components/vf-hero/vf-hero.config.yml
@@ -13,7 +13,7 @@ preview: '@preview--nogrid'
 context:
   component-type: container
   modifier_class: vf-hero--easy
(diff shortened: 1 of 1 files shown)


sturobson (approved):

sturobson:
@khawkins98 do you want to do the url change for the `.png` in this PR before it's merged?

khawkins98:
👍

PR #951, opened 2020-06-03
the one where we are making the hero component better, again, again. 

sturobson:
fixes issue with colours
adds custom property for image height that informs grid and layout
start of documentation

### DIFF
--- a/components/vf-hero/vf-hero.config.yml
+++ b/components/vf-hero/vf-hero.config.yml
@@ -19,7 +19,7 @@ context:
     - The Hentze group combines biochemical and systems–level approaches to investigate the connections between gene expression and cell metabolism, and their roles in human disease.
   vf_hero_link_text: Learn more.
   vf_hero_heading: About the Hentze group!
-  hero__initial_row: 384px;
+  vf_hero__initial_row: 384px;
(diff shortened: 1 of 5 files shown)


components/vf-hero/vf-hero.config.yml
```diff
@@ -13,12 +13,13 @@ preview: '@preview--nogrid'
...
   component-type: container
   modifier_class: vf-hero--easy
-  vf_hero_href: 'JavaScript:Void(0);'
+
   vf_hero_subheading: an example of some ancillary text
   vf_hero_text:
     - The Hentze group combines biochemical and systems–level approaches to investigate the connections between gene expression and cell metabolism, and their roles in human disease.
   vf_hero_link_text: Learn more.
   vf_hero_heading: About the Hentze group!
+  hero__initial_row: 384px;
```
khawkins98: Better as `vf_hero__initial_row`?

khawkins98 (approved):

sturobson:
I've also amended the classname too - `--vf-hero-grid__row--initial` - in case we ever need a second, or third, etc grid row custom property for `vf-hero`

PR #954, opened 2020-06-04
the one that only slightly changes tabs

sturobson:
I went around the houses, including a couple of 'redesigns' on this - and the result is 'not much changed'.


khawkins98:
Missed this one, just needs a changelog.

khawkins98 (approved):

PR #964, opened 2020-06-17
the one that makes the background image width the same as the content

sturobson:
We have an issue using `::before` in that it ignores the `vf-hero--inlay` variant and still displays the background image at full with.

This fixes that (for the short term). The CSS can probably be improved to reduce the amount of `:not` statements.


khawkins98 (approved):
So many (k)nots :D 

image

PR #965, opened 2020-06-17
update tab layout in fractal theme

sturobson:
I've changed the markup for the tabs in the fractal theme so that the "This component has variants:" text is now on it's own line.

This should save a bit of space for 'all the tabs'
Screenshot 2020-06-17 at 11 04 36

### DIFF
--- a/components/vf-card-container/vf-card-container.config.yml
+++ b/components/vf-card-container/vf-card-container.config.yml
@@ -21,22 +21,20 @@ context:
     - To promote molecular biology across Europe
     - To create a centre of excellence for Europe's leading young molecular biologists
   vf_cards:
-    - card_type: easy
-      card_title: One card
+    - card_title: One card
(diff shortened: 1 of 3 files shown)


components/vf-card-container/vf-card-container.config.yml
```diff
@@ -21,22 +21,20 @@ context:
     - To promote molecular biology across Europe
     - To create a centre of excellence for Europe's leading young molecular biologists
   vf_cards:
-    - card_type: easy
-      card_title: One card
+    - card_title: One card
```
khawkins98, line 24: did you mean to also update vf-card-container?

sturobson: yikes - not in this one

sturobson: I'll close this and start afresh

PR #975, opened 2020-06-23
the one where we create a new vf-profile component

sturobson:
This will replace `vf-summary--profile` which will be deprecated.

Notes:

- The content order is specific here - any changes will affect the layout.

Todo:

- I need to add CSS so if content is missing, or not required, there are still relative gaps between content items
- Investigate how this works with the yml/njk separation of data -- we might need a generic 'vf-container' where we can just dump stuff and it's nicely aligned.

### DIFF
--- a/components/vf-componenet-rollup/index.scss
+++ b/components/vf-componenet-rollup/index.scss
@@ -117,6 +117,8 @@ html, button {
 @import 'vf-activity-list/vf-activity-list.scss';
 @import 'vf-article-meta-information/vf-article-meta-information.scss';
 
+@import 'vf-profile/vf-profile.scss';
+
 @import 'vf-search/vf-search.scss';
 
 @import 'vf-social-links/vf-social-links.scss';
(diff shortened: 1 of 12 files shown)


khawkins98:
The profile image is still getting a bit stretched, anything we can do about that? Perhaps instead of an `<img scr=` we could do a CSS background image?

sturobson:
> The profile image is still getting a bit stretched.

I think this is because currently all avatars coming out of photo lab are 144px? We can make that the 'large size'? Or ask photo lab to redo them all larger? Which would help with retina blurring too.

> Perhaps instead of an <img scr= we could do a CSS background image?

I think we would still get the stretch-iness. 

What does using `object-fit: cover;` look like to you on the avatars?

khawkins98:
`object-fit: cover;` does seem to sort the squishing, I'd say +1 for it.

khawkins98 (approved):

PR #978, opened 2020-06-24
The one that makes figures without a class 'fit' better in vf-content

### LINKED ISSUE #959: add a margin reset for figure when inside of a vf-content block
looking at - https://wwwdev.embl.org/groups/chemical-biology/ - for a (currently) live example.

before:

Screenshot 2020-06-09 at 16 04 09

after: 

Screenshot 2020-06-09 at 16 04 55

sturobson:
This will close #959 

on [this page](https://wwwdev.embl.org/groups/chemical-biology/) there is a `figure` element with no `vf-` class that wraps a table element. This CSS removes the margin added by the browsers user stylesheet so that the content fits the page better.

### DIFF
--- a/components/vf-content/CHANGELOG.md
+++ b/components/vf-content/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change Log
 
+## 1.1.6
+
+* Adds CSS to remove margin from a figure inside vf-content that has no `vf-figure` class.
+
 ## 1.1.5
 
 * fixes an issue when content creators add the bold/strong tags to hedaing to make them bolder - when they shouldn't be.
--- a/components/vf-content/vf-content.scss
+++ b/components/vf-content/vf-content.scss
@@ -200,7 +200,7 @@
   }
 
   // WordPress is adding a figure around tables and we get default margins
-  figure:not([class*="vf-"]) {
+  figure:not([class*='vf-']) {
     margin: 0;
   }
 
--- a/components/vf-content/vf-content.scss
+++ b/components/vf-content/vf-content.scss
@@ -199,6 +199,11 @@
     margin-bottom: 32px;
   }
 
+  // WordPress is adding a figure around tables and we get default margins
+  figure:not([class*="vf-"]) {
+    margin: 0;
+  }
+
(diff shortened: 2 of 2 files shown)


khawkins98:
I wonder if this is better for some sort of `vf-normalise` or `vf-reset`? 

Two mixes of concerns I worry:

1. I install vf-figure but don't want you touch my `<figure class="some-other-thing">
2. I don't install vf-figure but do want `figure` elements to behave consistently

sturobson:
@kasprzyk-sz is looking at getting WP to remove the figure when including a table in `vf-content` as it's something WP does 'for you'. 

So this is a temporary code fix for `vf-content. 

@kasprzyk-sz will be linking to this PR from the one they have for figures and tables in `vf-wp` at some point.

khawkins98:
in this case perhaps better to add to vf-wp's `vf-local-overrides`?

sturobson:
regarding 

> vf-normalise or vf-reset?

It'd be nice to see if we can re-include a reset or normalise - but this would stop usages with VF1.x … because of it's reset/normalise. 

Now, several months in, I think a reset in `vf-core` would be a good thing - and where it breaks with VF1.x there needs to be something in that version of the codebase that fixes it. I think it's starting to hold us back a bit - and aids a little 'code bloat' too.

sturobson:
> in this case perhaps better to add to vf-wp's vf-local-overrides?

@kasprzyk-sz thoughts?

khawkins98:
I can't recall how we did the reset the first time ... i think _everyone_ got it? If it's a distinct component, at least EBI services could opt-out.

sturobson:
> I can't recall how we did the reset the first time ... i think everyone got it? If it's a distinct component, at least EBI services could opt-out.

I created a new issue #983 for this topic

khawkins98:
> I created a new issue #983 for this topic

multiply

sturobson:
A quick slack chat with @kasprzyk-sz and they're going to add the relevant CSS to `vf-wp` and then look to get WP to remove its need for adding a figure. 

We can close this now and close #959

PR #984, opened 2020-06-24
the one that hides some files from the documentation for a component.

sturobson:
this tidies up the 'files in the folder shown' list to only be:

- `*.scss`
- `*.js`
- `*.config.yml` - because of the template for the fractal docs this isn't included in the tabs, but good to keep here
- `*.njk` - because of the template for the fractal docs this isn't included in the tabs, but good to keep here


khawkins98:
So this would drop the:

- package.json: cool
- changelog.md: is there another way of seeing this?

sturobson:
> changelog.md: is there another way of seeing this?

perhaps we still need that for now.

khawkins98 (approved):

PR #985, opened 2020-06-24
the one that refactors vf-card so it works better in vf-11ty

### LINKED ISSUE #910: make vf-card a better linkable component
In the midst of trying to fix up better markup for the embl.org news grid container it would be to get an enhancement on the `vf-card` component to allow for a better, more accessible way to make the whole thing a link.

sturobson:
Lots of changes in this one as we shift to moving towards having 'content' and component decisions inside of a `.yml` file rather than inline in a `.njk` file.

- makes the link on the title, but also creates pseudo element so whole card is clickable
- only hover effect if card is link
- allows for inline `vf-card__link` inside of `vf-card__text` and make sure it's clickable over the card (using z-index)
- complete change in how the `.njk` variables are implemented

- gives `vf-card-container` and update for new `vf-card` component
- adds some default spacing to `vf-card-container

probably more things

this will close #910

### DIFF
--- a/components/vf-card-container/CHANGELOG.md
+++ b/components/vf-card-container/CHANGELOG.md
@@ -1,4 +1,9 @@
-## 1.0.0-alpha.2 
+## 2.0.0
+
+* makes better use of using .yml for data rather than having to inline it in a `vf-card`
+* better default spacing
+
+## 1.0.0-alpha.2
 
 * Tweaks layout centring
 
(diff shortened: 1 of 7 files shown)


khawkins98 (approved):

khawkins98:
Looks good. Expect this will also require a code change for the WP Gutenberg block (ping @kasprzyk-sz)

sturobson:
The major version bump should help avoid accidental code explosion

PR #986, opened 2020-06-24
Fix hiding of default variant

khawkins98:
image


khawkins98:
I've also uncapped the other variants as that's how fractal likes it, also the hyphens (`inline - easy - primary`) might freak out nunjucks

sturobson (approved):

PR #988, opened 2020-06-25
the one that brings relative paths back 

sturobson:
These updates will

- adds relative paths where appropriate
- allow us to site creators to import Sass modules without needing to rely on our gulp tasks
- doesn't break existing gulp task usage that uses include paths
- tidies up rollup.


khawkins98 (approved):
Looks like some nice optimisation work 👍

PR #989, opened 2020-06-25
The one that includes a 'starter kit' of Sass files

sturobson:
This PR creates a new component `vf-sass-starter` that can be used to get new projects off the ground.


sturobson:
should we rename this `vf-core-support`?

khawkins98 (approved):
Looks good. trimmed some boilerplate

khawkins98:
> should we rename this vf-core-support?

That makes me think of diagnostics. 🤷

PR #990, opened 2020-06-25
the one where we tentatively move to Dart Sass

sturobson:
Dart Sass is the "primary implementation of Sass" which means it will always supersede LibSass with features and bug fixes.

This is the first step in moving towards it inside of `vf-core` with only a few minor changes.

### DIFF
--- a/tools/gulp-tasks/vf-css.js
+++ b/tools/gulp-tasks/vf-css.js
@@ -220,7 +220,7 @@ module.exports = function(gulp, path, componentPath, componentDirectories, build
 
   // Sass Lint
   // For stylelint config rules see .stylelinrc
-  const vfScssLintPaths = [componentPath+'/**/embl-*.scss', componentPath+'/**/vf-*.scss', '!'+componentPath+'/**/index.scss', '!assets/**/*.scss', '!'+componentPath+'/vf-design-tokens/dist/**/*.scss', '!'+componentPath+'/vf-sass-config/**/*.scss'];
+  const vfScssLintPaths = [componentPath+'/**/embl-*.scss', componentPath+'/**/vf-*.scss', '!'+componentPath+'/**/index.scss', '!assets/**/*.scss', '!'+componentPath+'/vf-design-tokens/**/*.scss'];
   gulp.task('vf-lint:scss-soft-fail', function() {
     return gulp
       .src(vfScssLintPaths)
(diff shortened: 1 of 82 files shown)


sturobson:
It looks like we can keep using `gulp-sass` by adding the `sass.compiler` argument.

https://www.npmjs.com/package/gulp-sass

tools/gulp-tasks/vf-css.js
```diff
@@ -10,6 +10,7 @@ module.exports = function(gulp, path, componentPath, componentDirectories, build
 
   // Sass and CSS Stuff
   const sass = require('gulp-sass');
+  sass.compiler = require('dart-sass');
```
khawkins98, line 13: I guess we can't drop the node-sass dependency without changing all the things?

sturobson: Actually - we need that to tell `gulp-Sass` to use `dart-Sass` rather than `node-Sass`

sturobson: So I think node-Sass is a dependency of `gulp-Sass` that'll just 'stick around'. It's not in our package.json?

khawkins98: It's a shame as node-sass adds to the project boat and needs compilation. 

Looks like gulp-sass isn't interested in changing that: https://github.com/dlmanning/gulp-sass/issues/731

Stunt know if gulp-dart-sass would be a big change?

khawkins98: Looks like gulp-dart-sass is a form of gulp-sass, do may be a drop in change https://github.com/dlmanning/gulp-sass/compare/master...mattdsteele:master

sturobson: I went back to `gulp-Sass` and use of the `compiler` option because it's a more 'mature' npm package. 

I did have `gulp-dart-Sass` working at first

khawkins98:
Haven't forgotten about this. I'd still like tinker with `gulp-dart-sass` a bit, but in any case would be good to tag a maintenance release of vf-core (2.0.1) before switching the compiler.

sturobson:
Cool.

I think there's also two `gulp-dart-sass` packages too. Which doesn't help.

https://www.npmjs.com/search?q=gulp-dart-sass

Is the problem with `gulp-sass` that it simply makes the bundle a little larger on our laptops? 

I'm still in favour of sticking with a package that's 'mature'.

khawkins98:
This branch was behind `develop`, resolved a conflict.

khawkins98:
Investigated gulp-dart-sass int #1016, I think this PR's approach is just a technically performant (if a bit redundant) but tis the way of npm.

khawkins98 (approved):

khawkins98:
Note: After we merged I think we should now tag this as a 2.1 release as this a rather important change.

PR #991, opened 2020-06-25
the one that tidies the `vf-banner` up a little bit

sturobson:
This PR

- reduces the size of the alert banner by reducing padding
- makes the dismissible button accesible
- adds some more documentation
- makes the icon in the dismissible button a little smaller
- changes the `--grey-dark` to `--grey--darkest` so it has more contrast on the `--info` variant.

### DIFF
--- a/components/vf-banner/README.md
+++ b/components/vf-banner/README.md
@@ -6,10 +6,6 @@
 
 Informs visitors about important changes or persistent conditions. Use this component if you need to communicate to visitors in a prominent way. Banners are placed at the top of the page or section they apply to, and below the page or section header.
 
-### Exceptions
-
-The GDPR Banner, is the `--fixed` variant which needs to be 'sticky' to the bottom of the browser viewport
-
 ## Reflecting status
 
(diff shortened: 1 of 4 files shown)


components/vf-banner/README.md
```diff
@@ -2,6 +2,14 @@
 
 [![npm version](https://badge.fury.io/js/%40visual-framework%2Fvf-banner.svg)](https://badge.fury.io/js/%40visual-framework%2Fvf-banner)
 
+## Usage Guidelines
+
+Informs visitors about important changes or persistent conditions. Use this component if you need to communicate to visitors in a prominent way. Banners are placed at the top of the page or section they apply to, and below the page or section header.
+
+### Exceptions
+
+The GDPR Banner, is the `--fixed` variant which needs to be 'sticky' to the bottom of the browser viewport
```
khawkins98: Is this an exception to the rule, seems consistent with the above?

sturobson: I think you might be right here. I’ll amend.

khawkins98 (approved):
banner flag

PR #992, opened 2020-06-26
Bug/vf hero  inlay

sturobson:
There is an issue with the new `vf-hero` when we use the `vf-hero--inlay` modifier because we use margin shorthand rather than be explicit.

This PR simply changes 

```
margin: 0 auto;
```

to 

```
margin-left: auto;
margin-right: auto;
```

There's also a linter issue that got picked up -- I have no idea why.


khawkins98 (approved):
Approving with a changelog nag.

kasprzyk-sz (approved):

sturobson:
> Approving with a changelog nag.

Missed this

PR #995, opened 2020-06-26
the one that fixes the bug making tabs really large

sturobson:


khawkins98:
Approving + changelog nag

khawkins98 (approved):

PR #1001, opened 2020-06-30
the one that gets the Sass starter working 'as expected'

sturobson:
I missed a few things when publishing this package at first.

We now install sub packages as `dependencies` rather than `devDependencies`.

We also use relative paths for importing the other packages Sass–y CSS.

### DIFF
--- a/components/vf-sass-config/variables/vf-variables.scss
+++ b/components/vf-sass-config/variables/vf-variables.scss
@@ -1,19 +1,19 @@
 // Rollup of global Sass variables.
 /* stylelint-disable */
-@import 'vf-design-tokens/dist/sass/vf-border-radius.variables.scss';
-@import 'vf-design-tokens/dist/sass/vf-breakpoints.variables.scss';
-@import 'vf-design-tokens/dist/sass/maps/vf-breakpoints.map.scss';
-@import 'vf-design-tokens/dist/sass/maps/vf-colors.map.scss';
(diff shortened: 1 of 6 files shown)


components/vf-sass-config/variables/vf-variables.scss
```diff
@@ -1,19 +1,19 @@
...
-@import 'vf-design-tokens/dist/sass/maps/vf-breakpoints.map.scss';
-@import 'vf-design-tokens/dist/sass/maps/vf-colors.map.scss';
-@import 'vf-design-tokens/dist/sass/maps/vf-ui-colors.map.scss';
-@import 'vf-design-tokens/dist/sass/vf-font--monospace.scss';
-@import 'vf-design-tokens/dist/sass/vf-font--sans.scss';
-@import 'vf-design-tokens/dist/sass/vf-font-families.variables.scss';
-@import 'vf-design-tokens/dist/sass/vf-links.variables.scss';
-@import 'vf-design-tokens/dist/sass/maps/vf-spacing.map.scss';
-@import 'vf-design-tokens/dist/sass/maps/vf-zindex.map.scss';
+@import '../../vf-design-tokens/dist/sass/vf-border-radius.variables.scss';
```
khawkins98, line 3: Checked this with a local project using a customised `compontents/vf-design-tokens` and it still works as expected.

khawkins98 (approved):

PR #1005, opened 2020-07-07
The one that doesn't squish the default avatar.

sturobson:
Chrome liked to make `height: 100%` not look the same as it does in Firefox and Safari.

Switching `100%` to `auto` fixes this - and gives the same, correct, result in Firefox, Chrome, and Safari.


khawkins98 (approved):

khawkins98:
slinky

PR #1007, opened 2020-07-09
the one that adds contextual data option to vf-hero

sturobson:
I've added 

```
{%- if context %}
  {% set card_href = context.card_href %}
  {% set vf_hero__initial_row = context.vf_hero__initial_row %}

  {% set vf_hero_heading = context.vf_hero_heading %}
  {% set vf_hero_summary_title = context.vf_hero_summary_title %}
  {% set vf_hero_subheading = context.vf_hero_subheading %}
  {% set vf_hero_image = context.vf_hero_image %}

  {% set vf_hero_href = context.vf_hero_href %}
  {% set vf_hero_text = context.vf_hero_text %}
  {% set hero_text = context.hero_text %}

  {% set theme = context.theme %}
  {% set variant = context.variant %}

  {% set id = context.id %}
  {% set modifier_class = context.modifier_class %}
  {% set override_class = context.override_class %}
{% endif %}
```

to the top of `vf-hero.njk` so we can make use of `{% include %}` rather than `{% render %}` in 11ty

If there's no `context` - like there is not in `vf-core` then it is ignored. 

If there is a context that has been `{% set %}` then it will pass through to the components. 

Making cleaner `.njk` files and better separation of data and markup -- meaning better improvements will be easier to update in the future

Once merged - my plan is to edit all of the `vf-hero` that are used in the embl.org repo to use this and `.yml` files before moving onto the next components … and so on and so on and so on.

### DIFF
--- a/components/vf-hero/CHANGELOG.md
+++ b/components/vf-hero/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 1.4.5
+
+* adds an if statement for contextual data if applicable 
+
 # 1.4.4
 
 * fixes issue with `--intense` variant so parallax image is full width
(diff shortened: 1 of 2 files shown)


khawkins98 (approved):

khawkins98:
adding context

PR #1009, opened 2020-07-11
Fix: remove vf-design-tokens hard requirement from vf-sass-config

khawkins98:
The recent changes to the relative paths meant that a `node_modules/vf-design-tokens` would always be prefered over a local version (think: a site doesn't use the EMBL colours or typography sizing tokens)

This restores the ability to have local design tokens by not requiring a npm-installed vf-design-tokens. 

Not a bug exactly, more of regression in flexibility.

### DIFF
--- a/components/vf-sass-config/CHANGELOG.md
+++ b/components/vf-sass-config/CHANGELOG.md
@@ -1,8 +1,13 @@
 # Change Log
 
+## 1.1.3
+
+* Restores the ability to have local design tokens by not requiring a npm-installed vf-design-tokens.
+  * https://github.com/visual-framework/vf-core/pull/1009
+
 ## 1.1.2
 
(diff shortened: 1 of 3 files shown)


sturobson:
So, this won't run because it's missing things without the `vf-design-tokens` package.

I suggest we remove the instructions on how to install it - or link to `vf-sass-starter`

sturobson (approved):

sturobson:
So, this:

> So, this won't run because it's missing things without the vf-design-tokens package.
>
> I suggest we remove the instructions on how to install it - or link to vf-Sass-starter

isn't going to effect existing projects - but it might be good to add before merging in:

- remove the installation code snippet
- link to `vf-Sass-starter` as what they need to use to get up and running

sturobson (approved):

PR #1012, opened 2020-07-13
the one that refactors the vf-card and vf-card containers

sturobson:
`vf-card`: becomes a minor bump as it doesn't effect functionality of existing usage

- Adds nunjucks if statement for adding context if component used as sub-component

`vf-card-container`: becomes a major bump because the code has had significant changes. 

- makes `vf-section-header` work as a subcomponent.
- changes async for a for loop.
- adds contextual data changes if context is passed.


khawkins98 (approved):

khawkins98:
Looks good. I wonder if we should start making better use of alpha releases (v3.0.0-alpha.1).

PR #1017, opened 2020-07-20
the one that makes the profile a little more malleable.

sturobson:
- replaces `width: max-content;` with `width: 100%;` for the `vf-profile` component so that long text doesn't break layout.
- adds a little bit of margin to `vf-profile__text` if it follows `vf-profile__job-title`.


khawkins98 (approved):

khawkins98:
Oh, changelog nag.

PR #1018, opened 2020-07-20
the one that tidies up the `vf-inlay` layout as it moves towards deprecation.

sturobson:
this tidies up the markup required and relevant CSS for the `vf-inlay` layout component and it's available child elements.

You no longer need to include `vf-inlay__content` - I wonder if we should wrap this with our Sass-y deprecation code.

Because `vf-inlay__content` is essentially 'optional' and is in use - there is some CSS duplication of the grid code so that the `--full-width`, `--main`, and `--additional` variants work still. 

We've moved towards `grid-area:` for placement here too - instead of `grid-column: 2/4` etc.


khawkins98:
What'd this one get held up on, does it need more things? (other than a changelog 😄 )

sturobson:
iirc, I think it's only a changelog needed.

components/vf-inlay/vf-inlay.scss
```diff
@@ -11,100 +11,53 @@
...
     ;
-  }
-  /* stylelint-enable */
 
+    @media (min-width: $vf-breakpoint--lg) {
+      grid-template-columns:
+        [inlay--full-start inlay--main-start]
+          1fr
+        [inlay--main-end inlay--sidebar-start]
+          minmax(auto, var(--embl-grid-sidebar-width, 21.25rem)
```
sturobson: I've added a `--embl-grid-sidebar-width` custom property here. 

So when I give `embl-grid` layout the extra 80px we have now that `vf-body` is 1300px width. We can give this column a bit more space too

khawkins98 (approved):

components/vf-inlay/vf-inlay.scss
khawkins98: There's a yearning here to not have "embl" in the component, but probably not important as this component will slowly fade away.

sturobson: hmm, I didn't think of this 'cross signals'.

The only time the sidebar seems to be used is for the 'intro' or in place of the 'intro'

Perhaps we leave this for now and then I can come back to it.

Ideally - things like that should be created using `vf-grid` and the relevant span classes - to avoid creating a component just to affect the layout.

sturobson: leaving this in place for now. The CSS custom property comes as part of `vf-sass-config` so it shouldn't get missed.

PR #1020, opened 2020-07-21
Remove gulp-sass dependency

### LINKED ISSUE #1019: Sass includePaths ignored after switch to Dart Sass
In #990 we moved to Dart Sass, but it's not using the `includePaths`.

(note: this isn't a big deal for vf-core, but will be a big issue for things like vf-eleventy)

Docs suggest [includePaths should work with Dart](https://sass-lang.com/documentation/js-api#includepaths), so either gulp-sass isn't passing the option or something else has gone off. 

I have seen other issues about Dart Sass having issues with includePaths and gulp, investigation.

khawkins98:
Closes #1019 

Uses vanilla [dart-sass](https://www.npmjs.com/package/sass) over gulp-sass

Also optimises the sass build process to be cleaner.


package.json
```diff
@@ -39,6 +38,7 @@
     "npm": "^6.14.6",
     "rollup": "^2.22.1",
     "rollup-plugin-includepaths": "^0.2.3",
+    "sass": "^1.26.10",
```
sturobson, line 41: 👍

khawkins98:
This should now be good; does:

- Drops gulp-sass (and its node-sass+libsass dependencies) for a native implementation of dart-sass
- The above means a lighter npm load
- Makes steps towards a more portable `gulp vf-css` task
- Resolves a race condition when making package.variables.scss 
- Makes matching of .scss files more flexible
  - no longer specifies specific include paths
  - supports more flexible imports (most notably relative `../vf-table/vf-table.scss` and/or `vf-table/vf-table.scss`), relative paths are more opininated in the importer

Generally, it's a large rebuild of `gulp vf-css` that should be fully backwards compatible and more future-proof. 

Suggest after merging we release in a `vf-core@2.1.0-beta.1`

khawkins98:
This also drops node_modules from 658MB to 617MB. And from 43807 packages to 43024.

sturobson:
> This also drops node_modules from 658MB to 617MB. And from 43807 packages to 43024.

I’m sure visitors to the website will be pleased 🤪

sturobson:
So, we’re going all in on with letting the magic happen with file paths and not [this](https://github.com/visual-framework/vf-core/issues/1019#issuecomment-661696928)?

khawkins98:
Supplying specific paths only works well if you're exclusively using components from npm-installed vf-core. So it's good for EMBL.org's use case, but not things like covid19dataportal.org

sturobson (approved):
Good job!

PR #1029, opened 2020-07-30
the one fixes issue where SVG didn't appear in vf-links-list

sturobson:
I missed adding the `context` for the `svg` variable in `.yml` and `.njk` so it never showed up in the real world

This fixes that.


filipe-dias (approved):
Let there be SVG!

PR #1033, opened 2020-08-03
the one that finally deprecates the vf-summary--profile variant

sturobson:
- adds relevant code for deprecation
- points to `vf-profile` as needed
- hides from component previews
- updates version to `1.1.0` as I think it's more a 'minor' change because it's not totally removed, yet.


khawkins98 (approved):
Looks good.

PR #1034, opened 2020-08-03
The one that fixes up issues with vf-intro

sturobson:
- adds ability for `vf-badge` to be customised with it's variants
- allows for fallbacks so older versions will still work
- removes old way of doing `vf-badge` from demo code
- makes `vf-lede` work.

### DIFF
--- a/components/vf-intro/CHANGELOG.md
+++ b/components/vf-intro/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Change Log
 
+# 1.3.0
+
+* adds ability for `vf-badge` to be customised with it's variants
+* allows for fallbacks so older versions will still work
+* removes old way of doing `vf-badge` from demo code
+* makes `vf-lede` work.
+
 ## 1.2.0
 
 * wraps all content except heading in if statement
--- a/components/vf-intro/vf-intro.config.yml
+++ b/components/vf-intro/vf-intro.config.yml
@@ -5,9 +5,12 @@ status: live
 context:
   component-type: container
   vf_intro_heading: Cancer
-  intro_heading_href: 'JavaScript:Void(0);'
-  vf_intro_phase: beta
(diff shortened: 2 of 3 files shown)


khawkins98:
Are these breaking change and should this be 2.0.0 as:

- vf_lede_text is now vf_intro_lede
- vf-intro now needs to be set as a nested attribute

khawkins98:
Looking at this again, it seems `vf-intro` will work in the new or old way — could we also do the same for `vf_lede_text`?

Maybe also adding a note that the method is deprecated and will be removed in 2.0?

sturobson:
"vf-intro now needs to be set as a nested attribute" - it's been nested since [March 2019](https://github.com/visual-framework/vf-core/commit/26239c14328c3dca00331cc1f4d19570dba66158#diff-3e3e74f659bc7106cbdfd36e9bd0d3f8R8)?

khawkins98:
> "vf-intro now needs to be set as a nested attribute" - it's been nested since [March 2019](https://github.com/visual-framework/vf-core/commit/26239c14328c3dca00331cc1f4d19570dba66158#diff-3e3e74f659bc7106cbdfd36e9bd0d3f8R8)?

I mean that it now needs to be set as a further-nest Yaml: https://github.com/visual-framework/vf-core/pull/1034/files#diff-3e3e74f659bc7106cbdfd36e9bd0d3f8

But it looks like the old method will work so it's not breaking yet.

sturobson:
"vf_lede_text is now vf_intro_lede" We can do. the compatability stuff here too -- but it's weird because this should never have worked :sob: 

It's looking for `vf_lede_text` but in the render declaration(?) it's setting the context for `vf_intro_lede` … weird!

sturobson:
also -- the embl static pages make use of `vf_intro_lede` throughout too, rather than `vf_lede_text`

khawkins98 (approved):
I misunderstood the bug at play, this resolves it.

PR #1038, opened 2020-08-06
Feature: mo-mono repo vf core

khawkins98:
This makes vf-core be more of a true mono-repo; on the to-do list:

- [x] Move [vf-extensions](https://github.com/visual-framework/vf-extensions) to this project
- [x] Make the component generator a project (vf-component-generator)
- [x] Move sass build a tools/vf-sass-compilation project
- [x] Move fractal to a tools/vf-component-initialisation project
- Bonus: Explore a tools/vf-component-library project that may be 11ty based


khawkins98:
This is now working but I need to:

- [x] Figure out the bloat (`./node_modules` has doubled in size)
- [x] Note what, why

khawkins98:
I think the bloat is as I'm locally installing the packages and it copies everything in (no `.npmignore` at work)

That's probably why i was having issues on CI vs local.

khawkins98:
So what have I done here:

- the main vf-core directory is no longer intended to be an npm-installed project. I've renamed it to `@visual-framework/vf-core-repo` 

What we get:

- clearer visibility into dependencies of each part of the build process
- gets us closer to being able to just install the VF Sass build process, and not have to install Fractal, etc (this should now work, but needs testing)
- easier to ship updates, as vf-core is now lerna-managed and we no longer need a separate release process for vf-core and one for the components
- more flexibility, we should also be able to add an Eleventy documentation builder and use that with (or maybe even instead of) Fractal

Breaking things:

- should be nothing 🤞 
- for developing vf-core, you need to:
  - `yarn install` (as always)
  - `cd tools/vf-core`
  - `gulp vf-your-command-here`

Still to-do (I think these should wait until after we've shipped and be future enhancements)

- Break apart the gulp tasks from being only part of vf-core

I think we would ship this as VF 2.2.0 (2.2.0-alpha.1 at first, as it needs further testing in other vf-y projects)

Overall this should allow more modularity and flexibility for users of the VF, it means vf-core is no longer a wrapper around everything and allows it to become a truer build process.

sturobson:
this looks great!

sturobson (approved):

khawkins98:
strangelove rocket

There's a chance I've missed something important here, so I'll ease the roll-out of this:

- [x] test the standard CI deploy to github pages/gitlab after merging to develop
- [ ] test the CDN deploy to https://dev.assets.emblstatic.net/vf/develop/css/styles.css
- [x] publish the new components
- [x] test those new components as fit
- [x] publish a test of vf-core@v2.0.0-alpha.1
- [x] test that alpha as appropriate
- [x] once we know it works, update docs as needed in all the places (vf-core, vf-welcome)
- [x] repoint https://github.com/visual-framework/vf-extensions to the code on vf-core
- [ ] do a stable 2.2.0 release

khawkins98:
> this looks great!

And full marks for _the_ correct review comment 🤣

khawkins98:
☝️  And I think it's best to wait until Monday to start that rollout

khawkins98:
As the CDN assets (https://dev.assets.emblstatic.net/vf/develop/css/styles.css) are now in `/tools/vf-core/build/*` I'll need to get with Sandeep to update `bin/deploy-aws` or do something in the project code.

khawkins98:
The only thing (i've found) that requires updating in client projects is for vf-11ty:

```diff
- require('./node_modules/\@visual-framework/vf-core/tools/gulp-tasks/_gulp_rollup.js')(gulp, path, componentPath, componentDirectories, buildDestionation);
+ require('./node_modules/\@visual-framework/vf-core/gulp-tasks/_gulp_rollup.js')(gulp, path, componentPath, componentDirectories, buildDestionation);
```

I don't have a good idea for a workaround on this, but the number of projects to change is pretty limited.

PR #1053, opened 2020-08-17
The one that makes your computer fan a little quieter

sturobson:
I was finding that `gulp vf-dev` was often max'ing out my CPU and making my fans run flat out. 

It transpired that it's 'a thing' with `gulp-watch` which can be easily fixed by [adding an `interval` to your watch tasks](https://github.com/gulpjs/gulp/issues/634).

This PR does just that.

### DIFF
--- a/tools/vf-core/CHANGELOG.md
+++ b/tools/vf-core/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change Log
 
+## 1.0.1
+
+* adds an `interval` to each watch task to give the CPU a breather.
+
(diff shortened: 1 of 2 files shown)


khawkins98:
Nice one. I had never heard about the watch interval before!

khawkins98 (approved):

sturobson:
> Nice one. I had never heard about the watch interval before!

Likewise - I only found out because my 2019 16" Pro was about to take off. :sob:

PR #1054, opened 2020-08-17
Update deploy-aws

khawkins98:
In #1038 we changed the way the static assets are built, this should make things work but checking before I merge.

https://github.com/visual-framework/vf-core/pull/1038#issuecomment-671252858

> As the CDN assets (https://dev.assets.emblstatic.net/vf/develop/css/styles.css) are now in `/tools/vf-core/build/*` I'll need to get with Sandeep to update `bin/deploy-aws` or do something in the project code.

@sandykadam is what I've done here ok?


sturobson (approved):

khawkins98:
Success. All the things are now deploying.

PR #1057, opened 2020-08-19
🎨 makes all of the changelog files consistent

sturobson:
- 🔥 removes all headings of "change log"
- 🔥 removes any dates
- 🔥 removes any placeholder text
- ♻️  replaces `*` with `-` so lists can be generated
- ♻️ replaces `##` with `###` so it's semantic when used as part of new documentation


khawkins98 (approved):

khawkins98:
Just a footnote: this is going to result in a new version of all the see components. The next lerna release will be a beast.

khawkins98:
I'd suggest we do a component release before merging this — so that way we know any changed components are 0.0.1 minor bumps.

sturobson:
> Just a footnote: this is going to result in a new version of all the see components. The next lerna release will be a beast.

I did wonder about adding a patch release change for everything - but thought that it being only the changelog.md file would make it 'just fine' 😭 



> I'd suggest we do a component release before merging this — so that way we know any changed components are 0.0.1 minor bumps.

I'll work on adding to all the changelogs now with a patch bump up.

khawkins98:
I don't think we need to bother adding a changelog entry for "Updated the changelog" but maybe it's the "right" thing to do 🤷

sturobson:
Perhaps I've misunderstood. If I do:

> I'd suggest we do a component release before merging this

They'd all need a chagelog update?

khawkins98:
> They'd all need a chagelog update?

When you type `lerna changed` all the components will show up for a release. We could "cheat" by adding a tag before doing the release — as we don't need to ship the components to npm.

`git tag -a components.YYYYMMDD-01 -m 'Snapshot of components for lerna'`
`git push origin --tags`

sturobson:
> We could "cheat"

let's do that.

PR #1063, opened 2020-08-25
🎨 the one that replaces vf-grid-page with vf-body

sturobson:
🎨 adds `vf-body` component
🗑 deprecates `vf-grid-page`
♻️ reorders `vf-sass-config` variable import order
💄 adds some basic 'layout width' design tokens

### DIFF
--- a/components/_previews/_preview--nogrid.njk
+++ b/components/_previews/_preview--nogrid.njk
@@ -15,6 +15,7 @@
   {% if _config.project.environment.production %}
   <link rel="stylesheet" href="https://dev.assets.emblstatic.net/vf/develop/css/styles.css">
   {% endif %}
+  <style>body { margin: 0; }</style>
 </head>
 <body>
   {{ yield | safe }}
--- a/components/vf-body/vf-body.scss
+++ b/components/vf-body/vf-body.scss
@@ -12,10 +12,9 @@
 /* The Body
  * -------------- // --------------
  *
- * All content except for a few design patterns are tied into a grid with a maximum
+ * All content except for a few design patterns are tied into a layout with a maximum
  * width of 1300px. We center the `body` element to create a simple 3 column
- * grid with the central grid set to a maximum of 81.25em (1300px).
- *
+ * layout with the central grid set to a maximum of 81.25em (1300px).
  *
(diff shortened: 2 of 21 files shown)


components/_previews/_preview--nogrid.njk
```diff
@@ -15,6 +15,7 @@
   {% if _config.project.environment.production %}
   <link rel="stylesheet" href="https://dev.assets.emblstatic.net/vf/develop/css/styles.css">
   {% endif %}
+  <style>body { margin: 0; }</style>
```
khawkins98, line 18: Is this necessary, or was is it just a local development artefact?

components/vf-body/vf-body.scss
```diff
@@ -0,0 +1,32 @@
...
+/*!
+ * Component: #{map-get($componentInfo, 'name')}
+ * Version: #{map-get($componentInfo, 'version')}
+ * Location: #{map-get($componentInfo, 'location')}
+ */
+
+/* The Body
+ * -------------- // --------------
+ *
+ * All content except for a few design patterns are tied into a grid with a maximum
```
khawkins98: I think the "3 column grid" bit is out of date?

sturobson: ¯\_(ツ)_/¯ - it does technically create a 3 column grid. 

Happy to remove - if you think it would confuse

components/_previews/_preview--nogrid.njk
sturobson: yeah - the body element in fractal has margin - os it makes things look like they have more space than they would in a real web page - removing it - makes things like `vf-body` work 'as expected'.

components/vf-body/vf-body.scss
khawkins98: Maybe just sub "grid" for "layout"? Or maybe I'm overthinking it.

components/_previews/_preview--nogrid.njk
khawkins98: Gotcha. Fractal thingy 😄

khawkins98 (approved):

PR #1065, opened 2020-08-25
Component: vf-show-more

khawkins98:
Ported and improved from the one on EMBL.org. Bonus points to Stu for reminding me I had already started this!

To do:

- [x] Review what I've done so far
- [x] Add a way to specify number of items to show (or show per reveal)
- [x] Docs


khawkins98:
Ready for review.

components/vf-show-more/vf-show-more.scss
```diff
@@ -0,0 +1,83 @@
...
+ */
+
+@import 'vf-show-more.variables.scss';
+
+// You need to add this Sass file to ./components/vf-componenet-rollup/index.scss
+// @import 'vf-show-more/vf-show-more.scss';
+
+// Animation
+// Hiding CSS is added in JS
+.vf-show-more .vf-show-more--item {
```
sturobson: I'm not sure we need to increase the specificity of the CSS here? It could just be `.vf-show-more--item` etc.

Similarly with most of this CSS

components/vf-show-more/vf-show-more.njk
```diff
@@ -0,0 +1,48 @@
+{# Make sure any variables are listed inside the following if statement #}
+{% if context %}
+  {# {% set vf-show-more = context.vf-show-more %} #}
```
sturobson, line 3: Do you want me to go in and update this as needed?

components/vf-show-more/vf-show-more.njk
```diff
@@ -0,0 +1,48 @@
...
+{#
+{% if item.image == "blank" %}
+  <span><!-- no image --></span>
+  {% else %}
+  {{ item.image }}
+{% endif %}
+#}
+
+<section data-vf-js-show-more data-vf-js-show-more-pager-size="3" class="vf-show-more">
+  {% render '@vf-section-header--default', {
```
sturobson, line 7: this could be problematic when imported into the 11ty projects. I can change this so it 'follows suit' with other components.

Yeah - I need to document how to create a component now that 11ty has some 'special way'. Because of the `{% render %}`.

components/vf-show-more/vf-show-more.njk
```diff
@@ -0,0 +1,48 @@
...
+    "section_title": "Show more demo"
+  } %}
+
+  <div class="vf-grid vf-grid__col-3">
+    {% set items = [
```
sturobson, line 12: we could probably set this in the yml as I think `vf-card` is 'ready'

components/vf-show-more/vf-show-more.njk
```diff
@@ -0,0 +1,48 @@
...
+      { title: "foo", id: 4 },
+      { title: "foo", id: 5 },
+      { title: "foo", id: 6 },
+      { title: "foo", id: 7 },
+      { title: "foo", id: 8 },
+      { title: "foo", id: 9 }
+      ]
+    %}
+    {% for item in items %}
+      {% render '@vf-box--easy', {
```
sturobson, line 25: I meant `vf-box` not `vf-card` above.

Similarly here - need to check it's working fine with 11ty

components/vf-show-more/vf-show-more.njk
khawkins98: You've more experience here so I'll defer to you — but in practice I don't think we'll use this NJK template and rather graft the data attributes as needed.

sturobson: If/When it gets documented it should be easy … famous last words.

components/vf-show-more/vf-show-more.njk
khawkins98: As it's just demo data, is it worth extracting?

sturobson: It all depends with how well it plays in 11ty - I've had an issue with one components with this or similar. So I'll check.

Another thing for the checklist of new components -- does it work in 11ty 'as expected'

khawkins98: Ah, have actually been developing in the vf-component-library, and all is well there :)

image

components/vf-show-more/vf-show-more.scss
khawkins98: Now done

components/vf-show-more/vf-show-more.njk
sturobson: Although both 11ty and fractal use nunjucks -- 11ty is a little strict on the whitespace you give it. 

So there is a need for testing the component 'in the real world' because of that. Annoying that it is.

sturobson:
@khawkins98 I've done some more changes -- I had a merge conflict I think I fixed. Can you pull and see if works fine?

khawkins98:
Fixed a couple accidental things. Looking good from my end now.

khawkins98:
approve and/or merge when ready.

PR #1072, opened 2020-08-28
🎨 the one that refactors the `embl-grid` code and gives columns a little more space

sturobson:
- 🔥 removes unused `--alt` version (use a utility class for background colour).
- 🔥 removes weird margin that was used beforehand.
- 🎨 adds more space for content in initial (prime) column and sidebar column (if used).
- 💄 tidies up CSS by using css custom properties for variants.
- ✨ takes the hairline out of `--has-sidebar` and gives it it's own modifier class.

### DIFF
--- a/components/embl-grid/CHANGELOG.md
+++ b/components/embl-grid/CHANGELOG.md
@@ -1,3 +1,12 @@
+### 1.1.0
+
+- removes unused `--alt` version (use a utility class for background colour).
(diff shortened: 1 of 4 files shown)


khawkins98:
Good stuff. Should we call it a 2.0 as it removes/possibly breaking?

khawkins98 (approved):

PR #1073, opened 2020-08-28
bug: css nano aggressive grid optimizations

khawkins98:
what was `.vf-inlay__content--main {grid-area: inlay--main;`
was becoming `.vf-inlay__content--main{grid-area:a;`

related: https://twitter.com/jonsuh/status/849689928934010880?lang=en

ping @meladawy


sturobson (approved):
I think this hasn't been caught earlier because we haven't really used grid names for things until now. But `embl-grid` will also suffer this

PR #1076, opened 2020-09-01
the one that updates max-width in a few places and sets context for logos

sturobson:
It seems I missed a couple of changes to `max-width` on some components and their variants. This fixes that.

It also makes use of `{% if context %}` to allow the `vf-logo` that's rendered in `vf-global-header` to be 'whatever you want it to be' using `.yml`.

💄 updates max-width value
🎨 adds context options so vf-logo can be used properly
📦 updates changelogs

### DIFF
--- a/components/vf-global-header/vf-global-header.config.yml
+++ b/components/vf-global-header/vf-global-header.config.yml
@@ -6,4 +6,7 @@ status: live
 context:
   component-type: block
 
-  text: Visual Framework for Life Science websites
+  global_logo:
+    logo_href: /
+    image: ../../assets/vf-logo/assets/logo.svg
+    logo_text: Visual Framework for Life Science websites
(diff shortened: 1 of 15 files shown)


components/vf-global-header/vf-global-header.config.yml
```diff
@@ -6,4 +6,7 @@ status: live
 context:
   component-type: block
 
-  text: Visual Framework for Life Science websites
+  global_logo:
```
khawkins98, line 9: This one's probably another breaking change if you're already using `text`?

sturobson: I guess - using the `{{text}}` in the `.njk` felt like a 'hack' because you have the logo that you add text to.

So this will fix that.

Still needs work to make this container component fully customisable. 

Happy to update to a breaking change if you give this a 👍

khawkins98: sounds good 👍

sturobson:
@khawkins98 I think I've just pushed some of these to npm -- waiting for the all clear to merge into `develop` ** cough **

khawkins98 (approved):
Forgot about this one.

PR #1079, opened 2020-09-02
chore: vf-card conditional title and alt example

khawkins98:
- adds alt text to examples
- conditional display of `card_title`

### DIFF
--- a/components/vf-card/CHANGELOG.md
+++ b/components/vf-card/CHANGELOG.md
@@ -1,3 +1,8 @@
+### 2.1.1
+
+- adds alt text to examples
+- conditional display of `card_title`
+
 ### 2.1.0
 
 - makes theme variant naming and decisions consistent.
--- a/components/vf-card/vf-card.config.yml
+++ b/components/vf-card/vf-card.config.yml
@@ -23,6 +23,7 @@ variants:
       variant: very-easy
       card_image: ../../assets/vf-card/assets/vf-card-example.png
       card_text: Lorem ipsum dolor sit amet, consectetur adipisicing elit. Sapiente harum, omnis provident saepe aut eius aliquam sequi fugit incidunt reiciendis, mollitia quos?
+      card_image__alt: Image alt text
   - name: easy
     label: Easy
     hidden: false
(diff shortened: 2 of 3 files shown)


sturobson:
Have you got the latest from `develop` because at the top of the `vf-card.mjk` I get:

```
{%- if context %}
  {% set card_href = context.card_href %}

  {% set card_title = context.card_title %}
  {% set card_subtitle = context.card_subtitle %}
  {% set card_text = context.card_text %}
  {% set card_image = context.card_image %}
  {% set card_image__alt = context.card_image__alt %}

  {% set theme = context.theme %}
  {% set variant = context.variant %}

  {% set id = context.id %}
  {% set modifiers = context.modifiers %}
  {% set override_class = context.override_class %}
{% endif %}
```

Which looks different to what you're editing here.

khawkins98:
Think that's the same, what's different? https://github.com/visual-framework/vf-core/blob/6958d05cbe3541acc35c3d0a83fbee9d8e8020cf/components/vf-card/vf-card.njk

sturobson:
looks like there's less here -- https://github.com/visual-framework/vf-core/pull/1079/files#diff-5d4072670831661cea61311fba63cfd5 -- is that 'because github'?

khawkins98:
Ah, gitlab has just collpased the "same" bits. 
image

I added white space trim to some tags

sturobson:
I'll take this branch and add somethings that got picked up (theming) so we can get it all in one merge - if that's cool?

khawkins98:
👍

sturobson:
can you explain more in what this means?

> conditional display of card_title

I read it as if there's no title, don't show anything - but the code changes the `h3` to a `span`‽

khawkins98:
At first, I had the whole thing commented out but then realised the href needs to be wrapped in `.vf-card__title` to do its thing.

This is the best way I thought of to change the fewest things while not putting out an empty h3.

sturobson:
How does a screenreader react if there's no title to the link? I'm assuming it will read out the url.

sturobson:
I actually don't have anything to change by the looks of it - the issue I saw with link underlines on wwwdev must be inheriting it from some other CSS somewhere - so will investigate there instead.

sturobson (approved):

khawkins98:
> I actually don't have anything to change by the looks of it - the issue I saw with link underlines on wwwdev must be inheriting it from some other CSS somewhere - so will investigate there instead.

Yeah that was because the markup was so old. the whole vf-card was an `a` element

sturobson:
> Yeah that was because the markup was so old. the whole vf-card was an a element

Ah, yes. That'll explain it - I'm guessing that the updates to `static-html-pages` will fix this too.

Issue #1077, opened 2020-09-01
Colour of visited links in Normal Primary box


spivass81:
The visited links in the Normal Primary box turn purple, which is kind of hard to read. 

<img width="1063" alt="Screenshot 2020-09-01 at 16 52 25" src="https://user-images.githubusercontent.com/50863581/91870451-43254200-ec77-11ea-8446-e19a5f950d20.png">

Ideally, the color should remain white.
Haven't tested this on the other type of boxes, @sturobson leave this up to you to decide whether the fix should be applied to all of them.

PR #1084, opened 2020-09-03
💄 the one that updates :visited colour for links in text

sturobson:
If you added a link inside of `vf-box__text` it would inherit the browser styles if the link had been `:visited`.

This PR makes the `:visited` class stay the same colour.

💄 adds `:visited` to the list of pseudo-classes that inherit the `currentColor` of the node.
📦 updates changelog


khawkins98 (approved):

sturobson:
This should close #1077

PR #1087, opened 2020-09-03
🐛 The one that fixes bug with image height in Safari

sturobson:
🐛 fixes bug where Safari would give you a really tall image
💄 adds `object-position: center;` to always centre the image when viewport at various sizes.
📦 updates the changelog.

### DIFF
--- a/components/vf-card/vf-card.scss
+++ b/components/vf-card/vf-card.scss
@@ -17,16 +17,17 @@
 
   background-color: var(--card-background-color);
   display: grid;
-  grid-template-rows: 216px 1fr;
+  grid-template-rows: 288px 1fr;
   position: relative;
 }
 
 .vf-card__image {
   grid-column: 1 / -1;
-  grid-row: 1 / span 3;
+  grid-row: 1;
   height: 100%;
   max-width: 100%;
   object-fit: cover;
+  object-position: center;
   width: 100%;
 }
 
--- a/components/vf-card/CHANGELOG.md
+++ b/components/vf-card/CHANGELOG.md
@@ -1,3 +1,8 @@
+### 2.1.2
+
+- fixes issue with `vf-card__image` height in safari.
+- adds `object-position: center` for image.
+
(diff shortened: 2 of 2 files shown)


components/vf-card/vf-card.scss
```diff
@@ -17,16 +17,17 @@
...
 }
 
 .vf-card__image {
   grid-column: 1 / -1;
-  grid-row: 1 / span 3;
+  grid-row: 1;
   height: 100%;
   max-width: 100%;
   object-fit: cover;
+  object-position: center;
```
khawkins98, line 30: When we have Safari/IE/etc specific fixes, I think it would be good if we had a standardised way to note the fix; something like

```css
  object-fit: cover;
  object-position: center; // safari centring fix: https://github.com/visual-framework/vf-core/pull/1087
  width: 100%;
}
```

sturobson: `object-position: center` will work for all browsers [that support it](https://caniuse.com/#feat=mdn-css_properties_object-position).

The issue with Safari was having `grid-column: 1 / span 3` on the `vf-card__image` (and I cannot see a reason why it's there) because it was making the grid row for the image 3 x 216px tall. It didn't do this in other browsers (I'm guessing Gecko and Chromium work out the image isn't that large and makes it 'as big as it actually is'.

So I've bumped up the height of the first row to make the `vf-card__image` as tall as it was in other browsers.

So this works 'as expected' across all browsers except IE11 - which just defaults to letting the content flow.

sturobson: but yes, specific browser bug fixes should get a nice little comment so we know why it's there.

sturobson: in this case it was more 'bad CSS' than 'bad browser' :D

khawkins98: Sounds good :)

khawkins98 (approved):

PR #1088, opened 2020-09-03
🎨 The one that adds the `loading` attribute to component images

sturobson:
🎨 adds `loading="lazy"` to all `img` elements in components, except for `vf-logo` which gets `loading="eager"`.
📦 updates all the changelogs


khawkins98 (approved):
Noice.

Issue #1047, opened 2020-08-11
turn `vf-u-fullbleed` it's own container


sturobson:
- keep `vf-u-fullbleed`
- create a new container that is full width with a centred column - `vf-fullbleed`
- allow for theming with primary, secondary, tertiary, quaternary, quinary.
- appropriate 'container defined' text colour

sturobson:
With more implementations on `vf-u-fullbleed` I don’t think an actual ‘container’ works here as the cod base is moving towards an ‘additive’ approach with layout - having an ‘fullbleed’ container contradicts it. And _this works_.

Closing

PR #1089, opened 2020-09-04
✨ the one that adds a new layout component for vertical rhythm / flow

sturobson:


sturobson:
@khawkins98 this should be good enough to merge in now I reckon.

khawkins98 (approved):
Looks good. One question: should this be a utility class/component instead? `vf-u-stack`?

sturobson:
> Looks good. One question: should this be a utility class/component instead? `vf-u-stack`?

I think it meets the criteria for layout more. With a utility class being something we could use to override the stacking value if needs be (I'll work that in here now).

There's also [this issue](https://github.com/visual-framework/vf-core/issues/1047) where I think we should look to update `vf-u-fullbleed` to a fully blown layout component.

This also goes back to me thinking … perhaps we should prefix classes with things like `l` for layout `vf-l-stack` `vf-l-grid` etc.

But we're probably too far down the road for that complete rewrite :sob:

khawkins98:
Sounds reasonable enough to me.

PR #1093, opened 2020-09-07
Documentation: begin to make more pattern/boilerplate guidance

khawkins98:
Todo:

- [x] search page
- [x] blank content page
- [x] containers with sidebares
- [x] sample conversion of a service or something application-y (like http://www.ebi.ac.uk/interpro/ or https://www.ebi.ac.uk/services)
- [x] example grafting VF 2.0 into existing non-branded service like Uniprot


tools/vf-component-library/src/site/patterns/index.njk
```diff
@@ -2,11 +2,10 @@
 is_index: true
 title: Patterns and boilerplates
 subtitle: More than components, this is how you should do things.
+intro: Where components offer technical implementations, patterns and boilerplates show how things should come together to facilitate tasks and goals.
```
sturobson, line 5: I'm leaning towards 'content' being in the front matter being a bit of an anti pattern when we have `.yml` available for it all - probably not the case here -- but in some places it has made using `{% include component %}` 'icky' because you're having to `{% set content as thing %}` in the markup more. 

Anyways - just a drive–by comment for the Monday afternoon

khawkins98:
This should be good to go now. There's more work to be done but it's:

1. Better than what we have now
2. Will be of use for the UX clinic later today

sturobson (approved):

PR #1095, opened 2020-09-08
💄 the one that updates all the things for the vf-box component

sturobson:
- 🎨 removed bottom margin from component. You should now make use of `vf-stack` in a parent element to create vertical spacing between multiple components.
- 🎨 made the `vf-box__title` the link rather than the whole component, using ::after pseudo element so the whole of `vf-box` is clickable.
- 📝 added some more examples (easy primay with link, normal primary with link, normal primary with link with text link).
- 🔥 removed 'default' component `.yml` data.
- 🔥 removed `box` from `.yml` and `.njk` for setting context as it was an additional step.
- 🔧 adds `variant` and `theme` options to replace `vf-box__modifier` (`vf-box__modifier` will be removed in `v3,0`).
- 🔥 removes previously deprecated variants `vf-box--inlay` and `vf-box--factoid`.


components/vf-box/vf-box.scss
```diff
@@ -57,7 +74,9 @@
   .vf-content & a,
   a {
     color: currentColor;
+    position: relative;
     text-decoration: underline;
+    z-index: 1984;
```
khawkins98: Should use the design token z-indexes for these?

khawkins98 (approved):

components/vf-box/vf-box.scss
sturobson: I was thinking that when I quickly type it. Will update

PR #1098, opened 2020-09-09
bug: vf-box specificity

khawkins98:
margin reset on `.vf-box :last-child` needs more specificity as `.vf-box > :last-child` affects the last child of every nested element, not just the last child of `<div class="vf-box">`


sturobson (approved):
good catch!

sturobson:
I mean - we could say that the layout in `vf-box` should be defined by adding `vf-stack` too - but that'll take some work -- looking at how many boxes there are on embl.org :sob:

Issue #1105, opened 2020-09-09
Further enhance boilerplates


khawkins98:
In #1093 we added some new boilerplates, but of course, we need to do more. 

- [ ] Error pages (404, 403, 500) Example 404: https://wwwdev.embl.org/kdjafljdls
- [x] Further examples to the search page
    - https://visual-framework.github.io/vf-core/patterns/search-results/
    - [x] summary list results
    - [x] table results
    - [x] card results
    - [ ] mobile versions of the above

khawkins98:
A conceptual example of search results for courses from @nikiforosk 

![course snippets - v2 - 08Sep20](https://user-images.githubusercontent.com/928100/92706023-78efb800-f354-11ea-8836-54bfe62db2d7.jpg)

gabsie:
HI, I also am attaching some explorations from the Covid-19 portal, but wouldn't stand by any of them too much - they have been done quickly and trying to put out a fire.. I can probably spend more time thinking about this and also about some search result scenarios relevant to HCA.

![Nucleotide Sequences - search results in tables](https://user-images.githubusercontent.com/19713125/92739499-dc89dd80-f374-11ea-90a3-03cab4bb2dba.png)
![Related Resources - filtering and tagging](https://user-images.githubusercontent.com/19713125/92739509-e01d6480-f374-11ea-9841-289e56062d86.png)

nikiforosk:
Below are links to more complete examples which show all six type of courses and the kinds of summaries that we have been considering for each search result:
- [First ~200 characters of the overview without the keyword highlighted
](https://invis.io/2NYEWTQEURS#/431380347_Course_Snippets_-_V2_-_First_200_Chars_-_11Sep20)
- [Fragments with keywords highlighted](https://invis.io/2NYEWTQEURS#/431380348_Course_Snippets_-_V2_-_Highlighted_Keywords_-_11Sep20)

It looks like we'll go with the first option because the context surrounding the keyword in the fragments doesn't always seem to be very informative.
 
@khawkins98 As I mentioned yesterday, I'd appreciate any suggestions on how we can make certain types of courses (e.g. virtual course) stand out using existing elements of the visual framework.

khawkins98:
The closest existing example  is https://wwwdev.embl.org/events page which has many similar semantic fields

![image](https://user-images.githubusercontent.com/928100/93081067-f423d780-f68e-11ea-9325-0436125b1b17.png)

[...]

PR #1112, opened 2020-09-14
Update search-results.njk

khawkins98:
For #1105, this makes the search example more helpful.

- adds note: facet layout is yet to be made mobile friendly

### DIFF
--- a/components/vf-summary/vf-summary--event.scss
+++ b/components/vf-summary/vf-summary--event.scss
@@ -94,7 +94,7 @@
 .vf-summary-theme--secondary {
   --vf-summary-theme-color--foreground: #{set-ui-color(vf-ui-color--black)};
   --vf-summary-theme-color--border: #{set-color(vf-color--green)};
-  --vf-summary-theme-color--background: #{set-color(vf-color--green)}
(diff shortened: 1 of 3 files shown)


khawkins98:
Working on a fairly generic solution that is suitable to a few types of result: 

image

Though I fear it may mean yet another summary variant? `vf-summary__search-result`?

sturobson:
These are very much `vf-summary`–like. 

As this is having a bit of a mass overhaul to simplify the design(s), HTML and CSS - I'd worry that:

a) there's a little duplication of efforts
b) these patterns might be 'gazumped' with the later update.

However, if there's more `b` than `a` and it's possible to quickly update the usage of this across all products - go for it :)

khawkins98:
> However, if there's more b than a and it's possible to quickly update the usage of this across all products - go for it :)

I'll try to explore a bit more tomorrow, but frankly, I'm not sure if I'll get to it before my holiday break. 

I'd say let's go ahead and get this one in as it's better than what is there now and also fixes the component search for the component library: https://github.com/visual-framework/vf-core/pull/1112/files#diff-516bf46638d140b95582b50dcba78d8dL10

sturobson (approved):

PR #1125, opened 2020-09-17
💄 the one that makes `vf-divider` a little more malleable.

sturobson:
💄 updates `vf-divider` so it can allow for use alongside `vf-u-fullbleed`.
♻️ refactors `vf-u-fullbleed` so it can work 'as expected' but allow for 'context customisation' as needed from associated components. 
♻️ refactors the `_divider` Sass `@mixin` to allow customisation for inline (left and right) and block end (bottom) margins.
📝 updates `vf-divider` documentation to explain the possibilities `v2.0.0` allows.
📦 updates all the changelogs required.

### DIFF
--- a/components/_previews/_preview--body.njk
+++ b/components/_previews/_preview--body.njk
@@ -1,0 +1,29 @@
+<!DOCTYPE html>
+<html lang="en" class="vf-no-js">
+<head>
+  {% render '@vf-no-js' %}
+  <title>{{ _target.title }} | {{ _config.project.title }}</title>
+
+  <meta charset="utf-8">
+  <meta content="width=device-width, initial-scale=1, minimum-scale=1" name="viewport">
+
+  {% render '@vf-favicon' %}
+
+  {% if _config.project.environment.local %}
+  <link rel="stylesheet" href="{{ '/css/styles.css' | path }}">
+  {% endif %}
+  {% if _config.project.environment.production %}
+  <link rel="stylesheet" href="https://dev.assets.emblstatic.net/vf/develop/css/styles.css">
+  {% endif %}
+  <style>
+  body {
+    padding-top: 1rem !important;
+  }
+  </style>
+</head>
+<body class="vf-body">
+  {{ yield | safe }}
+
+  <script src="{{ '/scripts/scripts.js' | path }}"></script>
+</body>
+</html>
--- a/components/vf-sass-config/CHANGELOG.md
+++ b/components/vf-sass-config/CHANGELOG.md
@@ -1,3 +1,9 @@
+### 2.0.0
+
+- removes the inline margin from the component.
+- adds the block end margin an creates a Sass variable for `$margin--block-end`.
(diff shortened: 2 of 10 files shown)


components/_previews/_preview--body.njk
```diff
@@ -0,0 +1,29 @@
+<!DOCTYPE html>
```
khawkins98, line 1: Do we need to add this preview to Fractal?  Tried using the default and it still seemed to work "as expected"?

image

components/vf-sass-config/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+### 2.0.0
```
khawkins98, line 1: Doesn't seem like a breaking change, but maybe I'm overlooking something ...

components/vf-u-fullbleed/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+### 1.2.0
+
+- adds a CSS custom property for when a component with `vf-u-fullbleed` needs inline margins.
```
khawkins98, line 3: Not really sure what this means? Is it easy to add an example of when/where this might be used?

sturobson: I guess it's more 'inset' spacing -- so you can have a full bleed divider but not fill all the way to the edges of the viewport -- and edge case use for sure. Will rethink how this can be demonstrated and worded.

components/vf-sass-config/CHANGELOG.md
sturobson: I guess it's not as we're not really breaking functionality here -- just adding to it - can happily be pulled back to `1.5.0`

components/_previews/_preview--body.njk
sturobson: the default one adds layout that only will ever be seen in fractal–land - I think it was to show the difference with the full bleed - I'll have to take another look

components/vf-u-fullbleed/CHANGELOG.md
khawkins98: Ah, so something like `<hr class="vf-divider vf-u-fullbleed">` but not going 100%?

components/_previews/_preview--body.njk
sturobson: having a look at this -- if we use the defaultt, when we open it 'with layout open' - you can't see the divider because of the lack of top padding.

components/vf-sass-config/CHANGELOG.md
sturobson: taking a look at this - I think because the HTML or Nunjucks has changed dramatically - it warrants the major / breaking version change.

components/vf-u-fullbleed/CHANGELOG.md
sturobson: yeah - you can make the left / right margin smaller or larger -- if you ever need to.

khawkins98 (approved):

PR #1130, opened 2020-09-28
the one that updates the layout components for eleventy and adds the cluster layout

sturobson:


sturobson:
I've added 'documentation' for `vf-cluster`.

I'd like to get this integrated - so that I can test out how it can be used with `vf-eleventy` so I can create better docs.

components/_previews/_preview--nogrid.njk
```diff
@@ -17,7 +17,7 @@
   {% endif %}
   <style>body { margin: 0; }</style>
 </head>
-<body>
+<body class="vf-body">
```
khawkins98, line 20: Adding vf-body here might break the component preview in some cases. But only for the vf-core fractal previews.

khawkins98 (approved):

components/_previews/_preview--nogrid.njk
sturobson: `--no-grid` is used for 13 components.

1 is deprecated, leaving 12. There's `vf-grid` and `embl-grid` which takes it down to 10.

Some of these have a `<div class="vf-body">` in the code - which shouldn't be there now.

Some of these have a `max-width` set which is on the 'remove list'.

As it's only the component view - I think we can merge this in - and then there's a task to tidy up the strays. 

Thoughts?

khawkins98: I'm not terribly bothered by it as  the vf-frctl theme is on the way to deprecation (hopefully) or at least some sort of refactoring with the view on making it more functional and less of a full component library.

sturobson: 👍 

although for local development, it'll still be used when inside `vf-core`? 

it'd be good to clear up any components that are mixing themselves up in having other components hard coded to show the example though.

resolving and merging :)

PR #1134, opened 2020-09-29
bug: design tokens demo not shown

khawkins98:
As vf-core now lives in subdirectory the paths were broken in `vf-design-tokens.config.js`, this means we weren't getting the nice demo on the component page (both for Fractal and 11ty)

 This fixes it by referencing a project's `vfConfig`, which should make things more stable in general.


sturobson:
I've just checked and this breaks locally when developing `gulp vf-dev` - I assume this fixes both local 'fractal based' development and the live site?

sturobson (approved):

PR #1139, opened 2020-09-30
Enhancement: vf section header ids

khawkins98:
Allows IDs to be added to section headers for anchor links

### DIFF
--- a/tools/vf-component-library/src/site/troubleshooting/index.njk
+++ b/tools/vf-component-library/src/site/troubleshooting/index.njk
@@ -5,30 +5,35 @@ date: 2019-04-09 12:24:50
 tags:
   - troubleshooting
 layout: layouts/base.njk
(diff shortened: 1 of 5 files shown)


tools/vf-component-library/src/site/troubleshooting/index.njk
```diff
@@ -5,30 +5,35 @@ date: 2019-04-09 12:24:50
...
+    ],
+    "text": "beta",
+    "theme": "primary",
+    "badge_href": "JavaScript:Void(0);"
+  }
 } %}
 
-
 <section class="vf-content | embl-grid embl-grid--has-centered-content">
+{% render '@vf-section-header', {
```
sturobson, line 30: it would be nice to eventually get this using `.yml` instead of relying on `{% render … %}`

sturobson (approved):

PR #1144, opened 2020-10-02
Docs: cleanup and boilerplates

khawkins98:
an ongoing work in progress of some cleanup and further enhancement to the boilerplates and docs.


tools/vf-component-library/src/site/404.njk
```diff
@@ -11,7 +11,7 @@ permalink: 404.html
 <div><!-- empty --></div>
 <div>
   <h1 class="vf-intro__heading">{{ title }}</h1>
```
sturobson, line 13: shouldn't the elements of a `vf-intro` component be inside a parent that has the `vf-intro` class on it? ;)

khawkins98: Yeah, this bit of code is quite old. I'm going to attempt a refactor of the landing page to better address some of the use cases we've seen .. less pontificating about the VF and making it more "How do I get the code and use it"

khawkins98: Just realised this isn't the homepage! 😬 

There is a vf-intro parent element (code is collapsed in the diff). But this code is also still quite old and needs a "modernisation".

sturobson: ugh, I think this is the third time you've reminded me of code being collapsed in a diff 😭

sturobson (approved):

khawkins98:
I think this has reached enough maturity that it can be committed. As always, the documentation and component interface will never be "done". But I think this is much better and it would be good to have it for Tuesday's clinic. 

Main differences:

- Updates the "content funnel" for the homepage to talk less about the merits of the VF and get more quickly into how and where to get the code
- Minor content tweaks/fixes
- Adds a variable to the 11ty site for the latest VF version so we can easily template things like 

```html
<!-- CSS only -->
<link rel="stylesheet" href="https://assets.emblstatic.net/vf/v2.2.9/css/styles.css">
<!-- JS -->
<script src="https://assets.emblstatic.net/vf/v2.2.9/scripts/scripts.js"></script>
```

Homepage screengrab:
image

sturobson:
Looking good @khawkins98

PR #1147, opened 2020-10-05
enhancement: add support for the autofocus attribute

khawkins98:
For things like embl.org/search it's nice if a user doesn't have to click on the search box. 

Adds support for both vf-search and vf-search-clientside. 

https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/autofocus

### DIFF
--- a/components/vf-search-client-side/CHANGELOG.md
+++ b/components/vf-search-client-side/CHANGELOG.md
@@ -1,9 +1,13 @@
 # Change Log
 
-# 1.0.0-alpha.2
+## 1.0.0-alpha.3
+
+* adds support for `autofocus` on search input
+
+## 1.0.0-alpha.2
 
 * Improve documentation
 
-# 1.0.0-alpha.1
+## 1.0.0-alpha.1
 
 * Initial prototype
--- a/components/vf-search-client-side/README.md
+++ b/components/vf-search-client-side/README.md
@@ -66,6 +66,7 @@ And you should build that search index after updating your html pages, a la:
 
 - You can pass a query to the search page with `?search_query=myQuery`
 - The search will live update as the user enters text
+- You can enable `autofocus` on the search element, but should only do so if most users intend to search on the page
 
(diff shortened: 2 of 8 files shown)


sturobson:
👏

could-should

🤣


As cool as this is, how does this effect the tabbing order for screen readers etc?

Should we ask UX folks on their thoughts and understandings about it too?

khawkins98:
I think [this](https://stackoverflow.com/questions/2180645/is-automatically-assigning-focus-bad-for-accessibility) sums it up well:

> The short answer is no it doesn't make things inaccessible but it can make it confusing. The longer answer follows. Will your users know that there going to a page with a form, and does there need to be any descriptive text you should read before filling out the form? I'm a screen reader user and it can be annoying having focused put in random fields. It's clear why your focus winds up in the Google search box so that doesn't bother me. If my focus were automatically placed in the answer edit field every time I viewed a question on Stackoverflow I would be annoyed since I'd have to force my screen reader to navigate away from the form field and to the top of the page.

It's a standard HTML form element which like most things need to be used in the right context.

sturobson:
Cool, so it announces itself via VoiceOver? 

I've tried pulling this down locally - but the search isn't automatically focusing on the input for me in Safari or Firefox - with VoiceOver on it does nothing in Safari and in Firefox it automatically 'tabs' through input > button > loading text

Does it need to be working 'with real content' to work as expected?

sturobson (approved):

khawkins98:
Aside: had the thought that this should not default on. setting to false.

khawkins98:
> Does it need to be working 'with real content' to work as expected?

I think having it defaulting to autofocus may have been confusing matters, especially with vf-search where there were two search forms shown. Clearly two autofocus items is a bad idea. :)

khawkins98:
Or even one autofocs is a bad idea, if most users don't intend to type on that page.  ... adding docs.

PR #1155, opened 2020-10-13
the one that takes the hero component back to 1999

sturobson:
reduced the padding for `--intense` and also reduced the height of every other `vf-hero`.

All based on 16px/1em. 

This is still 'one size fits all' … making use of what will be the `xxxl` spacing unit of `48px`. Which would be 'the most' space taken up by imagery around the text content of `vf-hero`.

bonus from previous WIP on the `vf-hero` - you get the context data for use in 11ty.


khawkins98:
Squeeze that space compactor

khawkins98 (approved):

PR #1156, opened 2020-10-13
the one that lets us have more space, what a concept. 

sturobson:
"t-shirt" sizing will at some point become unmanageable. 

In fact, introducing a spacing token of `48px` or `3rem` gives us an extra extra extra large spacing token.

This PR wholesale changes all tokens in use to make use of the value `100` meaning `.25rem`.

```
  - name: vf-spacing--0
    value: 0
    
  - name: vf-spacing--50
    value: .125rem
    
  - name: vf-spacing--100
    value: .25rem
    
  - name: vf-spacing--200
    value: .5rem
    
  - name: vf-spacing--400
    value: 1rem
    
  - name: vf-spacing--500
    value: 1.25rem
    
  - name: vf-spacing--600
    value: 1.5rem
    
  - name: vf-spacing--800
    value: 2rem
    
  - name: vf-spacing--1200
    value: 3rem
```

This allows for more additions as the system gets bigger.

### DIFF
--- a/components/vf-box/CHANGELOG.md
+++ b/components/vf-box/CHANGELOG.md
@@ -1,6 +1,7 @@
 ### 2.1.0
 
 - updates spacing design tokens
+- requires `v2.0.0` of the `vf-design-tokens` package or newer.
 
 ### 2.0.2
 
--- a/components/vf-box/CHANGELOG.md
+++ b/components/vf-box/CHANGELOG.md
@@ -1,3 +1,7 @@
+### 2.1.0
+
+- updates spacing design tokens
+
 ### 2.0.2
 
 - fixes bug with `-link` variant by adding position relative to `vf-box`
--- a/components/vf-banner/CHANGELOG.md
+++ b/components/vf-banner/CHANGELOG.md
@@ -1,6 +1,7 @@
 ### 1.6.0
 
 - updates spacing design tokens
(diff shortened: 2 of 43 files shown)


sturobson:
Noted, from the last commit.

This PR makes a changed to the generated utility classes that make use of the spacing design tokens. 

These are, currently, utility classes for `margin`, `padding`, and `gap`.

We would need to inform teams of this change - because it will, until fixed, break layouts that make use of these utility classes. Enforcing the idea that utility classes should be a rarity, and not the norm.

crude table: 

| Current Term | New Term |
| 0                     | 0                |
| xxs                  | 50             |
| xs                    | 100           |
| sm                   | 200          |
| md                   | 400          |
| lg                     | 500          |
| xl                     | 600          |
| xxl                   | 800.         |
| xxxl                 | 1200.        |

components/vf-box/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+### 2.1.0
+
+- updates spacing design tokens
+
```
khawkins98, line 5: On these could we add a note that it `requires vf-design-tokens@2.0.0 or newer`?

sturobson: yep - when I opened this up this morning I did think it was a little 'vague'.

khawkins98 (approved):

PR #1159, opened 2020-10-16
bug: vf-hero css typo

khawkins98:
@sturobson one for your review looks like there's an extra `)` in this bit of CSS. Was stopping the `gulp vf-core:prepare-deploy` and I suspect not doing what was intended.

No need for new changelog as we've not yet push to npm


sturobson (approved):
spotted this yesterday -- not sure how it managed to get it :-/

PR #1160, opened 2020-10-16
Fix: Wallace failing CI builds

khawkins98:
Disabled Wallace as its been failing with `curl: (22) The requested URL returned error: 412 Precondition Failed`

https://travis-ci.org/github/visual-framework/vf-core/builds/736292643#L713

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,8 @@ before_script:
   - cd tools/vf-component-library
   - yarn install
   - gulp build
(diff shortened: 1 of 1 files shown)


khawkins98:
@sturobson: as an aside, is it worth keeping the Wallace integration? I can't recall the last time I looked at the stats, you?

sturobson:
> as an aside, is it worth keeping the Wallace integration? I can't recall the last time I looked at the stats, you?

At the present time, it's probably not worth it.

sturobson (approved):

Issue #974, opened 2020-06-22
FEATURE - Add timestamp to compiled assets


khawkins98:
Would be nice to have a timestamp on the compiled CSS and JS files to see when they were last built.

khawkins98:
Achieved in #1166 

> /*! Component: @visual-framework/vf-componenet-rollup
>  * Version: 1.2.0
>  * Location: components/vf-componenet-rollup
>  * VF Core version in use: 2.2.7
>  * Build time: Mon, 19 Oct 2020 07:53:11 GMT
>  */
> It adds support in vf-core, the component generator and the vf-component-rollup. To do the reaminder of the components, this line will needto be added to each component's Sass file:
> 
>  * Build time: #{map-get($componentInfo, 'buildTimeStamp')}
> Which will of course bump all the versions, perhaps best saved for a less busy time.
> 
>

PR #1166, opened 2020-10-19
Enhancement: echo component generation time

khawkins98:
This addresses #974 so that we get:

```css
/*! Component: @visual-framework/vf-componenet-rollup
 * Version: 1.2.0
 * Location: components/vf-componenet-rollup
 * VF Core version in use: 2.2.7
 * Build time: Mon, 19 Oct 2020 07:53:11 GMT
 */
```

It adds support in vf-core, the component generator and the vf-component-rollup. To do the reaminder of the components, this line will needto be added to each component's Sass file:

```
 * Build time: #{map-get($componentInfo, 'buildTimeStamp')}
```

Which will of course bump all the versions, perhaps best saved for a less busy time.

### DIFF
--- a/components/vf-componenet-rollup/CHANGELOG.md
+++ b/components/vf-componenet-rollup/CHANGELOG.md
@@ -1,3 +1,8 @@
+### 1.2.1
+
+* adds `buildTimeStamp` from `componentInfo` to Sass template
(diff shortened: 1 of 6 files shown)


sturobson:
echo

sturobson (approved):

Issue #1143, opened 2020-10-01
standards: choose a JS linting/formatter


khawkins98:
As raised in #1141, it would be good to have a standardised code format for JS ([like we do for Sass](https://github.com/visual-framework/vf-core/blob/develop/.stylelintrc)).

I have no strong opinions. 

I see that:

- bootstrap uses ESLint https://github.com/twbs/bootstrap/blob/main/.eslintrc.json
- 11ty uses ESLint https://github.com/11ty/eleventy/blob/master/.eslintrc.js

So I'd be happy with [ESLint](https://eslint.org/), going so far as to use 11ty's config as that's where we most often use our JS -- and it's a fairly small config, so it leaves space for future changes. 

Any opinions @pwalter-ebi @sandykadam @sturobson @dbushell ?

khawkins98:
Might be a wider question about using typescript at some point, but ... not yet :D

khawkins98:
In #1168 we've added the JS linting. 

There are many updates needed to the files, and should be done in a separate step.

PR #1168, opened 2020-10-19
Enhancement: adds js linting

khawkins98:
For #1143.

This does not yet usethe the hard fail check as there are many linting errors to fix and it would result in many compoennt updates.


sturobson (approved):
something I hope I never have to wrangle with 😭

khawkins98:
java fun

PR #1174, opened 2020-10-20
the one that refactors vf-intro

sturobson:
- adds the option to have a nicer styled 'sub-heading'.
- introduces the use of `vf-stack` to layout out the vertical stacking of the content.
- use `--vf-stack-margin--custom` to align things as required.
- removes use of `embl-grid` and lets `vf-intro` roll it's own.

### DIFF
--- a/components/vf-intro/CHANGELOG.md
+++ b/components/vf-intro/CHANGELOG.md
@@ -1,3 +1,10 @@
+### 1.4.0
+
+- adds the option to have a nicer styled 'sub-heading'.
+- introduces the use of `vf-stack` to layout out the vertical stacking of the content.
+- use `--vf-stack-margin--custom` to align things as required.
+- removes use of `embl-grid` and lets `vf-intro` roll it's own.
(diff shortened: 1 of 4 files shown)


khawkins98 (approved):

khawkins98:
Not a problem specific to this PR, but perhaps a chance to fix it. The vf-badge looks out of place if the `__heading` is two lines. 

image

sturobson:
Yeah, this is because we're using flexbox. 

We could use `start` or `flex-start` but because the typography has that inherent whitespace (see #865 for what could/would fix this) it looks even worse.

We could, ahem, go back to the late 90s(?) and use some absolute positions - but that will cause more issues than it's worth.

I think we note it for now - and when #865 gets finished and pulled in - we can take another look.

The main bonus in this PR is reducing the whitespace (oh my all those blank pixels) for the bottom margin (and exposing a custom property if you want to make it, dare I say, bigger).

PR #1178, opened 2020-10-22
fix: linting defaults

khawkins98:
Avoid build failure on missing .eslintrc.js config in child projects


sturobson (approved):

khawkins98:
Fyi @filipe-dias Once this is released to npm it will no longer require a local linting file.

PR #1179, opened 2020-10-22
📝 The one that updates the design tokens docs.

sturobson:
🔥 removes quite a bit of design tokens documentation as it's now redundant as you should use the documentation in `vf-component-library`.
✨ introduces new gulp tasks (that might need a tidy up.
📝 updates design tokens documentation in `vf-component-library`.


tools/vf-component-library/gulpfile.js
```diff
@@ -42,3 +47,33 @@ gulp.task('dev', gulp.series(
   'vf-build-search-index',
   gulp.parallel('watch','vf-watch')
 ));
+
```
khawkins98: Would be best if this was a separate gulp file (to keep separation of concerns). Happy to do that if you'd like.

sturobson: that line? 

Happy for you to fix the gulp tasks I wrote into one `series` one and put it in the file system you've got for all the tasks.

khawkins98: Just added a commit that does a tidy and fixes a couple missing tags

https://github.com/visual-framework/vf-core/pull/1179/commits/88e4b003e8e8a1f42d83209b0136460aea72ae3e

khawkins98 (approved):

PR #1180, opened 2020-10-23
the one that tidies up some of the space

sturobson:
I thought I'd already made this a PR.

It tidies up the spacing. For now. Ideally `body` should also have `vf-stack` as an HTML class.

That way we can standardise the spacing between things - and then adjust as needed with the css custom properties.

I've probably got changelogs to write.


khawkins98 (approved):

khawkins98:
Oh, changelog nag

PR #1186, opened 2020-10-26
the one that makes 11ty look for design tokens data

sturobson:
I think this is probably a cleaner way of doing it rather than relying on gulp to do the heavy lifting, we get 11ty to do it.


khawkins98 (approved):

khawkins98:
Nice one

PR #1188, opened 2020-10-26
the one that renames and moves all grid components as layout components

sturobson:
I'm guessing I'd only need to add changelogs to `vf-core` and `vf-component-library` here


khawkins98:
With the way lerna works, this will trigger releases for al these components. — however, we can sneak these in without tagging a release. I was already thinking along those lines for some of the README.md rearranging that is necessary.

Something like this:

1. We tag a release with all the needed npm component updates
2. We merge in all the updates to components that don't need to go npm
3. We add a tag but don't do `lerna publish`
4. PROFIT

khawkins98 (approved):

sturobson:
@khawkins98 did this one need marging in still?

khawkins98:
Yep. Merging now.

PR #1194, opened 2020-10-28
the one that removes card and adds inline checkerboard background for component views

sturobson:
Screenshot 2020-10-28 at 09 01 07


khawkins98:
checkered suit

khawkins98 (approved):

PR #1195, opened 2020-10-28
the one that refactors the masthead

sturobson:
todo: 

- [ ] changelog.
- [ ] update documentation as needed.
- [ ] show other theme variants.

@khawkins98 note - you might have to go into `_preview--nogrid.njk` and change the urls for the css link and the JS script - so they're pointing to 

`<link rel="stylesheet" href="http://localhost:3003/vf-core/css/styles.css">`

`<script src="http://localhost:3003/vf-core/scripts/scripts.js"></script>`

if you're running `gulp dev` from inside `tools/vf-component-library` < that's how I've been doing it so far.


components/vf-masthead/vf-masthead.config.yml
```diff
@@ -12,23 +12,31 @@ variants:
...
+      Primary: true
 
-  - name: kitchen sink
+  - name: image
     context:
-      mastheadImageClass: vf-masthead--has-image
-      mastheadImage: "../../assets/vf-masthead/assets/group-bg_2d4155.png"
-      hasTitleBlock: vf-masthead--with-title-block
+      mastheadImageClass: true
+      mastheadImage: "http://localhost:3003/vf-core/assets/vf-masthead/assets/group-bg_2d4155.png"
```
khawkins98: I know this is in progress, but if we put the images on https://www.embl.org/files we can make use of the CDN and not have to worry about paths or storing the demonstration images in version control.

khawkins98:
Having a look at the JS, I think there's a couple ways to make it work "cleaner".

khawkins98 (approved):

PR #1204, opened 2020-10-30
The one that deprecates the vf-inlay pattern.

sturobson:
Now that we are using and improving the sites that the visual framework use daily. We have noticed that the original ideas around the usage of `vf-inlay` to show a 'site within a site' make things look more jarring than consistent and understandable. 

Since the latest tagged release of `vf-wp` has removed the use of `vf-inlay`. We are now deprecating it from `vf2.0`. 

- removes `vf-inlay` from the component rollup of `vf-core`.
- adds relevant deprecated terms and code to the `vf-inlay` component.
- updates `vf-inlay` changelog. 
- removes and 'inlay' variants of components and related CSS.


khawkins98 (approved):
Snuck in a commit to link the changelogs to this PR. It's useful when doing the update on 
 https://visual-framework.github.io/vf-core/updates/

PR #1209, opened 2020-11-02
The one that gets the logos 'on point'

sturobson:
- Removes `--extreme` responsive version.
- Makes the maximum height 3rem, rather than 6rem.
- Removes some unneeded CSS.

### DIFF
--- a/components/embl-logo/embl-logo.scss
+++ b/components/embl-logo/embl-logo.scss
@@ -12,20 +12,13 @@
 .embl-logo {
   background-image: url('../assets/embl-logo/assets/logo-medium.svg');
   background-repeat: no-repeat;
-  background-size: auto 100%;
-  color: set-ui-color(vf-ui-color--black);
   display: flex;
   flex: 1; // necessary for when embl-logo is inside flex-based containers like vf-global-header
-  height: 32px;
+  height: 2rem;
   text-decoration: none;
-}
-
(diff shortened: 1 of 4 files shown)


components/embl-logo/embl-logo.scss
```diff
@@ -12,20 +12,13 @@
 .embl-logo {
   background-image: url('../assets/embl-logo/assets/logo-medium.svg');
```
khawkins98, line 13: Design mentioned that this should be just plan `/logo.svg` except for on tablet/mobile, is that worth addressing in this PR?

khawkins98: Oh, wait, it looks like that should now work. Disregard!

khawkins98 (approved):

components/embl-logo/embl-logo.scss
sturobson: Oh, so the "more dots" one for everything? I thought it was less dots on mobile/tablet -- which this is.

khawkins98: It's "more dots on desktop". Quickly scanning your code, it looks like you have it right.

sturobson: yeah - I think the SVGs aren't named well. Perhaps that's a `N.x.x` breaking change to add to the todo list

PR #1215, opened 2020-11-05
fixes bug on hover of link and easy variant

sturobson:
@kasprzyk-sz found an issue if the `vf-card` was the `--easy` variant and the heading has a `__link` where the `:hover` shadow wasn't full height.

### DIFF
--- a/components/vf-card/vf-card.scss
+++ b/components/vf-card/vf-card.scss
@@ -112,6 +112,7 @@
     --card-text-color: #{set-ui-color(vf-ui-color--black)};
   }
 
+  // fixes bug on hover of link and easy variant #1215
   .vf-card__link {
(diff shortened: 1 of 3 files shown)


components/vf-card/vf-card.scss
```diff
@@ -111,6 +111,12 @@
   .vf-card__text {
     --card-text-color: #{set-ui-color(vf-ui-color--black)};
   }
+
+  .vf-card__link {
```
khawkins98, line 116: Could we slip a link/comment in here? as I'm sure we'll forget why it's here.

Something like:
```
// fixes bug on hover of link and easy variant #1215
```

khawkins98 (approved):

PR #1221, opened 2020-11-09
Bug: fixes vf-inlay typos

khawkins98:
Fixes typos in Sass introduced in vf-inlay v2.0.0


sturobson (approved):

sturobson:
I guess when this gets pushed to npm we should do the npm deprecate dance - so there's a notice on the `npm install` now too.

sturobson:
https://docs.npmjs.com/deprecating-and-undeprecating-packages-or-package-versions

PR #1225, opened 2020-11-10
docs: update-roadmap

khawkins98:
Adds a couple of milestones and fixes typos. Also adds a blog post that we now have these.


sturobson (approved):
the minor gripe about using `{% render %}` - but it's not a hold-up and I've been thinking perhaps `.md` would be better than `.yml` for using `{% includes %}`

Issue #878, opened 2020-04-22
COMPONENT - A Search Results Page


sturobson:
With the launch of the [Covid 19 Portal](https://www.covid19dataportal.org) and with the [jobs page](https://www.embl.org/jobs/) live too - we should think to standardise the search results pages so they have, at least, they're name spacing. 

The Covid 19 Portal results have some fun things that could be addressed and standardised to meet the system.

khawkins98:
Picking back up on this today.

There's a placeholder at https://latest.visual-framework.dev/patterns/search-results/

We may find ourselves needing two (or perhaps more) variants of a search page, but let's see how far we get with one.

Some use cases to incorporate: 

### embl.org/jobs
![image](https://user-images.githubusercontent.com/928100/98657019-b0251a00-2341-11eb-886a-4fd14b9ee14e.png)

### [COVID19DataPortal](https://www.covid19dataportal.org/sequences?db=embl-covid19)

![image](https://user-images.githubusercontent.com/928100/98657059-bdda9f80-2341-11eb-9f34-44f87adb6766.png)

### [EBI Training](https://wwwdev.ebi.ac.uk/training-beta/search-results?query=dna&domain=ebiweb_training_events&page=1&facets=)

![image](https://user-images.githubusercontent.com/928100/98657157-d6e35080-2341-11eb-9116-85bbccf8b229.png)

### www.embl.org/search

![image](https://user-images.githubusercontent.com/928100/98657242-eebad480-2341-11eb-8015-ffb33b804f0e.png)

### www.embl.org/documents

![image](https://user-images.githubusercontent.com/928100/98657949-bff12e00-2342-11eb-8ab3-a7ed5bd2409f.png)


There are probably more examples already, but this is plenty to chew on.

khawkins98:
A very much work in progress for tomorrow's clinic 

![image](https://user-images.githubusercontent.com/928100/98689185-f7280500-236b-11eb-9f8c-3ec242413244.png)

khawkins98:
Copying some comments from slack in as while I think they've been previously raised, they're not captured here.

![image](https://user-images.githubusercontent.com/928100/101161811-b0a78c80-3631-11eb-9624-6d435a79a6e5.png)


> yeah, would be nice to either push/pull the filters to show at the bottom (and add a “view filters”) or use the vf-details/vf-show-more to hide some by default
> but that’s part of the wider search discussion
> 
[...]

PR #1233, opened 2020-11-13
Docs: search example

khawkins98:
For #878 (create an example Search Results Page)

Also includes some various refinement on form behaviour. 

Very much a WIP.


khawkins98:
The overall search discussion is still a work in progress, however I think it's worth getting this in as it can form the basis of future work and is far better than [the current example](https://stable.visual-framework.dev/patterns/search-results/).

Screen grab:

image

sturobson (approved):

PR #1235, opened 2020-11-13
bug: in vf-intro don't force styles on vf components

khawkins98:
Sometimes it useful to have a button for call-to-actions inside vf-intro, but the current CSS forces `@include inline-link;` on `a.vf-button`

And this ensures any `a.vf-*` element inside vf-intro won't get accidental styles applied.

image


sturobson (approved):
As mentioned at a stand-up this week. I’d like to revisit where we’re using the universal selector on older components (so not `vf-stack` or `vf-cluster`, for example) as I think we can improve them a little. 

This one is a good start!

PR #1242, opened 2020-11-17
The one that introduces variable fonts

sturobson:
- Two http requests
- a saving of ~115kb on page weight
- uses local sources if they're there first still


khawkins98:
Standard "what about IE 11" question? 

https://caniuse.com/variable-fonts

I presume this would mean IE 11 wouldn't get the fonts, is it considered essential?

khawkins98:
Not trying to be a downer, this is good stuff!

sturobson:
>  is it considered essential?

We'd need to determine our level of support, and what that means. 

I'll see if there's some way that IE11 get the other fonts … but I'm not sure who/if that will work.

sturobson:
happy to keep this one languishing for a while. It's not mission critical - but a nice to have.

sturobson:
we'd probably want to look at introducing [font display](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/font-display) here too

khawkins98:
In this case, as Plex is on Google fonts, we could:

- Use JS to detect if IE
```js
if(navigator.userAgent.indexOf('MSIE')!==-1
|| navigator.appVersion.indexOf('Trident/') > -1){
   /* Microsoft Internet Explorer detected in. */
}
```
- If is IE, then inject HTML to load the google fonts

Would obviously not be as performant, but, hey, it's IE11.

khawkins98:
Or, better: use JS to detect if the browser supports variable fonts: https://stackoverflow.com/questions/55729826/detect-variable-font-support-with-javascript

PR #1255, opened 2020-11-25
The one that refactors, enhances, and starts to deprecate some things for the vf-card.

sturobson:
From the changelog:

* adds new `--bordered` and `--striped` design variants.
* added `vf-stack` to the `vf-card__content` element to determine spacing.
* allows for the lack of `vf-stack` for older components.
* started the deprecation of the 'Mortal Kombat' variants, initally by hiding them in vf-core.
* introduced `newTheme` so the 'Mortal Kombat' variants can live side-by-side with news versions for now.
  * the `newTheme` moves us back to the 'primary' being the embl green, the secondary the embl blue, etc.
  * we now remove the `-theme` part of the css class moving forward as it's cleaner, easier to read, and states the same thing without it.
* created theme variants of the new design variants (these are hidden, and should not be used).

### DIFF
--- a/components/vf-card/vf-card.scss
+++ b/components/vf-card/vf-card.scss
@@ -21,7 +21,6 @@
   display: grid;
   grid-template-rows: 288px 1fr;
   position: relative;
-  max-width: 360px;
 }
 
 .vf-card__image {
--- a/components/vf-card/vf-card.scss
+++ b/components/vf-card/vf-card.scss
@@ -21,6 +21,7 @@
   display: grid;
   grid-template-rows: 288px 1fr;
   position: relative;
+  max-width: 360px;
 }
 
 .vf-card__image {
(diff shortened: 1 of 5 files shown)


sturobson:
The only two cards - that should be used for the short future:

`vf-card--bordered`
Screenshot 2020-11-25 at 09 37 42

`vf-card--striped`
Screenshot 2020-11-25 at 09 37 34

components/vf-card/vf-card.scss
```diff
@@ -35,12 +35,14 @@
 
 .vf-card__content {
   grid-column: 1 / -1;
-  padding: 16px 24px;
+  padding: 1rem;
```
khawkins98: Should we use the spacing tokens instead?

khawkins98 (approved):

khawkins98:
As an aside: I wonder if we should change the example image, as the "blobs" are no longer favoured.

components/vf-card/vf-card.scss
sturobson: Yes! I'll go back and do these where applicable.

sturobson:
I've updated the example image for the cards, using an image from emb.org/files now:

Screenshot 2020-11-25 at 10 09 09
Screenshot 2020-11-25 at 10 08 59

PR #1257, opened 2020-11-25
fix: eleventy-command-process

khawkins98:
with beta.1 eleventy has some changes in how its build process works, we'll need to update vf-extensions.

- won't need this until the new 11ty version ships and then we'll need to coordinate releases
- the new 11ty version should trim 50% off our 11ty build times

Todo:

- [ ] I've had to reintroduce `elev.watch();` in the command file, I'm not sure if this can be done differently or the overall watch approach can be updated (CSS updates shouldn't trigger an 11ty rebuild, for example)


khawkins98:
Still waiting on eleventy to do their next release. They've been pretty quiet for a while, last merge to main was in November.

khawkins98:
A new release of 11ty is here https://github.com/11ty/eleventy/releases/tag/v0.12.1

khawkins98:
11ty seems to have made changes since i first opened this PR that makes this fix no longer needed.

However building a site gives many:

```bash
Deprecated as of 10.7.0. highlight(lang, code, ...args) has been deprecated.
Deprecated as of 10.7.0. Please use highlight(code, options) instead.
https://github.com/highlightjs/highlight.js/issues/2277
```

but that's a separate issue

--- 

**UPDATE:** fixed in https://github.com/visual-framework/vf-core/pull/1435

khawkins98:
Ah, figured it out. 0.12.1 is a small update of just a few dependencies. 

There's is a separate 1.0.0-canary release: https://www.npmjs.com/package/@11ty/eleventy

So it does seem probable we'll still need this patch once 1.0.0 beta or rc is out.

khawkins98:
@nitinja with 11ty now at 1.0.0-beta.2 I think we should release this `vf-extensions` as an alpha (`2.0.0-alpha.1`) so that we can begin testing integration. 

It's also worth doing a KT session on how 11ty integrates with the build process.

nitinja (approved):
I didn't understand few of the code blocks, but will try to in possible KT session.

khawkins98:
Merging this in as we have quite a few other dependency updates to test (style lint, marked)

sandykadam:
@khawkins98 does this also need update in EBI Corporate sit build process? If so @nitinja please note & update

khawkins98:
Not immediately, it's an opt-in update at this point. 

If you want to use the eleventy 1.0.0 betas, you'll also need to update to `vf-extensions` 2.0.0-alpha.1

khawkins98:
(I expect that eleventy 1 stable will be out any day now and brings many changes — hopefully not many breaking ones)

nitinja:
@sandykadam in future we can bring this update to our projects, when it releases in stable form

PR #1258, opened 2020-11-25
The one that updates spacing and color functions

sturobson:
* introduces a `space` Sass function to save the keystrokes.
  * instead of typeing `map-get($vf-spacing-map, vf-spacing--400)` you can write `spacing(400)` for the same result.
* I've added this terse naming of the function for `set-color` and `set-ui-color` to be something like `color(green)` instead of `set-color(vf-color--green)`. The old way still works.


sturobson:
I'm going to sneak this one in 😮

khawkins98:
Nice. Would be good to get these in the documents somewhere

PR #1259, opened 2020-11-25
enhancement: multiple-vf-show-more

### LINKED ISSUE #1243: Enhancement: support more than one vf-show-more on a page
For the work to improve forms (#1229) we need to improve [`vf-show-more`](https://stable.visual-framework.dev//components/vf-show-more/) to support more than one element per page.

> Note: currently only one set of "show more" is supported per page

We should be able to draw from vf-tabs to use [`.closest`](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest) (with polyfill for IE)

khawkins98:
Support more than one vf-show-more on a page 

Closes #1243

### DIFF
--- a/components/vf-show-more/vf-show-more.njk
+++ b/components/vf-show-more/vf-show-more.njk
@@ -34,3 +34,22 @@
   <button class="vf-button vf-button--primary vf-button--sm | vf-show-more__button" data-vf-js-show-more-button>Show more</button>
   <button class="vf-button vf-button--primary vf-button--sm | vf-show-more__button--less" data-vf-js-show-more-button--less>Show less</button>
 </section>
+
+
+{# A second set to show you can have more than one set on a page #}
+
+<section data-vf-js-show-more data-vf-js-show-more-pager-size="2" class="vf-show-more">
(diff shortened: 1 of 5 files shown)


components/vf-show-more/vf-show-more.njk
```diff
@@ -34,3 +34,22 @@
...
+
+{# A second set to show you can have more than one set on a page #}
+
+<section data-vf-js-show-more data-vf-js-show-more-pager-size="2" class="vf-show-more">
+  {% render '@vf-section-header--default', {
+    "section_title": "Show more demo part 2"
+  } %}
+
+  <ul class="vf-list vf-list--ordered">
+    <span class="vf-show-more__item"><li class="vf-list__item">a list item</li></span>
```
sturobson: the HTML is invalid here - the only direct child of an `ul` (or `ol`) that's valid is an `li`.

khawkins98: Good catch. my css method could have been better. resolved.

sturobson (approved):

PR #1260, opened 2020-11-25
Fix: update and optimise vf core install

khawkins98:
This greatly de-complexifies the VF npm development process.

- no longer need to install the root level
- `yarn install` the thing you need (i.e. tools/vf-core)
- updates the related docs
- also deploys the `develop` branch to on develop

When testing this branch be sure to do a `git clean -fxd`


khawkins98:
@sturobson  This should also make the `(node:97882) UnhandledPromiseRejectionWarning: Error: Fractal themes must inherit from the base Theme class.` message go away for good.

khawkins98:
Looks like the netlify deploy process would need to be updated after this.

sturobson (approved):

sturobson:
@khawkins98 can confirm this works

sturobson:
good job

sturobson:
@khawkins98 actually -- maybe not -- getting 'errors' on lots of component pages _and_ all the grid variants have disappeared.

ran git clean -fxd in both the root and tools/vf-core

khawkins98:
It looks like there have been deeper changes to the theming layer in Fractal. I'll see how much of a rebuild it might be for vf-frctl-theme. 

Probably a separate issue.

PR #1263, opened 2020-11-26
fix: vf-sass-config import order

khawkins98:
`variables/vf-global-custom-properties.scss` should be imported after `mixins/vf-mixins.scss`

From:

```css
/* --- test */
.test {
  padding: space(400);
  --test: #{space(400)};
}
```

I now can get the expected:

```css
/* stylelint-enable */
:root {
  --vf-body-width: 81.25em;
  --page-grid-gap: map-get($vf-spacing-map, vf-spacing--400);
  --embl-grid-module--prime: 16rem;
  --embl-grid-spacing-normaliser: 0px;
  /* custom prop test */
  padding: 1rem;
  --test: 1rem;
}
```


sturobson (approved):
it was as simple as that! 

simple

PR #1266, opened 2020-11-27
the one that makes some small tweaks to various summary variants

sturobson:
This should help with the 2020 home page re-jig-athon.

### DIFF
--- a/components/vf-summary/vf-summary.scss
+++ b/components/vf-summary/vf-summary.scss
@@ -92,6 +92,7 @@
 
 .vf-summary__title {
   @include set-type($vf-summary__title-typography);
+  font-weight: 400;
 }
 
 .vf-summary__author {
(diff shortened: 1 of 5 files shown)


khawkins98 (approved):

components/vf-summary/vf-summary.scss
```diff
@@ -92,6 +92,7 @@
 
 .vf-summary__title {
   @include set-type($vf-summary__title-typography);
+  font-weight: 400;
```
sturobson, line 95: @khawkins98 the more I look - the more this looks wrong. 

Can you remember the rationale - was it because the preceding heading loses significance slightly because they're similar?

khawkins98: I think it was just design feedback that the news headings here looked "a bit too bold" at the original, so they were made 400: 

image

Agree that it's a hack in the system. Not sure if the other typography changes might make it redundant

PR #1268, opened 2020-11-27
the one that adds a max-content width so shorter titles

sturobson:
one thing I noticed while pre-emptively looking at `vf-hero` being used on embl.org was that our current (`v2.0.1`) has a `max-width` of `60ch`. 

This makes some things look wrong and gives them too much of the dreaded white space.

before:
Screenshot 2020-11-27 at 14 45 24

after:
Screenshot 2020-11-27 at 14 46 01


khawkins98 (approved):

sturobson:
@khawkins98 note the component version has jumped here too

Issue #1271, opened 2020-12-02
📒 Documentation: a basic how-to on integrating VF JavaScript into React


khawkins98:
While the VF uses ES6 JS Modules, it's not straightforward on how to pass these to React and integrate them into the component lifecycle.

I was discussing vf-tabs on the [EBI People service](https://gitlab.ebi.ac.uk/ebiwd/people-service/-/merge_requests/61) with @meladawy and have a method we could integrate as reusable guidance.

## Things for all VF JS components

- [ ] add a TypeScript definition file so React can integrate more easily 
    - `declare module '@visual-framework/vf-tabs/vf-tabs.js';`
    - https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-d-ts.html
- [ ] update [`vf-react`](https://github.com/visual-framework/vf-react) to demonstrate this approach

## General guidance

1. yarn install your VF component as usual
2. Import the VF Tabs js module
    - `import { vfTabs as VfTabsJs } from '@visual-framework/vf-tabs/vf-tabs.js';`
3. Use memo components with a "callback" to execute the VF Tabs JS on the client (much as you would for jQuery in React)

(see full example below)

Note: we acknowledge this is not the most DOM performant method, but for the basic JS components in the VF, this should not be a major consideration.

---

## Example

`src/components/vf-tabs/vf-tabs.tsx`

```tsx
import React from 'react';
import { vfTabs as VfTabsJs } from '@visual-framework/vf-tabs/vf-tabs.js';

// This is an initial POC that:
// 1. Imports the VF Tabs js module
// 2. A very basic React memo template
// 3. Executes the VF Tabs JS on the client (much as you would for jQuery in React)
//   - The JS will update when the memo template is re-rendered

// Note: we acknowledge this is not the most DOM performant method, but for the basic JS components in the VF, this should not be a major consideration.

type Props = {
  title: string;
};

class VfRendTabs extends React.Component {
  componentDidMount() {
    console.log('VF Tabs rendered');
    VfTabsJs();
  }

  render() {
    return <div></div>;
  }
}

export const VfTabs = React.memo(({ title }: Props) => {
  return (
    <section className="comment-tabswrapper">
      <div className="vf-tabs">
        <ul className="vf-tabs__list" data-vf-js-tabs>
          <li className="vf-tabs__item">
            <a className="vf-tabs__link" href="#vf-tabs__section--1">
              {title}
            </a>
          </li>
          <li className="vf-tabs__item">
[...]

PR #1278, opened 2020-12-04
Enhancement: create reusable tooling for react

khawkins98:
For #1271, I've come up with a way that we can use components in React like so:

`<VfLogo title="test" VfLogoPath={VfLogoPath} />`

We do this by:

- reusing our precompiled NJK templates
- loading nunjucks-slim in react
- writting a small amount of wrapper code, extract:

```js
export const VfLogo = React.memo(({ title, VfLogoPath }: VfLogoProps) => {
  return (
    <div>
        vfNunjucks.render('vf-logo', { logo_href: '/vf-react', image: VfLogoPath, logo_text: title })
    </div>
  )
});
```


sturobson (approved):
Will using `render` here create a pain point when things change like we had with recent `vf-hero` updates in static-html-pages?

It only will be a problem if it's `render vf-component--variant` rather than `render vf-component` with the variants added 'inside' the render.

Or have I got the implementation all wrong here?

khawkins98:
So far I've only tried with very simple components and no variants, I'll have a look at how variant would be handled.

khawkins98:
> Will using render here create a pain point when things change like we had with recent vf-hero updates in static-html-pages?

As an aside (and deep dive on Nunjucks templating, as I understand it) you can't "include" a template in this context. The precompiled templates are "self-isolated" and can only be created and rendered.  https://mozilla.github.io/nunjucks/api.html#template

`include` is a way of extending an existing "render environment".

So that's why include works better in 11ty: it extends the existing render function instead of creating a new isolated "render environment"

With react there is, of course, no pre-existing "global nunjucks render environment".

Again, I might be wrong here, but this how I've come to understand the architecture.

sturobson:
Cool. 

From working on updating the `vf-hero` and `vf-card` in static-html-pages I think the way forward is to:

- never `render` the variant specifically but add the variant key/value pair 'inside'

So rather than (in nunjucks):

```
{% render '@vf-card--striped', {
      "component-type": "block",
      "newTheme": "primary",
      "card_href": "https://www.embl.es/",
      "card_title": "Barcelona",
      "card_image": "/assets/emblstatic-homepage/assets/sites/barcelona.jpg",
      "card_text": "EMBL’s site in Spain specialises in tissue biology and disease modelling"
} %}
```

Do

```
{% render '@vf-card', {
      "component-type": "block",
      "newTheme": "primary",
      "variant": "striped",
      "card_href": "https://www.embl.es/",
      "card_title": "Barcelona",
      "card_image": "/assets/emblstatic-homepage/assets/sites/barcelona.jpg",
      "card_text": "EMBL’s site in Spain specialises in tissue biology and disease modelling"
} %}
```

It's still a 'hassle' if there are breaking change updates -- but it would, I think, be easier in the long run.

khawkins98:
Nice would play well with precompiled templates too

khawkins98:
Continuing through this I've hit a complication that we can't import .jsx template from node_modules while using React Create App (https://github.com/facebook/create-react-app/issues/5274#issuecomment-426721106) 

I've worked around this by using native ES6 templates. The syntax is more tedious, but it's much of the same. Update coming.

khawkins98:
Todo for monday:

- [x] revisit docs written
- [x] add a couple more example components (to be considered: tabs, box, button)
- [x] add vf-component.react.js template
- [x] add docs to the component library
- [x] add placeholder "react example" to library

Once published
- [ ] update vf-react

khawkins98:
This should now be ready. I'll plan to review again in the morning.

PR #1285, opened 2020-12-07
the one that improves spacing and usability of video teaser

sturobson:
I made a start on removing the `margin-bottom` of components so we can use `vf-stack` to help space things properly but…

vf-video-teaser seemed to be missing some things and needed an update all by itself, although I don't think this has been used anywhere yet (there's no sign of it static-html-pages).

Update includes:

* Uses `vf-stack` to space component.
* Adds `if` statement for the heading.
* Makes it possible to have more than one teaser.
* Makes the link the whole 'item' like `vf-card`.
* Move the default to a variant, as this should help for when people use `{% render … %}`

### DIFF
--- a/components/vf-video-teaser/vf-video-teaser.scss
+++ b/components/vf-video-teaser/vf-video-teaser.scss
@@ -32,9 +32,6 @@
     position: absolute;
     right: 0;
     top: 0;
-    transition-duration: 300ms;
-    transition-property: box-shadow;
-    transition-timing-function: ease-in-out;
     z-index: 512;
   }
 
--- a/components/vf-video-teaser/vf-video-teaser.scss
+++ b/components/vf-video-teaser/vf-video-teaser.scss
@@ -12,12 +12,42 @@
 .vf-video-teaser__image {
   display: block;
   height: auto;
-  margin-bottom: 16px;
   max-width: 100%;
   width: 100%;
 }
 
 .vf-video-teaser__title {
-  @include set-type(text-heading--3);
-  margin: 0 0 16px 0;
+  @include set-type(text-heading--3, $custom-margin-bottom: 0);
+}
+
+.vf-video-teaser__link {
+  @include set-type(text-body--2, $custom-margin-bottom: 0);
+
+  display: inline-block;
+
+  &::after {
+    bottom: 0;
+    content: '';
+    left: 0;
+    position: absolute;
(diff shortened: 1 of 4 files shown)


components/vf-video-teaser/vf-video-teaser.scss
```diff
@@ -12,12 +12,42 @@
...
+  display: inline-block;
+
+  &::after {
+    bottom: 0;
+    content: '';
+    left: 0;
+    position: absolute;
+    right: 0;
+    top: 0;
+    transition-duration: 300ms;
```
khawkins98: What's the animation for? I didn't seem to notice any practical effect?

khawkins98:
Two other comments:

- I'm not quite sure what the use of vf-video-teaser is? Is it like a "click here to see the video"? Might be a documentation thing.
- I've not been using periods at the end of changelogs, should we? Is there a "best way" or is this a tabs vs spaces thing?

components/vf-video-teaser/vf-video-teaser.scss
sturobson: Oh, a left over from the `vf-card` interaction. an be removed.

sturobson:
> * I'm not quite sure what the use of vf-video-teaser is? Is it like a "click here to see the video"? Might be a documentation thing.

As mentioned, I can't see this being used in the `static-html-pages` repo and I guess _really_ it's more like a `vf-section-header` with `vf-card`s as needed. I'd be much happier removing this as something 'specific' and pushing people to use already available components.

> * I've not been using periods at the end of changelogs, should we? Is there a "best way" or is this a tabs vs spaces thing?

I try to add periods, from an old memory that some screen readers will read out the lists as one big sentence without them (but I'm going back 5/6 years).

sturobson:
> I'm not quite sure what the use of vf-video-teaser is? Is it like a "click here to see the video"? Might be a documentation thing.

I'd be quite happy to change this to a 'deprecated' type of PR

khawkins98:
Deprecation seems reasonable. But happy to wait on it, will leave it up to you. 

I suppose we do have other components that could fill the use case.

khawkins98 (approved):

sturobson:
I think we sneak this one in with the look to deprecate it as a minor release.

khawkins98:
👍

PR #1291, opened 2020-12-09
Use react-dom-fragment to return html fragments

khawkins98:
Death to div wrappers in React.


khawkins98:
Added some more commits here as it:

- add support for nunjucks extensions
- simplifies maintainability 
- removes some unneeded dependencies

sturobson (approved):

Issue #1231, opened 2020-11-12
📒 Documentation: clarify component interoperability


khawkins98:
Many developers expect that VF components should all share the same version, that is: it feels weird to use `vf-button@1.2.3` with `vf-card@2.2.2`.

From a developer:

> it's also difficult to know which version of the vf package to use. I started updating the versions to the latest but some are at version 3xx and other are at 1xx then it's not clear if they are compatible with each other

We should make the versioning practice clearer.

We have already have these pages, which aren't so easy to find:

- https://stable.visual-framework.dev/developing/components/updating-a-component/
- https://stable.visual-framework.dev/developing/guidelines/versioning/

sturobson:
I know Fractal shows the components used in a component.

- Is this possible with our new layout? 
- Is it possible to find the version number and attach it?

khawkins98:
Some documentation at https://latest.visual-framework.dev/guidance/versioning-and-component-interoperability/

By my view, this can now be closed.

Issue #1230, opened 2020-11-12
📒 Documentation: write reference on image handling best practice


khawkins98:
In short: it's best to not let CSS do the image cropping. Resize your images or have your CMS or image proxy do the cropping.

sturobson:
Some thoughts:

- we should still make use of CSS for cropping as the 'fallback' for when content creators don't follow the size guidelines*
- * we need to make the size guidelines
- we should investigate the reintroduction of using width and height to help the [layout engine in browsers](https://www.smashingmagazine.com/2020/03/setting-height-width-images-important-again/).

- more to follow, I'm sure.

khawkins98:
Stub documentation at https://latest.visual-framework.dev/guidance/image-handling/

Issue #1202, opened 2020-10-29
Create documentation for browser support


peter-walter:

sturobson:
Current documentation from VF1.x 

http://wwwdev.ebi.ac.uk/style-lab/websites/meta-patterns/browser-compatibility.html

khawkins98:
As part of this would be good to gather browser usage stats for:

- [x] EBI.ac.uk
- [x] EMBL.org
- [x] Major scientific service using VF 2.0
- [ ] Bonus: EMBL.de

Might be others, but that should probably be more than enough "fact-finding".

I know @sturobson is also gathering some examples of prominent support policies -- I think most go like "Last 2 versions of major browsers get gold support, everything else from last 5 years (also IE 11) get a usable experience"

- [ ] Todo: Define what "usable" is

sturobson:
Just some 'titbits': 

The 7 year old IE11 has been given it's 'end of life' dates - https://death-to-ie11.com

Microsoft is starting to drop support for it already - https://techcommunity.microsoft.com/t5/microsoft-365-blog/microsoft-365-apps-say-farewell-to-internet-explorer-11-and/ba-p/1591666

I also remember seeing a tweet, maybe an article - about new Edge having a 'Legacy Mode' for people that are stuck on IE11 because of the proprietary software they're using. The Edge team will actively help you move to this mode in Edge rather than use IE11. I'll ping some Microsoft folks I know if I cannot find it.

khawkins98:
Some stats for 1 Aug - 29 Oct 2020 users.

## EMBL.org

Browser | Note | Traffic %
-- | -- | --
Chrome | 62 and older account for less than 1% | 54.66%
Safari | 11 and older account for less than 1% | 17.84
Firefox |  77 and older account for less than 1% | 9.57
Safari (in-app) | | 4.96
Android Webview | 83 and older account for less than 1%   | 4.29
Edge (chromium) |   | 2.51
Edge classic |   | 1.30
Samsung | 11 and older account for less than 1%  | 1.86
IE | 8, 9, 10 acount for less than 0.3% | 1.35
Opera (desktop) | 69 and older account for less than 0.1%  | 0.60


## EBI.ac.uk

Browser | Note | Traffic %
-- | -- | --
Chrome | 63 and older account for less than 1% | 65.73%
Safari | 11 and older account for less than 1% | 11.70
Firefox |  77 and older account for less than 1% | 11.27
Edge (chromium) |   | 3.72
[...]

PR #1298, opened 2020-12-14
Documentation: add guidance section

khawkins98:
This adds a new section of "guidance" to accommodate a number of issues around how and why: 

- #1202 Create documentation for browser support 
- #1230  Documentation: write reference on image handling best practice
- #1231 Documentation: clarify component interoperability
- how to approach UX, accesibility
- screen size support
- updating components
- how to handle printing
- customising the VF for a non-embl project 
- lots of other odds and ends 

Some of these might be specific to the VF, and some might be a collection of links to other resources.


khawkins98:
This should now be ready, it adds some stub guidance and organises things a bit into a new "Guidance and principles" section.

Summary: 

> Here you will find guidance on what, why and how.
> This is a collection of how-tos and best practice when building for the Visual Framework and its users.

This allows us to streamline the "Developing" section and accommodate this type of best practice guidance. 

As always, docs aren't done, but a work in progress.

sturobson (approved):

khawkins98:
Made a couple more additions/tweaks here, including adding an "edit this page" link to each footer.

Merging in and will probably pick up more docs works later, including a look at adding an "11ty nunjucks syntax" example that uses "context"

Issue #1036, opened 2020-08-04
Component: vf-summary--news block


khawkins98:
The [vf-summary--news-has-image](https://visual-framework.github.io/vf-core/components/detail/vf-summary--news-has-image.html) is inline, but it's shown through component-hackary as a block in many places. 

![image](https://user-images.githubusercontent.com/928100/89271760-74413100-d63d-11ea-94e5-e9cb6153046c.png)

This is the code that seems required to do it

```css
  .summary-inline .vf-summary {
    grid-template-columns: none;
  }

  .summary-inline .vf-summary > .vf-summary__date {
    grid-row: 7;
  }

  .summary-inline .vf-summary > .vf-summary__image {
    max-width: 100%;
    margin-bottom: 16px;
  }

  @media (max-width: 600px) {
    .summary-inline .vf-summary {
      grid-row-gap: 0;
    }
  }
```

Aside: the [--has-image](https://visual-framework.github.io/vf-core/components/detail/vf-summary--news-has-image.html) demo image isn't showing. We could use the [image proxy](https://acxngcvroo.cloudimg.io/v7/https://www.embl.org/news/wp-content/uploads/2020/07/organ_formation_embryo-Hartmann-wordpress.jpg?w=300&h=200) to simplify things.

@sturobson seems like one for you, assigning.

khawkins98:
Leaving this one open until we've swapped all the local CSS for the container so we can ensure a smoother upgrade in the futuere.

sturobson:
Can this be closed @khawkins98 ?

khawkins98:
I think so, in any case if a page has "done it wrong" that's not this issue's fault.

PR #1301, opened 2020-12-15
Enhancement: Add vf-news-container--featured

khawkins98:
This addresses the long-lingering #1036, where we've been sprinkling local CSS to achieve a "Featured news" container.

Even though the vf-summary and vf-news-container have uncertain futures, this would stop the propagation of "local CSS" across sites to enable a smoother transition to "the new thing", once we figure out what it is.

### DIFF
--- a/components/vf-news-container/CHANGELOG.md
+++ b/components/vf-news-container/CHANGELOG.md
@@ -1,3 +1,8 @@
+### 1.0.0-rc.1
+
+* Adds vf-news-container--featured
+  * https://github.com/visual-framework/vf-core/issues/1036
+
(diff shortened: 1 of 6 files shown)


sturobson:
Do we want to introduce this -- with the potential of `vf-region` replacing it early Q1 of 2021?

khawkins98:
My view is less that we're introducing something as we're clearly already using it. I think by switching to this, it will be a lot easier to replace it with however vf-region is implemented.

It will be "change X to Y" instead of "go look for some local CSS and try using Y".

sturobson (approved):

PR #1302, opened 2020-12-16
The one that makes components stack (a little)

sturobson:
Now that we are adding `vf-stack vf-stack--400` to the <body> element (blogpost on this is in the works) there are some components that we need to remove the margin from. 

To start with I've only changed two components:

- `vf-global-header`
- `vf-navigation--main`

As _everything else_ has a mix-match of top or bottom margins and we are looking to create a more 'generic' `vf-region` component I think we can leave it for now.

We are aiming to:

-  'push' down by only using margin-top.
- make components 'free' from their margins and get it dictated by the parent. So a blocks margin will get it from the container and a container will get it from the page.

One thing to note is that the `vf-footer` on the homepage might need some additional CSS in the `static-html-pages` projects CSS to remove the margin.

We also need to update the global header used via the content hub -- the markup is woefully out of date.

### DIFF
--- a/components/vf-global-header/CHANGELOG.md
+++ b/components/vf-global-header/CHANGELOG.md
@@ -1,3 +1,9 @@
+### 3.1.0
+
+* updates to expect `vf-stack` on the body.
+* removes the backwords compatability as we're at v3.x.
+* tidies up and removes some unrequired CSS.
+
 ### 3.0.1
 
 * removes padding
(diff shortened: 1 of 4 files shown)


khawkins98:
Looks good.

On the contentHub global header, I think we intentionally were using slightly older markup as there were quite a few sites that still needed to update, I think that's been sorted now. So we can revise in the contentHub or, at least, say: "you should have updated by now"

khawkins98 (approved):

sturobson:
> So we can revise in the contentHub or, at least, say: "you should have updated by now"

As it was deprecated in `v2.x` and we're now in `v3.x` I think it's a good time to update it.

If only I could find it in the gitlab repo there'd be an MR already 😭

khawkins98:
The global header is actually "content" in the contentHub, I'll update it in a few.

PR #1304, opened 2020-12-17
The one that makes the `vf-figure` responsive.

sturobson:
The `vf-figure` was pretty strict in the amount of space it took up, and the use of a CSS custom property to change the width isn't really viable in places like blog pots. 

Highlighted by @kasprzyk-sz - this PR makes the `vf-figure__image` act like a 'responsive image' by defaulting to the image size as the width. Rather than 'brute forcing' it with `width: 100%`.

This should make things a lot easier for usage, moving forward.

### DIFF
--- a/components/vf-figure/CHANGELOG.md
+++ b/components/vf-figure/CHANGELOG.md
@@ -1,3 +1,9 @@
+### 1.4.0
+
+* Removed `width: 100%` from the `.vf-figure__image` class.
+* Added `display: block` to the `.vf-figure__image` class.
+* Removed CSS for the width when the `vf-figure` is using floats.
+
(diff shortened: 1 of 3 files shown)


khawkins98:
Looks good, two comments:

- should this be a breaking change?
- noticed the readme.md has a couple of references to that custom prop https://github.com/visual-framework/vf-core/blob/develop/components/vf-figure/README.md

khawkins98 (approved):

sturobson:
> should this be a breaking change?

I guess it should be more a `1.x.x` than `x.1.x`. I'll update it.

> noticed the readme.md has a couple of references to that custom prop

I'll update.

PR #1316, opened 2021-01-08
Chore: update ebi footer template

khawkins98:
Now that we have an EBI VF 2.0 footer, this updates the component to use it (as well as doing some other cleanup and docs).

We still need to make an EBI boilerplate example.


sturobson (approved):

sturobson:
I wonder, now that it is an import - if it needs to be in the vf-core at all? 

a `vf-imports` section?

khawkins98:
There's still actually quite a bit of CSS for the header. https://deploy-preview-1316--ecstatic-noether-150ab4.netlify.app/components/ebi-header-footer/#assets

But once (if?) that goes full 2.0 we could probably remove the component and this could just be a contentHub thing. But the EBI header will probably be a question for some time ...

sturobson:
> But the EBI header will probably be a question for some time ...

... and then some.

PR #1325, opened 2021-01-12
Pattersn: embl and ebi boilerplate

khawkins98:
This doesn't do anything exotic, but creates a standard reference of how an EBI page has the black bar and the VF 2.0 footer.

More work to be done but will ease a lot of confusion by EBI devs.

It also updates the _data/containers to link to the njk templates in a way that works for vf-component-library.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -39,8 +39,15 @@
   ],
   "eslint.lintTask.enable": true,
   "cSpell.words": [
+    "EMBL",
+    "EMBL’s",
+    "Wellcome",
     "eleventy",
+    "lede",
     "precompiled",
-    "roadmap"
+    "roadmap",
+    "s",
+    "stylesheet",
+    "subsidised"
   ]
 }
--- a/tools/vf-component-library/src/site/patterns/boilerplate-generic-embl-ebi.njk
+++ b/tools/vf-component-library/src/site/patterns/boilerplate-generic-embl-ebi.njk
@@ -34,12 +34,6 @@ layout: layouts/boilerplate.njk
   -->
 </head>
 <body class="vf-body vf-stack">
-
-<style>
-.ebi-header-footer.masthead-black-bar {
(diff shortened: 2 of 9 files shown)


.vscode/settings.json
```diff
@@ -39,8 +39,15 @@
   ],
```
sturobson, line 39: aside: I wonder if this should be on the `.gitignore` list?

tools/vf-component-library/src/site/patterns/boilerplate-generic-embl-ebi.njk
```diff
@@ -0,0 +1,129 @@
...
+    Want to build a site with the Visual Framework?
+    ---
+    You can copy the CSS and JS you find here, but for the best experience see the building guide:
+    https://stable.visual-framework.dev/building/
+  -->
+</head>
+<body class="vf-body vf-stack">
+
+<style>
+.ebi-header-footer.masthead-black-bar {
```
sturobson: is this because the gdpr banner loads first? Ideally, we'd want this in the actual CSS somehow - I did make a start before Xmas on something like this for ordered spacing depending on the preceding component -- but not sure where the CSS should live

sturobson (approved):

.vscode/settings.json
khawkins98: I go back and forth. Sometimes I commit to my user space and other times to the project. Usually I commit to the project if I already added exceptions on my local spell checker.

tools/vf-component-library/src/site/patterns/boilerplate-generic-embl-ebi.njk
khawkins98: Whoops, added that as a fix elsewhere didn't mean to commit it!

khawkins98: Now removed.

PR #1328, opened 2021-01-14
fixes some issues with vf-grid when in IE11 land

sturobson:
Now that I've IE11 running easily on my machine I noticed that the `vf-grid` was not laying out the components as well as it should.

The first was that if there were multiple rows only the first row 'fitted'. The others started with the left margin.

The second was that - the grid didn't seem to have the correct widths so they were not taking up as much space as they could do.

This PR fixes these issues.

Note: We don't do any 'responsive' things here - I've looked on several stat sites for browser market share and viewport market share and think the chances of an IE11 browser not being the full width of the screen will be extremely low. So we are treating it only as a 'desktop browser' with this flexbox grid fallback.

### DIFF
--- a/components/vf-grid/CHANGELOG.md
+++ b/components/vf-grid/CHANGELOG.md
@@ -1,4 +1,4 @@
-### 1.3.0
+### 1.4.0
 
 * fixes flexbox fallback grid when there are items on two or more rows
 * fixes widths on flexbox fallback grid.
--- a/components/vf-grid/CHANGELOG.md
+++ b/components/vf-grid/CHANGELOG.md
@@ -1,5 +1,10 @@
 ### 1.3.0
 
+* fixes flexbox fallback grid when there are items on two or more rows
+* fixes widths on flexbox fallback grid.
+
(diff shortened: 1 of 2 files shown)


components/vf-grid/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 ### 1.3.0
 
+* fixes flexbox fallback grid when there are items on two or more rows
+* fixes widths on flexbox fallback grid.
+
+### 1.3.0
```
khawkins98, line 6: I think you meant to add 1.4.0?

khawkins98 (approved):

khawkins98:
> Note: We don't do any 'responsive' things here - I've looked on several stat sites for browser market share and viewport market share and think the chances of an IE11 browser not being the full width of the screen will be extremely low. So we are treating it only as a 'desktop browser' with this flexbox grid fallback.

This might be an interesting point for "legacy" browser support. "We support legacy browsers on desktop, not mobile"

PR #1333, opened 2021-01-20
small changes to the max-width

sturobson:


khawkins98:
👍  

Also need a changleog for vf-design-tokens.

khawkins98:
Saw a few other places 81.25 is coded in, also should be updated?

image

sturobson:
Should be ready to go - I've not updated `grid-page` as it's deprecated (probably should be deleted from the codebase at some point?)

khawkins98:
> probably should be deleted from the codebase at some point?

Good point on the archive code. Maybe just hide the component and add it to the .gitingnore so it stays out of our searches?

khawkins98 (approved):

components/vf-body/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+### 1.2.0
```
khawkins98: Just noticed. should be 1.1.0?

khawkins98 (approved):

PR #1343, opened 2021-02-01
Starts adding 'design token decisions'

sturobson:
This should also help teams that want to use a different colour for something (like pink for links instead of blue) by:

a) defining the 'design decisions' for things as tokens
b) have them as Sass variables so they're:
  1. easier to implement.
  2. easier to override.

I don't think it makes sense to have these available outside of Sass maps.


components/vf-design-tokens/gulpfile.js
```diff
@@ -106,7 +114,17 @@ gulp.task("tokens2:json", () =>
...
 
 gulp.task("tokens:props", () =>
-  gulp.src(["./src/maps/*.yml"])
+  gulp.src([
+    "./src/maps/vf-colors.yml",
+    "./src/maps/vf-spacing.yml",
+    "./src/maps/vf-themes.yml",
+    "./src/maps/vf-ui-colors.yml",
+    "./src/maps/vf-zindex.yml",
+    "!./src/maps/vf-color__background.yml",
```
khawkins98, line 123: Is this so they don't get generated as CSS custom props in the CSS?

sturobson: Yes, I think as they're 'design decisions' they're not something that would necessarily be benefited by being CSS custom properties.

I like the idea of -- if it's something that can be changed by the something in the browser (viewport size for example) or by a content creator (inside a CMS) then it should be a CSS custom property, but if it's essentially a 'static value' then Sass can pick up that Dx weight.

khawkins98:
So, thinking through this. We would:
```
.vf-button {
 color: $ui-color--white; // stop doing this
 color: $vf-color__button__text--primary; // now do this
}
```

yeah?

sturobson:
> So, thinking through this. We would:
> 
> ```
> .vf-button {
>  color: $ui-color--white; // stop doing this
>  color: $vf-color__button__text--primary; // now do this
> }
> ```
> 
> yeah?

Yes, we could (although it might be more of a logistical issue in years to come) get those variables to be created inside `vf-button.variables.scss` so that they're 'collocated' with the component. 

I think this also might help towards the 'theming' aspect of things when it comes to sub-orgs like the covid portal where they can 'bring their own' `json` file for tokens or something. But it might be better to add yet-another abstraction for that.

khawkins98:
> I think this also might help towards the 'theming' aspect of things when it comes to sub-orgs like the covid portal where they can 'bring their own' json file for tokens or something. But it might be better to add yet-another abstraction for that.

That's where my mind is right now. To stop picking specific colours (typically) in the components and instead referring to the decisions/theme colours. 

Which would make it a lot simpler and less hacky when a team like the covid portal needs to retheme things.

khawkins98:
I don't think I saw a changelog -- how about releasing this one as an alpha or beta?

khawkins98 (approved):

sturobson:
I've moved things around a little so there's a separate ['theme'](https://github.com/visual-framework/vf-core/blob/3027cf873c5f933486db847e9f6b8871ba4cd7f6/components/vf-design-tokens/src/theme.palette.alias.yml) alias file that we can start to use for defining 'design decisions' 

This should start to pave the way for other teams (like the covid portal) to create their own `theme.palette.alias` file which we can (yet to be looked at) inject into the `vf-design-tokens` folder.

components/vf-design-tokens/src/theme.palette.alias.yml
```diff
@@ -0,0 +1,53 @@
...
+  # Primary Button
+
+  vf-color__button__text--primary: '#FFFFFF'
+  vf-color__button__border--primary: '#3B6FB6'
+  vf-color__button__background--primary: '#3B6FB6'
+  vf-color__button__shadow--primary: '#193f90'
+
+  # Secondary Button
+
+  vf-color__button__text--secondary: '#3B6FB6'
```
khawkins98, line 43: As this is "EMBL Blue", is there a way to avoid repeating the hex value ... or I suppose the (evenutal) Figma token export should take care of that extrapolation?

sturobson: Yeah, it _should_ do I think.

The only way to stop the duplication in code is to actually move everything to Style Dictionary - as Theo cannot do this.

It's a move we probably need to make at some point soon anyhow.

sturobson: the problem being we cannot nest and alias file inside an alias file in Theo, I think it's possible in Style Dictionary -- but as you say - coming from Figma into code should nullify this.

sturobson:
I'm going to merge this now and tomorrow I'll make the move to Style Dictionary in anticipation.

khawkins98:
scribing

PR #1348, opened 2021-02-03
The greatest documentation quest 2020 - vf-button edition.

sturobson:
I've added some usage criteria to the documentation for the `vf-button` here. 

I've also removed variants that we're not looking to use as part of a simpler system. 

note: I've switched the label for the `--outline` variants to 'secondary' as the use case.

I know we're still to confirm, so I've not removed any CSS so we don't break anything until we move to `v2.0.0`

### DIFF
--- a/components/vf-button/vf-button.config.yml
+++ b/components/vf-button/vf-button.config.yml
@@ -9,10 +9,11 @@ variants:
     context:
       text: a primary button
       theme: primary
-  - name: secondary
+  - name: outline
     context:
-      text: a secondary button
-      theme: secondary
+      text: an outline style
+      theme: primary
+      style: ["outline"]
   - name: tertiary
     context:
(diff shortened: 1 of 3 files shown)


components/vf-button/vf-button.config.yml
```diff
@@ -22,30 +23,6 @@ variants:
       text: a small button
       theme: primary
       size: sm
-  - name: regular
```
khawkins98, line 25: I think removing them will break templates explicitly using that variant (most likely in WP). To avoid a 2.0.0 release and make Szymon's life easier, a `hidden: true` should achieve the same aims in your PR?

sturobson: 👍

khawkins98 (approved):

PR #1353, opened 2021-02-05
The greatest documentation quest 2020 - vf-card edition.

sturobson:


components/vf-card/vf-card.config.yml
```diff
@@ -3,115 +3,150 @@ title: Card
 # Label shown on index pages
 label: Card
 status: alpha
+display:
+ width: 440px
```
khawkins98: Does this do anything? Doesn't seem to be used in the njk file?

khawkins98:
Changelog nag :)

components/vf-card/vf-card.config.yml
sturobson: It renders the max-width of thing in the component library view … which I'm not sure if I've added to `vf-component-library` by accident somewhere -- or haven't and need to raise a separate PR. 

I think it's useful here to show cards at 'roughly' what size they'd benefit from being 'in the real world'.

If it's not accidentally added into `vf-component-library` already. I'll do a separate PR.

sturobson: Yeah - I've totally burned the extra code in the components view for this -- another PR incoming (probably Monday now)

khawkins98: If it's for the component docs only I think we should namespace it as such, otherwise devs will probably think it has some connection to the njk template. Something like:

```
 component-library-display:
   width: 440px 
```

components/vf-card/vf-card.scss
```diff
@@ -118,23 +183,70 @@
...
-  .vf-card__title {
+  .vf-card__content {
+    padding-bottom: 0;
+
+    & > *:last-child:not(.vf-card__link) {
+      padding-bottom: 1rem;
+    }
+
+    .vf-list,
+    .vf-content * {
```
khawkins98, line 194: `.vf-content *` hah!

khawkins98 (approved):

components/vf-card/vf-card.config.yml
sturobson: thinking about the cards that don't have images -- they actually work outside of the 440px constraint. So, I've removed this (for now).

PR #1355, opened 2021-02-05
updates vf-section-header svg

sturobson:
- changes the inline width and height of the SVG from `24` to `1em` - this means it will scale with the font-size.
- fixed an alignment issue
- changed some spacing so it will match what is in the new `vf-card` updates

### DIFF
--- a/components/vf-section-header/vf-section-header.njk
+++ b/components/vf-section-header/vf-section-header.njk
@@ -23,7 +23,7 @@
     {{- section_title -}}
 
     {%- if href %}
-      <svg aria-hidden="true" class="vf-section-header__icon | vf-icon vf-icon-arrow--inline-end" width="24" height="24" xmlns="http://www.w3.org/2000/svg"><path d="M0 12c0 6.627 5.373 12 12 12s12-5.373 12-12S18.627 0 12 0C5.376.008.008 5.376 0 12zm13.707-5.209l4.5 4.5a1 1 0 010 1.414l-4.5 4.5a1 1 0 01-1.414-1.414l2.366-2.367a.25.25 0 00-.177-.424H6a1 1 0 010-2h8.482a.25.25 0 00.177-.427l-2.366-2.368a1 1 0 011.414-1.414z" fill="" fill-rule="nonzero"></path></svg>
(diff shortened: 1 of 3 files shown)


components/vf-section-header/vf-section-header.njk
```diff
@@ -23,7 +23,7 @@
     {{- section_title -}}
 
     {%- if href %}
-      <svg aria-hidden="true" class="vf-section-header__icon | vf-icon vf-icon-arrow--inline-end" width="24" height="24" xmlns="http://www.w3.org/2000/svg"><path d="M0 12c0 6.627 5.373 12 12 12s12-5.373 12-12S18.627 0 12 0C5.376.008.008 5.376 0 12zm13.707-5.209l4.5 4.5a1 1 0 010 1.414l-4.5 4.5a1 1 0 01-1.414-1.414l2.366-2.367a.25.25 0 00-.177-.424H6a1 1 0 010-2h8.482a.25.25 0 00.177-.427l-2.366-2.368a1 1 0 011.414-1.414z" fill="" fill-rule="nonzero"></path></svg>
+      <svg aria-hidden="true" class="vf-section-header__icon | vf-icon vf-icon-arrow--inline-end" width="1em" height="1em" xmlns="http://www.w3.org/2000/svg"><path d="M0 12c0 6.627 5.373 12 12 12s12-5.373 12-12S18.627 0 12 0C5.376.008.008 5.376 0 12zm13.707-5.209l4.5 4.5a1 1 0 010 1.414l-4.5 4.5a1 1 0 01-1.414-1.414l2.366-2.367a.25.25 0 00-.177-.424H6a1 1 0 010-2h8.482a.25.25 0 00.177-.427l-2.366-2.368a1 1 0 011.414-1.414z" fill="" fill-rule="nonzero"></path></svg>
```
khawkins98, line 26: Should we do the same in the vf-hero and vf-summary--event-easy arrows?

khawkins98 (approved):

components/vf-section-header/vf-section-header.njk
sturobson: Yeah, I think we should - nice way to scale with the font-size

PR #1357, opened 2021-02-05
Document use of Nunjucks include vs render

### LINKED ISSUE #1356: Documentation: clarify usage of include and render for Nunjucks
We have [a better understanding](https://ebi-comms.slack.com/archives/DBKCMR4E5/p1612535756036400) of when and where to use `include` vs `render` when using Nunjucks templates:

> it’s probably fair to say: server-side use `include`, precompiled browser use `render` 

We need to follow that through by updating the component docs to show the data structure and how/when to use `include` vs `render`. 

I'll also drop this note about how we might optimise the Nunjucks environment handling: we could continue to use fractal for fractal, but ensure the templates are portable so that 11ty doesn't need to bootstrap fractal (11ty would need to also scan the Yaml files which should be doable) .... but this is a separate issue. 

Aslo attaching this mockup @sturobson  made

image

khawkins98:
Closes #1356

Parking this for the weekend, may have more thoughts on Monday

image

### DIFF
--- a/tools/vf-component-library/src/site/_includes/layouts/components.njk
+++ b/tools/vf-component-library/src/site/_includes/layouts/components.njk
@@ -193,9 +193,47 @@ templateEngineOverride: njk
               {# Utility and grid variants have no practical njk or react templates #}
               <details class="vf-details vf-box vf-box--outline">
                 <summary class="vf-details--summary">Nunjucks syntax</summary>
+
+      {% markdown %}
+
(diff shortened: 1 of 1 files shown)


khawkins98:
I think I'm wrapped up with this one now. Ideally the syntax would be shown as Yaml format (or JS as the case may be), however that would require a new helper parser and isn't feasible time wise right now.

image

sturobson (approved):
I thought I'd already approved this one, apologies.

PR #1365, opened 2021-02-09
lets get design decision tokens in documentation

sturobson:
I have:

updated to `vf-design-tokens` to generate a .json file and put the button tokens in a `.tokens.scss` file.
created a helper.js to try and get the `if` statement to generate to the tokens in the `vf-button` readme file

### DIFF
--- a/components/vf-button/README.md
+++ b/components/vf-button/README.md
@@ -32,6 +32,8 @@ The button text should be short and clear. Depending on the action of the button
 
 <sup>*</sup>to be done.
 
+### Sass
+
 ## Install
 
 This component is distributed with npm. After [installing npm](https://www.npmjs.com/get-npm), you can install the `vf-button` with this command.
--- a/components/vf-button/vf-button.tokens.scss
+++ b/components/vf-button/vf-button.tokens.scss
@@ -1,0 +1,19 @@
+// This file has been dynamically generated from design tokens
+// Please do NOT edit directly.
+
+// Source: components/vf-button.yml
+
+$vf-button-map: (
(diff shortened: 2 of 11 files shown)


khawkins98:
I think the helper.js is missing from this branch? Presumably that's the `helpers.getTokenFile`?

Also, it looks like you're tying to access a data object at `components.button.properties`, but I don't think `components` is defined in this template's scope.

May just be that some of the technical plumbing is missing from this commit?

sturobson:
the `components.` bit is needed because I've nested the JS (which I'll upload) into `_data/components/buttons.js' so it's needed. Like we do with the tokens

sturobson:
What I want is using:
```
const path = require('path');

module.exports = {
  getTokenFile(text) {

    return "components." + path.basename(__filename) + ".properties";
  }
};
```

but the `path.basename(__filename)` to be the component's `title` in the `README.md`. Doing this with nunjucks alone doesn't seem to work although I can get it to print out the correct thing.

khawkins98:
The property name wasn't being interpreted as a variable, this fixes it https://github.com/visual-framework/vf-core/pull/1365/commits/2c62006a5bb694d7700049ea7d7618a7e312853d

I'll probably have a couple other comments on the build logic, but need to spend more time looking at "the big picture"

sturobson:
> need to spend more time looking at "the big picture"

I should write the thoughts going on in my head - which might help

sturobson:
As much as I think this will be a good thing moving forward and helpful to assist teams that want to use the VF but bring their own styles this should be closed here and no further development work made on it in lieu of time I have left on the project.

PR #1366, opened 2021-02-09
fixes a margin issue

sturobson:
There are several instances of `vf-content` being used where the first item is not an h1 … this leads to added spacing with margin to the 'box' that the content is in. 

This wholesale removes the top margin from the direct first child of `vf-content`.


khawkins98 (approved):

khawkins98:
As vf-wp has very liberal application of vf-content around all gutenberg blocks, this may cause unexpected top margins disappear, but that's probably more of an issue with vf-wp.

PR #1367, opened 2021-02-11
removes all variants

sturobson:


khawkins98:
all of it in the bin

khawkins98:
Looks good, just one thing: we've lost an example that shows usage of `vf-hero__heading--additional`

khawkins98:
It looks like `vf-hero__heading--additional` is now `vf-hero__heading--kicker`, worth a nunjucks alias so its less of a breaking change, particularly for vf-wp?

```
{% if context.vf_hero_heading_additional %}        
  {% set vf_hero_kicker = context.vf_hero_heading_additional %}
```

sturobson:
> worth a nunjucks alias so its less of a breaking change, particularly for vf-wp?

We can do this -- I was going for a `v3.0.0` release because of the varied breaking changes -- but I guess this addition (heh) will help make a smoother move

sturobson:
@khawkins98 I think this one is ready for another look over now

khawkins98 (approved):

khawkins98:
Looking good. thank for the tweaks.

PR #1375, opened 2021-02-16
BUG: vf-card issue with image height in Safari

sturobson:
This fixes an issue where we had used `height: 100%` on the image height. Safari does not respect any form of aspect ratio and simply stretches the image so it's the full height of it's container.


khawkins98 (approved):

sturobson:
This would also help fix an issue when printing things out too.

PR #1376, opened 2021-02-16
updates blockquote to remove pullquote

### LINKED ISSUE #1374: VF component list - bug in navigation on blockquote
Hi, Ken and Stu

When on this page, the navigation on the left becomes unclickable.

https://stable.visual-framework.dev/components/vf-blockquote/

Thanks
Gabby

sturobson:
I cannot recall why there was a `--pullquote` variant to this, and looking at the HTML and CSS for it - it has never been properly implemented.

I've removed it.

This will close #1374


khawkins98:
Sounds good. We can revisit the pull quote if/when we have a need and full designs.

khawkins98 (approved):

PR #1385, opened 2021-02-22
the one that tidies up the general `vf-link` styles.

sturobson:
After discussing things with @cindyebi I have:

- removed 'primary' and 'secondary' as an option as these are not really used at all because the 'secondary' use (components) is usually defined in the component CSS.
- re-jigged how the `disabled` style works.
- improved the examples a little

in doing the above:

- removed a call for `vf-link.variables.scss` in `vf-summry`
- updated the design tokens to match


khawkins98:
Spotted one thing, vf-content makes use of `$vf-link--dark-mode--hover-color`. I don't think that CSS has ever been used in production?

https://github.com/visual-framework/vf-core/blob/develop/components/vf-content/vf-content.scss#L234-L242

sturobson:
Yeah - we've not really made any in-roads on a potential 'dark mode'. So I thin we can remove this here too.

khawkins98 (approved):

PR #1392, opened 2021-02-25
the ones that resolves fractal wanting to render when using render in readmes

sturobson:


components/vf-card/vf-card.config.yml
```diff
@@ -9,7 +9,7 @@ status: alpha
 context:
   component-type: block
 
-  # modifier: vf-u-background-color--grey--lightest
+  -m :# modifier: vf-u-background-color--grey--lightest
```
khawkins98: Typo?

sturobson: indeed

khawkins98 (approved):

khawkins98 (approved):

PR #1405, opened 2021-03-04
The one that fixes any stylelint errors

sturobson:
Following @nitinja's update to using Husky this PR fixes any of the issues that arose from the gulp tasks running. 

I've made each component fix it's own commit for some history.

We'd need to do a few patch releases to npm to get this in at some point.

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -8,7 +8,7 @@
   ],
   "rules": {
     "string-quotes": "single",
-    "color-named": "never",
+    "color-named": ["never", { ignore: ["inside-function"] }],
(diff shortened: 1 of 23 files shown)


.stylelintrc
```diff
@@ -8,7 +8,7 @@
   ],
   "rules": {
     "string-quotes": "single",
-    "color-named": "never",
+    "color-named": ["never", { ignore: ["inside-function"] }],
```
khawkins98, line 11: magic

khawkins98 (approved):

PR #1406, opened 2021-03-04
the one that properly deprecates the masthead

sturobson:
Looks like we cannot deprecate the component in npm so this is the way suggesting --

https://npm.community/t/npm-deprecate-giving-error-e422-unprocessable-entity/2337


khawkins98:
I was thinking about this a bit last night and worry that it's too aggressive of a deprecation, as the vf-masthead CSS will disappear for anyone using the CDN release. 

Thought's on keeping the CSS but still:

1. Hiding the component
2. Keeping the readme
3. Wrapping in `html:not(.vf-disable-deprecated) {`
    - to follow: https://stable.visual-framework.dev//developing/components/deprecating-components/

sturobson:
I think it needs to be aggressive enventually :D 

Perhaps:

`3.0.0` do as you suggest
`4.0.0` remove it completely as per this PR

As you know, I'm not a fan of the CDN usage :D

sturobson:
updated and should be ready to go now

sturobson:
@khawkins98 ** coughs ** in case this one slipped you by...

sturobson:
@khawkins98 I think this resolves everything we've discussed and is _finally_ ready now.

khawkins98 (approved):

khawkins98:
excellent

PR #1390, opened 2021-02-24
The one that makes lots of really good changes to forms.

sturobson:
This takes up quite a lot of the recent agreements in changes found at #1229.

This also pulls out the forms in to their own section for the visual-framework.dev site too. 

todo: 

check the html order of input and labels for checkbox and radio with a screen reader and fix as required.
I might have missed a changelog, or we might need to change major/minor/patch accordingly. 

This will close #1229 and will close #820

PR #1408, opened 2021-03-04
the one that adds autocomplete variant of search input.

sturobson:
notes: 

This is using the 'current' styling of forms but will be updated when #1390 closes (as I'll add this there too).

This variant is **only** to be used when there is some sort of autocomplete / typeahead filtering needed when searching.

This is **not** to be used in place of the normal `input` and `button` because it's more 'visually appealing'.

### DIFF
--- a/components/vf-form/vf-form__input/CHANGELOG.md
+++ b/components/vf-form/vf-form__input/CHANGELOG.md
@@ -1,3 +1,6 @@
+### 2.1.0
+
+* adds an search input for when the search field is interactive showing 'live results'.g
 ### 2.0.1
 
 * changes any `set-` style functions to cleaner version
--- a/components/vf-form/vf-form__input/vf-form__input.njk
+++ b/components/vf-form/vf-form__input/vf-form__input.njk
@@ -15,7 +15,7 @@
   </div>
    <div class="vf-form__item vf-stack">
     <label class="vf-form__label | vf-u-sr-only" for="search">Search</label>
-    <input type="search" id="search" class="vf-form__input vf-form__input--autocomplete" placeholder="Search this part of the website">
+    <input type="search" id="search" class="vf-form__input vf-form__input--filter" placeholder="Search this part of the website">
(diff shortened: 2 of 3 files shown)


sturobson:
as mentioned in the opening comment - this has the current styling for forms:

normal:
Screenshot 2021-03-04 at 14 38 38
hovered/focused:
Screenshot 2021-03-04 at 14 38 56

khawkins98:
Would this be better/clearer as part of vf-search? 

https://stable.visual-framework.dev/components/vf-search/

khawkins98:
Another comment: I expected some sort of javascript-y thing with the name "autocomplete", would this be better called "compact" or "filter" or something like that?

sturobson:
hmm, it might do well to be part of `vf-search` too. I'd like to keep the form 'element' on it's own. As it could easily become something that's inside a 'filter section' in results page etc over time.

sturobson:
… this also reminds me of `vf-search` and that I can amend this to match the styles discussed for the 'search container'

sturobson:
> Another comment: I expected some sort of javascript-y thing with the name "autocomplete", would this be better called "compact" or "filter" or something like that?

I was toying with 'typeahead' but that is a specific JS solution for this - so ended up on autocomplete, but then there is an HTML autocomplete attribute.

Perhaps `filter` is better here.

sturobson:
I've changed `autocomplete` to `filter` in the codes now

khawkins98:
Would be good to add your notes:

> This variant is only to be used when there is some sort of autocomplete / typeahead filtering needed when searching.
>
> This is not to be used in place of the normal input and button because it's more 'visually appealing'.

Either in the NJKs example or to the Readme?

sturobson:
I think we might have to revisit this once the #1390 PR is merged in - as I'm editing files that won't exist :sob:

sturobson:
This is all wrapped up inside #1390 now so we can close this PR

PR #1411, opened 2021-03-05
the one that makes the inline search variant better when it has less space

sturobson:
We have a problem when the `vf-search--inline` is on smaller viewports or has a small amount of the page space to fill.

The button drops underneath but it's still a bit messy.

This PR fixes that by:

- introducing the CSS that will eventually become it's own layout component using [this great resource](https://every-layout.dev/layouts/sidebar/)
- adds a wrapping `div` around the `vf-form` content.
- replaces all the existing CSS for the `vf-search--inline` variant.

This, I feel, can be considered a 'short' or 'medium' term solution as we still need to create a better 'search container' now that we have some decisions made on what it should look like etc. etc.

I think, also, that the `--inline` variant can be the default version for search on embl.org properties with the merge now too. 

On small viewports where the search input cannot take up 60% of the space because of the button:

Screen Shot 2021-03-06 at 14 02 46

On larger viewports where there is more space:

Screen Shot 2021-03-06 at 14 02 37

### DIFF
--- a/components/vf-search/CHANGELOG.md
+++ b/components/vf-search/CHANGELOG.md
@@ -1,3 +1,8 @@
+### 2.0.0
+
+* adds a div around the form content.
+* makes changes to the CSS for `--inline` variant so it's responsive to the parent width.
+
 ### 1.2.0
 
 * don't allow inline search to get too large
--- a/components/vf-search/README.md
+++ b/components/vf-search/README.md
@@ -4,9 +4,13 @@
 
 ## About
 
+
+
 ## Usage
 
-- You can enable `autofocus` on the search element, but should only do so if most users intend to search on the page
+**note:** Version `2.0.0` of the `vf-search--inline` has no maximum width and will fill the space of it's parent. Because of this it is recommended to make sure the component is not too wide by wrapping it in the `embl-grid` with the `--centered-content` variant. If you wish to use ths `vf-search--inline` with `vf-grid` you will need to make use the `vf-search--inline` component also has an appropraite `.vf-u-grid__col--span` class.
+
+- You can enable `autofocus` on the search element, but should only do so if most users intend to search on the page.
 
 ## Install
 
--- a/components/vf-search/vf-search.config.yml
+++ b/components/vf-search/vf-search.config.yml
@@ -1,12 +1,10 @@
-# The title shown on the component page
 title: Search
-# Label shown on index pages
 label: Search
 status: beta
-collated: true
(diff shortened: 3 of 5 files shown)


khawkins98:
Looks good. One thing we've lost though is the max-width of the search box, which from previous discussions is non-helpful if it can grow wider than an optimal line-length:

image

I think previously I had limited it to a relatively arbitrary 600px (or so), but given our discussion about line lengths, would be best if we limited it to `60ch` or `72ch`. 

This:

1) Ensure the search button is never too far from the  typing
2) Would encourage long search queries (example: searching for a protein sequence) to use a `textarea` instead

sturobson:
Thinking in containers -- this would be where `vf-search` would need to be 'inside' something like `embl-grid--has--centered-content` or something. 

Which I think it does already in most places? 

```
<section class="vf-container vf-container--search | embl-grid embl-grid--has-centered-content | vf-u-fullbleed">
  <div><!-- put a section header if you want --></div>
  <div><!-- put the inline search variant here --></div> 
</section>
```

So, this is the first PR so that I can make the next PR in creating a 'design pattern' container for it, I guess. Rather than crossing the streams and making the PR more than it needs to be.

don't cross the streams

khawkins98:
> Which I think it does already in most places?

I'm mixed on it. I agree that it should be appropriately limited by the parent container, but I've already seen many examples where it's just been dropped in as the only child of vf-grid or as an embl-grid without a sidebar (which is why it was limited originally).

If we do go without a limit we should make a note in the usage notes that it should be inside a middle embl-grid column as to not exceed XXch.

sturobson:
> If we do go without a limit we should make a note in the usage notes that it should be inside a middle embl-grid column as to not exceed XXch.

👍

sturobson:
added:


**note:** Version `2.0.0` of the `vf-search--inline` has no maximum width and will fill the space of it's parent. Because of this it is recommended to make sure the component is not too wide by wrapping it in the `embl-grid` with the `--centered-content` variant. If you wish to use ths `vf-search--inline` with `vf-grid` you will need to make use the `vf-search--inline` component also has an appropraite `.vf-u-grid__col--span` class.

khawkins98 (approved):

khawkins98:
Looks good, thanks for the tweak.

khawkins98:
and +1 for:

> So, this is the first PR so that I can make the next PR in creating a 'design pattern' container for it, I guess. Rather than crossing the streams and making the PR more than it needs to be.

PR #1416, opened 2021-03-09
Added react button component

nitinja:
My First component - react button.
Any comments are welcome.


khawkins98:
Hey Nitin, this is looking good. 

Two requests:

- could you update the changelog's for vf-extensions-react and vf-button
- could you add a usage example of vf-button?

nitinja:
* Updated changelog and readme files
* Updated react component to add missing outline option.
* Added missing reference to sass utility mixin that's required by utility classes
I will be raising separate PR for react documentation / component showcase.
@khawkins98 can you please give me write access to https://github.com/visual-framework/vf-react.

components/vf-button/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "1.4.1",
+  "version": "1.4.2",
```
khawkins98: No need to bump the version here. When we do our release, `lerna` will automatically update this.

tools/vf-extensions-react/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+### 0.0.4
```
khawkins98: I don't think you mean to update the `vf-extensions-react` changelog and package.json in this PR?

khawkins98:
Looking good. Added a couple small notes. 

Also gave you access to vf-react.

tools/vf-extensions-react/CHANGELOG.md
nitinja: No I did it on purpose. If I have to add changelog, I will have to update version as well, right? Otherwise, under what version to put changelog? Or we just update changelog with fictional future version?

khawkins98: there's two things here:

- the `vf-extensions-react` changelog.md and package.json have been updated, but there aren't any changes to that component
- and then in `vf-button`, yes, no need to update the package.json. lerna will update that. You're preparing the release notes with than anticipated new version, but we can update the version if needed when publishing — I think it will become clearer the first time you see the release process

nitinja: ok, I wont update files in `vf-extensions-react` (but your comment "could you update the changelog's for vf-extensions-react and vf-button" remains unresolved. was that a mistake?).

khawkins98 (approved):

tools/vf-extensions-react/CHANGELOG.md
khawkins98: Thanks. Have just resolved/approved.

PR #1420, opened 2021-03-16
the one that adds more vf-stacks

sturobson:
I'm surprised I didn't add these from the start.

We add `vf-stack` variants for all existing spacing design tokens.

### DIFF
--- a/components/vf-stack/vf-stack--500.njk
+++ b/components/vf-stack/vf-stack--500.njk
@@ -1,0 +1,18 @@
+{% if context %}
+  {% set stack__spacing = context.stack__spacing %}
+  {% set stack__modifier = context.stack__modifier %}
+  {% set stack__spacing__custom = context.stack__spacing__custom %}
+  {% set custom_spacing_property = context.custom_spacing_property %}
+{% endif %}
+
+<div class="vf-stack
+  {%- if stack__spacing %} vf-stack--{{stack__spacing}}{%- endif -%}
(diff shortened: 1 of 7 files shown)


components/vf-stack/vf-stack--500.njk
```diff
@@ -0,0 +1,18 @@
+{% if context %}
```
khawkins98, line 1: Shouldn't one vf-stack.njk template be enough, or is there something I'm missing?

khawkins98 (approved):

components/vf-stack/vf-stack--500.njk
sturobson: It's in the hope that people using `njk` in their sites make use of nunjucks' `block` to save people writing more HTML (that could well change) than they have to.

khawkins98: I'd think one template should still be enough? There's probably some technical thing with `block` that I'm  missing, we can chat about it sometime. Certainly not a reason to hold back the PR.

sturobson: I'll try and remember why I had to go this way - rather than rely on using blocks in fractal -- I think it was because of their use of `@` to call things -- which didn't work in 11ty -- but would do if we kept a list of key/value pairs like we do with includes.

sturobson: I'll merge this in as it works 'as is' - and look to check how the alternative way works with fractal and 11ty

PR #1436, opened 2021-03-22
The One That Adds Neutral Colour Tokens

sturobson:
This PR:

- adds the neutral colour tint/shade stack that incorporates the greys from the VF.
- doing this replaces the `grey--lightest` tint as it looked awful.
- this deprecates colours that are now in the neutral stack.
- updates the documentation site.
- this does not get rid of the old way of calling colours, so not an entire breaking change.

- [x] issue with Markdown not working.


khawkins98:
Got this working by https://github.com/visual-framework/vf-core/pull/1436/commits/3fc122bb2864fedadf6045dd5b72b1e29beb216c

> vf-component-library is only designed to work with vf-core, so we don't need to rely on an npm install of vf-design-tokens.

sturobson:
@khawkins98 this one should be ready for a review and merge :)

khawkins98 (approved):

sturobson:
Realised that really these should still follow the `VF-color__type—variant` pattern so I’ll do a quick change in the morning and a fresh PR

PR #1440, opened 2021-03-23
The one that introduces a flexible, intrinsic sidebar layout component

sturobson:
A layout component that changes layout from inline to block depending on the width available of where it is put.

Can be used for the `vf-search` to make the button jump under if the viewport is small - or if the space available in a larger grid layout means that the main content (like the input) cannot take up the `min-`width` that has been set.

Probably better seen than trying to write about how it works.

The genus of the code is taken from here - https://every-layout.dev/layouts/sidebar/ - with CSS custom properties sprinkled and the component 'nunjucks-ified'

todo:

- [ ] make the mark up work like we have with `vf-stack` so that it can be used for `{{ block }}` content.


khawkins98:
Ah, made a couple of hotfixes on the boilerplate stuff and then realised this is still a draft.

sturobson:
I've noticed that if you've got the beta 'dark dimmed' theme in Github - the green and grey are quite close in colour.

khawkins98:
A couple things to suggest adding to the readme:

- How should this (not) be used to replace embl-grid.
- Can/should I combine this with vf-grid?
- Is giving all the spacing options opening too many choices to the user? Should we just use the global grid-gap?

I'm guessing the use case is things like jobs and other "sidebary" things where embl-grid has been hijacked?

sturobson:
> How should this (not) be used to replace embl-grid.

It's more for component layouts rather than grid layouts. So the search form is a good example, it could also be used in summarys too (potentially). It's like a `vf-flag` on steroids. Which could neuter the `vf-flag` but I think that'll still have it's place (like in a list of flags 😆)

> Can/should I combine this with vf-grid?

It's it's own layout (using `display: flex;` so adding something else to it will break it I reckon

> Is giving all the spacing options opening too many choices to the user? Should we just use the global grid-gap?

I like this

sturobson:
> Is giving all the spacing options opening too many choices to the user? Should we just use the global grid-gap?

Thinking on this and try it out…

As the `vf-sidebar` is independent from it's parent layout I don't think using the `--page-grid-gap` custom property would be that useful.

Perhaps we remove some of the options … 

`200`, or `.5rem` is probably too small to ever use.
`600`, or `.1.25rem` is an oddity that does have uses in the system, but probably not here
`1200`, or `3rem` is probably too big a space to use.
`1600`, or `4rem` is probably too big a space to use.

I think reducing the options to `400`(`1rem`), `600`(`1.5rem`), and `800`(`2rem`) is more than enough.

We could add `--vf-sidebar--custom-spacing`  but I think that might be could lead to inconsistency across the broader page layout (visually) and make it look 'odd'.

sturobson:
The above reduction in spacing could also be applied to the `vf-flag` layout too I think.

khawkins98:
Looks good. Maybe just one last request/idea: should we add a big "this is an early alpha component and subject to drastic change and refactor"?

khawkins98 (approved):

sturobson:
> should we add a big "this is an early alpha component and subject to drastic change and refactor"?

YOLO

Perhaps we should, but the package version number should make that apparent.

I'll add something :D

khawkins98:
Yeah, and we should use this line to show big scary warnings:

`status: alpha`

https://github.com/visual-framework/vf-core/pull/1440/files#diff-3f2393a4f36e94aaf4b3dd4c291ce6635c2bfc9a965c081584fa5eafb6773913R3

PR #1449, opened 2021-03-25
The one that fixes issues with browsers and forms being required

### LINKED ISSUE #1345: In search form, making the input mandatory turns borders red
Screenshot 2021-02-03 at 11 02 28

It adds the following class to the input element which sets the border color to red even without interacting with the search form.
```
.vf-form__input:invalid {
    border-color: #d32f2f;
}

```
Is it the expected behavior or borders should stay in default color until user submits the form without giving any input!

sturobson:
This will close https://github.com/visual-framework/vf-core/issues/1345


khawkins98:
Worth adding this as an example in the njk file?

khawkins98 (approved):

Issue #1452, opened 2021-03-26
Bug: vf-content requires non-existant version of vf-form


khawkins98:
Somehow vf-content wound up requiring `@visual-framework/form@^2.0.0` when an alpha of 2.0.0 is the latest.

https://github.com/visual-framework/vf-wp/pull/747/checks?check_run_id=2199452835

```
+ gulp@4.0.2
added 323 packages from 220 contributors in 11.382s
yarn install v1.22.10
[1/4] Resolving packages...
error Couldn't find package "@visual-framework/form@^2.0.0" required by "@visual-framework/vf-content@1.5.0" on the "npm" registry.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
Error: The operation was canceled.
```

This is causing updates to vf-content to fail.

sturobson:
looks like Github doesn't recognise anything after the first 'closes' in a PR. Closing this 'the old way' - by pressing a button.

PR #1454, opened 2021-03-26
Bug: vf-form version and vf-masthead js function

### LINKED ISSUE #1453: Bug: vfMasthead() function failing
In #1406 we removed the vfMasthead() JS but didn't remove the function call in vfComponentRollup.js

Suggest:

1. We remove the JS call from vfComponetnRollup.js
2. We add a dummy JS function in vf-masthead.js that says "This component has been deprecated" 

The current situation is causing JS to error (visible on https://latest.visual-framework.dev/)

khawkins98:
Closes #1453 and #1452

### DIFF
--- a/components/vf-componenet-rollup/CHANGELOG.md
+++ b/components/vf-componenet-rollup/CHANGELOG.md
@@ -1,6 +1,7 @@
 ### 1.3.1
 
-* adds deprecation comment for `vf-masthead`.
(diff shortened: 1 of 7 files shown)


khawkins98:
We'll still need to do an npm release to fully fix these. Perhaps today or Monday.

sturobson (approved):

PR #1455, opened 2021-03-26
The one that reworks vf-search, again

sturobson:
Based on latest discussions and decisions made on vf-search.

Introduces the 'container' version.

Adds more docs. 

This PR

- hides the `default` 
- exposes a `responsive` version
- introduces a `mini` version (not to be implemented)
- a version showing the search container.

The `responsive` version is essentially the 'default' search form label/input/button combo. 

When the viewport is below 600px the button content changes from text to an icon.

Screen Shot 2021-03-27 at 14 57 43

Screen Shot 2021-03-27 at 14 57 53

It also introduces the suggested HTML for 'auto complete' results

Screenshot 2021-03-26 at 15 01 44


sturobson:
@khawkins98 I’ve just had a look at the preview - https://deploy-preview-1455--ecstatic-noether-150ab4.netlify.app/components/ - and I think the CSS is having the same issue we had with the `latest` subdomain? As it looks (from 10,000 feet) that it’s only got the `stable` CSS and not what it should be from the changes in the PR?

Aside: It’d be nice to see if there’s a ‘netlify preview bot’ that spits the link as a comment … this might be the netlify settings - https://answers.netlify.com/t/netlify-bot-to-post-preview-urls-in-comments-of-github-pr/6254

khawkins98:
The search icon in the button makes me uneasy, but I can accept the logic as to why it's wound up like this.

sturobson:
> The search icon in the button makes me uneasy

Likewise - but we have found precedent of this on other large-scale websites that change from 'word' to 'icon' depending on viewport. 

It kinda gets nullified on phone/tablet because the keyboard layout changes to include a 'search' button - but hiding the button completely wouldn't be a good option as viewport size != mobile/tablet.

components/vf-search/vf-search.config.yml
```diff
@@ -1,34 +1,89 @@
...
+      search_placeholder: Enter your search terms
+      search_id: searchitem
+      search_value_default: ''
+
+  - name: container
+    label: Search Container
     context:
-      variant: default
+      responsive: true
+      search__background: "#F3F3F3"
```
khawkins98: Would it be better to pass in one of the css custom props that has a design token value? 

I think this would work?

```
      search__background: "var(--vf-ui-color--grey--light)"
```

sturobson: Ah, yes - that might be better. I'll take a look and update.

khawkins98:
> Aside: It’d be nice to see if there’s a ‘netlify preview bot’ that spits the link as a comment … this might be the netlify settings - https://answers.netlify.com/t/netlify-bot-to-post-preview-urls-in-comments-of-github-pr/6254

I've updated the netlify settings, we see if we get lucky. 

https://app.netlify.com/sites/ecstatic-noether-150ab4/settings/deploys

> I’ve just had a look at the preview - https://deploy-preview-1455--ecstatic-noether-150ab4.netlify.app/components/ - and I think the CSS is having the same issue we had with the latest subdomain? As it looks (from 10,000 feet) that it’s only got the stable CSS and not what it should be from the changes in the PR?

Ultimately we need a good way to pass the environment that the build is running in (netlfiy, vs head of develop, vs tag), I've yet to see a good way to do this. Maybe it's something Nitin could help us on.

sturobson:
In the `.yml` file I'ave changed `#F3F3F3` to `var(--vf-color--neutral--100`. We need to include the `var()` here and I don't want to add it into the inline HTML style as we cannot guarantee other teams that need to style things a little differently will use all the custom properties available.

khawkins98 (approved):

PR #1460, opened 2021-03-30
Bug: vf-sass-config neutral lookup function

khawkins98:
Fix key name in `vf-color--neutral` lookup.

This was causing sass like `background-color: neutral(0);` to return a null value.


sturobson (approved):
spotted this yesterday -- thanks for the quick fix.

PR #1461, opened 2021-03-30
status updates pre-release

sturobson:
updated lots of component status to match their current maturity.


khawkins98 (approved):

khawkins98:
Nice, thanks.

Thought for "future": would be nice if the status (if not defined) would fallback to the npm package version (x.x.x.-alpha.x = alpha)

PR #1462, opened 2021-03-30
Bug: vf-hero loop

khawkins98:
Found while helping update the static-html-pages: the current vf-hero.njk will error if receive a `vf_hero_text` from a yaml file under certain contexts.

This:
- fixes the above
- Allows vf_hero_heading_href to be passed in yaml
- Gobbles much of the whitespace

FYI @kasprzyk-sz


sturobson (approved):

sturobson:
I like whitespace 😆

PR #1464, opened 2021-03-31
the one that updates the navigation component. 

sturobson:
* deprecated the `--aditional` variant.
* removes `--main` and `--global` Sass files, incoroprating them into the base stylesheet.
* tidies up quite a bit of CSS.
* makes use of `vf-cluster` for layout in the markup.
* adds some documentation.

Known things that will need updating to make use of this:

- the use of `vf-navigation--global` in the `vf-global-header` (part of the content hub?)
- `vf-wp` and `embl-static-html-pages` if not done so already.
- any other product using `vf-navigation` that makes use of the latest CSS.


sturobson:
note: 

- [x] looks like some nunjucks got removed by accident.
- [x] I need to add some `set` things to make it more portable for 11ty

khawkins98:
With updates to the contentHub template, the worry is always "what about sites that haven't updated?"

But as the new class set in the njk is "vf-navigation vf-navigation--global | vf-cluster", it seems like this should be fine. (older sites will still be able to hook into "vf-navigation--global". 

Sites with the newer vf-navigation component will simply need to ensure they also have vf-cluster. 

Suggest we release this as 3.0.0-beta.0 as there may be dragons.

khawkins98 (approved):

PR #1465, opened 2021-03-31
the one that deprecates the phase variant of the banner

sturobson:
The `vf-banner--phase` was superseded by a more generic `vf-banner--notice` variant a while back but some things didn't get 'sorted':

- there was no deprecation notice added to `--phase`.
- the `vf-banner--phase` variant is still widely used.

This PR fixes the deprecation of the variant by adding all the relevant codes.

### DIFF
--- a/components/vf-banner/vf-banner--inline.njk
+++ b/components/vf-banner/vf-banner--inline.njk
@@ -1,3 +1,5 @@
+<h2>This component has been <span style="color: rgb(228, 0, 70);">deprecated</span>. Please use the <a class="vf-link" href="./vf-banner--info/"new</a> component.</h2> in the README.md file.
+
 <div class="vf-banner vf-banner--phase">
   <div class="vf-banner__content">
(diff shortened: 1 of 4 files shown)


components/vf-banner/vf-banner--inline.njk
```diff
@@ -1,3 +1,5 @@
+<h2>This component has been <span style="color: rgb(228, 0, 70);">deprecated</span>. Please use the <a class="vf-link" href="./vf-banner--info/"new</a> component.</h2>
```
khawkins98: Better to wrap this in `<!-- -->` as this could silently blow up a production site?

sturobson: Ah!

khawkins98: Noticed the href was broken, just fixed that.

khawkins98 (approved):

PR #1475, opened 2021-04-07
bug: vf-sass-config utility mixins import

khawkins98:
`mixins/vf-utility-mixins.scss` was missing from the rollup import, which breaks use in some projects (like vf-react that use the React sass build)


sturobson (approved):

sturobson:
D'oh! 

I won't talk about how I think generating utility classes from all the tokens is now a bad idea. Not now.

PR #1476, opened 2021-04-07
Enhancement: more flexible module import

khawkins98:
In React (and I imagine other places) they're more particular about the location of the JS module, by adding a `try/catch` we can improve the ability to use VF JS in places.

This idea won't work how I've currently done it. parking it here for now.


sturobson (approved):

khawkins98:
I've simplified this to use a relative path. The vf-js build is smart and doesn't care, and for less smart builds (like Create React App) the relative path is very helpful.

PR #1477, opened 2021-04-08
Bug: garbled version number of vf-navigation

khawkins98:
Will trigger a re-release of 3.0.0-beta.0 to fix garbled version number.

### DIFF
--- a/components/vf-navigation/package.json
+++ b/components/vf-navigation/package.json
@@ -1,5 +1,5 @@
 {
-  "version": "2.2.3-3.0.0-beta.0.0",
+  "version": "3.0.0-beta.0",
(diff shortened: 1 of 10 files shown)


components/vf-navigation/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "2.2.3-3.0.0-beta.0.0",
+  "version": "3.0.0-beta.0",
```
sturobson: should this be `beta.1` to match the changelog?

khawkins98: ah, lerna will bump it to .1

sturobson (approved):

PR #1485, opened 2021-04-13
enhancement: vf-hero safe text

khawkins98:
Allows html (for links) in vf-hero__subheading (and a few other fields) when using the Nunjucks template.


sturobson:
hmm, links in a sub heading?

khawkins98:
As they're often multiple sentences, I'd say they're closer to an "intro" than a "subheading" in practice, but that'd be a more complex change. 

image

sturobson (approved):

PR #1486, opened 2021-04-13
the one that updates and deprecates bits of the table

sturobson:
- makes the `--striped` variant the default styling.
- deprecates `--striped`, `--bordered`, `--tight`, `--loose` variants.
- tidies up CSS ordering of rulesets to match more the component.
- makes the `--sortable` icons on by default before a bigger refactor.

### DIFF
--- a/components/vf-table/CHANGELOG.md
+++ b/components/vf-table/CHANGELOG.md
@@ -1,3 +1,10 @@
+### 1.2.0
+
+* makes the `--striped` variant the default styling.
+* deprecates `--striped`, `--bordered`, `--tight`, `--loose` variants.
+* tidies up CSS ordering of rulesets to match more the component.
+* makes the `--sortable` icons on by default before a bigger refactor.
+
 ### 1.1.2
 
 * changes any `set-` style functions to cleaner version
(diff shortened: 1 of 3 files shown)


khawkins98:
Couple thoughts:

- Should we call this a `1.2.0-rc.0` as I feel like more will come soon?
- Is there an option to turn off striping? I know from past experiences it can be problematic in some cases of embedded "sparkline" figures

khawkins98:
Cross-linking #1474

sturobson:
is there an option to turn off striping? I know from past experiences it can be problematic in some cases of embedded "sparkline" figures

not in this PR -- and I did see `--no-stripe` or sometihng similar in the VF legacy code. 

I guess as the intention is for striped to be the good default we could adopt this type of classname and CSS

although - it does make me think we're adding CSS to remove CSS.

🤔

sturobson:
@khawkins98 I've just pushed a few more updates, taking in the feedback.

This is more the 'first pass' of reducing everything before any bigger changes (responsiveness etc) gets into play.

khawkins98 (approved):

Issue #1443, opened 2021-03-23
DOCUMENTATION: Add an accessibility statement


sturobson:
After the latest VF/UX chat I think we decided on vying for AAA where possible on new components and making sure existing things are A or AA compliant.

We should probably add some form of ‘statement’ to the docs site. 

I found this example: https://www.gov.uk/government/publications/sample-accessibility-statement/sample-accessibility-statement-for-a-fictional-public-sector-website

khawkins98:
W3C's WAI has some background and a generator that could be used:

- https://www.w3.org/WAI/planning/statements/
- https://www.w3.org/WAI/planning/statements/generator/#create

There's also a minimal example which may be good place to begin https://www.w3.org/WAI/planning/statements/minimal-example/

Also as discussed, would be good to add an "Accessibility" principle at https://latest.visual-framework.dev/guidance/, and from there we could describe a process by which sites should create their accessibility statement.

khawkins98:
Thinking about an accessibility statement, I wonder if it would be better for this to live as part of the EMBL.org website (much like we do for the privacy statement).

khawkins98:
Leaving this link here: https://latest.visual-framework.dev/guidance/accessibility/

PR #1488, opened 2021-04-13
Update accessibility.md

nikiforosk:
Replaced "impalement" (which I think is a method of torture) with "impairment".


khawkins98:
ah, very good 🩸

khawkins98 (approved):

khawkins98:
Cross-linking #1443

Issue #1175, opened 2020-10-20
Discussion/notes: improving the component library browser


khawkins98:
The component library remains an ongoing work in progress.

One of the big outstanding things that has been discussed a few times is how to handle "in context" component navigation ... that is, the left column:

![image](https://user-images.githubusercontent.com/928100/96611005-878aa100-12fc-11eb-9f1b-97f8049e2413.png)

## The problem

The left column works well from a UX approach, but as a simple grid it can disrupt the preview of some things that are intended to be used as full width.

[vf-u-fullbleed](https://visual-framework.github.io/vf-core/components/vf-u-fullbleed/)

![image](https://user-images.githubusercontent.com/928100/96611114-aab55080-12fc-11eb-91b2-7fb62b7061b7.png)

And even less obvious ones like [vf-intro](https://visual-framework.github.io/vf-core/components/embl-grid), where they're intended to be used in a full-width container:

![image](https://user-images.githubusercontent.com/928100/96611755-4b0b7500-12fd-11eb-8d8b-0a038e580a74.png)

---

## Soooo... 

how do other component libraries/systems handle component browsers?

### Orbit

Uses a tree-like left sidebar. It avoids layout breaks by not showing previews for these components on the page and instead of linking to an external page for previews.

https://orbit.kiwi/components/navigationbar/

![image](https://user-images.githubusercontent.com/928100/96612989-ba359900-12fe-11eb-8911-ad83a52d73b1.png)

### Patternfly

Uses a left-side tree and — where a component can't be shown in-context — shows a screenshot that opens in a new tab.

https://www.patternfly.org/v4/components/login-page

![image](https://user-images.githubusercontent.com/928100/96613472-4ba50b00-12ff-11eb-8fba-447383357ef7.png)

### USWDS

Also left side and uses a screenshot (like Patternfly)

https://designsystem.digital.gov/page-templates/#documentation-page

![image](https://user-images.githubusercontent.com/928100/96613670-873fd500-12ff-11eb-968b-59a8469aa556.png)

### Duet 

Uses resizeable iframes and requires that you make your browser sufficiently large to show "desktop in a desktop"

https://www.duetds.com/templates/hero-area-light/

![image](https://user-images.githubusercontent.com/928100/96613935-d6860580-12ff-11eb-8328-e869d980dee0.png)

### More?

I'm sure I've overlooked some.

---

## Do we already have a partial solution?

It's conceivable to adapt `vf-tree`

[...]

PR #1489, opened 2021-04-14
enhancement: component structure

khawkins98:
After recent conversations, it's become clear that:

1. examples/boilerplates are overlooked in their current location
2. most users not deeply familiar with the VF need broader guidance first
3. we should present broader component, blocks guidance before elements

This should make using the "correct" things more obvious.

image


sturobson (approved):

sturobson:
ooh - perhaps the title of the page should be 'components and patterns' - rather than 'components and examples'?

khawkins98:
> ooh - perhaps the title of the page should be 'components and patterns' - rather than 'components and examples'?

good shout, done.

tools/vf-component-library/src/site/components/index.njk
```diff
@@ -53,29 +74,28 @@ They're generated by the `vf-componenet-rollup`:
 {% endfor -%}
 
 <!-- some tools don't have a natural place in the component library and are linked to github -->
-<h3 class="vf-links__heading">Tools</h3>
-<ul class="vf-links__list vf-links__list--secondary | vf-list">
-  <li class="vf-list__item">
-    <a class="vf-list__link" href="https://github.com/visual-framework/vf-core/tree/develop/tools/vf-extensions">
+<h3 class="">Tools</h3>
```
sturobson, line 77: just spotted these classes have been remove.

khawkins98: yeah, it's because that looked visually awkward, here's how it looks on the current live site: 

image

I think the standard vf-content bullets are better, at least for now ... i could see a redesign of this page coming soonish. 

image

sturobson: Do we need the default `vf-list` though for styling or no? 

https://latest.visual-framework.dev//components/vf-list/index.html?q=list#vf-list--default

khawkins98: I think the default (grey, no bullets) make the lists in this context hard to scan — but as I alluded to earlier, I think this page could use a bit of a rethink in how we're show "all the the things", so I think the list is coming up short, with or without bullets.

khawkins98:
Cross-linking #1175

tools/vf-component-library/src/site/components/index.njk
khawkins98: Collecting my thoughts a little on this, I think "how others do it" show what I'm getting at. 

Little representative images are probably out-of-scope for the moment, but I think we can use the no-image cards to good effect. 

image

image

image

khawkins98: Something like this, but I dunno, doesn't feel quite right yet?

image

sturobson: maybe split patterns out completely … * coughs * live govuk 
Screenshot 2021-04-15 at 08 41 17

Issue #1483, opened 2021-04-12
Component: back-to-top element


khawkins98:
For the redevelopment of the Science in School website, we've had a request for a "back to top" button that could be shown as the user scrolls down a long page of content. 

Something like:

![image](https://user-images.githubusercontent.com/928100/114349173-db699a80-9b67-11eb-9897-9496709fc60f.png)

- In a few words, what does this component look like?
  I think this could be two variants,
  1. `vf-back-to-top--floating` the typical floating, bottom right button
  2. `vf-back-to-top--inline` an inline "Back to the top" bit of text that is fixed in position and may appear multiple times; example:
![image](https://user-images.githubusercontent.com/928100/114348939-875eb600-9b67-11eb-87bb-98c1c88bdf78.png)

- In a few words, what does this component do?
  A fancy `#top` anchor link. It should have some way of only appearing after XX% of the page or NN number of pixels has been scrolled.

- Would a rebrand change the structure or layout of this component?
  Needs to be link to the interactive colour design token (blue)

- Can other websites use this component? Or is it only useful in your organisation or website?
  It's widely usable.

- Do you have any designs or concepts?
  No, however hopefully we can base this off the existing vf-button and our Streamline Icons (https://app.streamlinehq.com/icons/streamline-bold — Cindy should be able to help pick an arrow as we move through the process, but for the first version I imagine any would do as a proof of concept)

- Have you done any user testing already?
  no; however the Nielsen Norman Group, as usual, has valuable guidance: https://www.nngroup.com/articles/back-to-top/

- How often do you expect to make use of this component?
  would be of interest to many content and longer-length scientific pages

- Can you list out major elements in this component?
  icon, screen text, button, sticky vs inline

- Will all elements of this component be used every time?
  Generally yes

- Does this component contain other components?
  vf-button

- Do you already have a component?
  No; but here are some links.

  Gov.uk:
  - https://government-frontend.herokuapp.com/component-guide/back-to-top
  - https://github.com/alphagov/govuk-design-system-backlog/issues/5

  A couple of technical examples:
  - ES6 only JS implementation: https://codepen.io/joshuamasen/pen/OYaYbL
  - CSS only implementation: https://moderncss.dev/pure-css-smooth-scroll-back-to-top/

  I imagine markup might be something like:
```
<div class="vf-back-to-top vf-back-top--{{varaint}}">
  {% if {{ floating }}  %} svg icon   {{screen reader text }} {% endif %}
[...]

PR #1490, opened 2021-04-15
Components/back to top

nitinja:
Component is fully working, except floating styles which has some issues. I am fixing those issues and will raise a new PR for that.


nitinja:
I will raise a separate PR for sibling react component.

nitinja:
For some reason, "back to top" button isnt working as expected, looking into that

components/vf-back-to-top/README.md
```diff
@@ -0,0 +1,50 @@
+# Back to top component
+
+[![npm version](https://badge.fury.io/js/%40visual-framework%2Fvf-back-to-top.svg)](https://badge.fury.io/js/%40visual-framework%2Fvf-back-to-top)
+
+"Back to top" visual framework component
```
khawkins98: No need for a line here.

components/vf-back-to-top/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+## About
+
+Takes user to top of the current page or to the target element.
+Floating variant of this component appears at right bottom of page once user scrolls down to 100% of the page height.
```
khawkins98: The "About" we try to keep at one sentence, so these two lines about the variants could be moved to the `## Usage` heading

components/vf-back-to-top/vf-back-to-top.js
```diff
@@ -0,0 +1,23 @@
+// vf-back-to-top
+
+/**
+ * JS scroll to top functionality
+ * @example vfBack-To-Top(firstPassedVar)
+ * @param {string} [firstPassedVar]  - An option to be passed
+ */
+export function vfBackToTop() {
+
+  const links = document.querySelectorAll(".vf-back-to-top");
```
khawkins98: Instead of looking for a class, we try to use JS data selectors, like:

https://github.com/visual-framework/vf-core/blob/f2d1077f48963a1cd7b04248868a185f5294ac53/components/vf-tabs/vf-tabs.js#L13

More on that: stable.visual-framework.dev//guidance/javascript/

components/vf-back-to-top/vf-back-to-top.njk
```diff
@@ -0,0 +1,22 @@
+{# Make sure any variables are listed inside the following if statement #}
+{% if context %}
+  {% set type = context.type %}
+  {% set text = context.display_text %}
+  {% set scrollToId = context.scrollToId %}
+{% endif %}
+
+<div class="vf-back-to-top vf-back-top--{{type}}">
+  <a type="submit" data-scroll-to-id="{{scrollToId}}" class="vf-button vf-button--primary vf-button--sm" aria-label="{% if text %} {{text}} {% else %} Back to top {% endif %}">
+    <span class="vf-button__text | vf-u-sr-only">Back to top</span>
```
khawkins98: I think you could drop the `vf-u-sr-only` and put your variables here, like:
```
    <span class="vf-button__text | vf-u-sr-only">
    {% if text %}
      {{text}}
    {% else %}
      Back to top
    {% endif %}
</span>
```

khawkins98 (commented):
Hi @nitinja — looking good, found a few things. If any comments aren't clear, happy to explain more. 

I still need to have a look at how it works in practice.

khawkins98:
One more thing, can we add an `href="#{{target_name}}"` or such, that way if JS fails an anchor element can still be used.

components/vf-back-to-top/vf-back-to-top.njk
nitinja: not 100% sure if I understand, but I did this:
```
{% if context %}
  {% set text = context.text %}
 ...
{% endif %}

{# Determine text, if not explicitly set -#}
{% if not text %}
  {% set text = 'Back to top' %}
{% endif %}

<div class="vf-back-to-top vf-back-top--{{type}}" data-vf-js-back-to-top {% if type === 'floating'%} data-vf-js-back-to-top-floating {% endif %}>
  <a {% if scrollToId %} href="{{'#'+scrollToId}}" {% endif %}} data-scroll-to-id="{{scrollToId}}" class="vf-button vf-button--primary vf-button--sm" aria-label="{{text}}">
    <span class="vf-button__text | vf-u-sr-only">{{text}}</span>
     ...
      {{text}}
  </a>
```

nitinja:
> One more thing, can we add an `href="#{{target_name}}"` or such, that way if JS fails an anchor element can still be used.

Implemented. I thought of doing this (hence it was anchor tag initially, then changed to button), but then I thought that No JS is an unlikely scenario nowadays.

khawkins98:
Nice. looking good. Are you done writing code? 

if so I'll open this on my local and make some final suggestions with a PR for this PR.

tools/vf-component-generator/templates/_README.md
```diff
@@ -1,7 +1,7 @@
-👋 Hi component authour, you've read the docs, right?
+👋 Hi component author, you've read the docs, right?
```
sturobson, line 1: (not a blocker)

Is there an extension you're using to change to US English? 

As EMBL is pan-eurpoean International English is the adopted version.

khawkins98: Here's a link on it https://latest.visual-framework.dev//guidance/language-british-american-english/index.html?q=British

nitinja: Is that true? I tried searching everywhere and its author, even wiktionary says it's an obsolete term for author.
image

while Cambridge dictionary has no such word.
What should I do?

nitinja:
> Nice. looking good. Are you done writing code?
> 
> if so I'll open this on my local and make some final suggestions with a PR for this PR.

Done coding, works fine in my testing. Feel free to add more comments or PR!

tools/vf-component-generator/templates/_README.md
sturobson: oh my -- I think I need an afternoon coffee -- completely read this incorrectly :D 

What's been edited in your PR is correct :D 

For 'author' it is indeed 'author' not 'authour'.

But colour should be colour etc etc still -- which isn't the issue here :D  

Resolving this conversation and going to drink a double espresso.

khawkins98:
Cross-linking #1483

khawkins98:
By my view this is looking good. 

There will be some things left to do around "what should it look like" or "what icon should it use". But as it's alpha status we can iterate on those later in collaboration with Cindy.

khawkins98 (approved):

PR #1493, opened 2021-04-19
Component: vf-back-to-top refinements

khawkins98:
As discussed, suggested enhancements for #1490


khawkins98:
Quick screen grab: 
image

nitinja (commented):
looks good

PR #1496, opened 2021-04-21
removes unsued and no longer viable `vf-badge` variants

sturobson:


sturobson:
Whilst this is still in draft I've been 'having a think'.

Originally the intention for the `vf-badge` was to denote the current 'phase' for a page. Whether it was in 'alpha', or 'beta' stage.

This idea has 'come to pass' so the original intention is now moot. 

There are, however, other instances of `vf-badge` being used for 'other things' that aren't documented (@khawkins98 I think I saw you use it for something the other day, in red‽).

So, I'm thinking

a) do we need this component? 
b) do we replicate the `banner` status variants to match

if we do b -- why not use `vf-banner` to start with?

khawkins98:
Yeah, I had stuck a vf-badge inside vf-banner to show the status of components:

image

My sense is that vf-badge is being pulled towards a more generic "label" type element. I've not thought much if that will be a good or bad thing. If it does go that way, the documentation is in need of an update: https://latest.visual-framework.dev/components/vf-badge/

sturobson:
> vf-badge is being pulled towards a more generic "label" type element

ah, now you've said that -- it makes sense to do some of the reductions in choice - but also open it up a little for `vf-banner` like variants.

Although -- if a `vf-badge` is inside a `vf-banner` should the badge be a tint or shade of the `vf-banner` colour rather than something different :D 

Perhaps -- for this PR -- we remove the green -- mainly because it needs the black text -- have the grey as `vf-badge` and then have the blue as `vf-badge--something-else`.

I'd say 'primary' here -- but we're still mixing blue/green all over the place - and it's not documented as to what/why (it's another thing 'in my head').

sturobson:
In lieu of any decisions on 'the future of the `vf-badge`' I think it'd be ok to merge this one in to make a better starting point for any future changes, additions, deprecations.

khawkins98 (approved):

PR #1499, opened 2021-04-23
the one that changes how width works so it responds to viewport

sturobson:
When introducing the new `--400` variant of the `vf-hero` we also made some small changes to how width works in the component to try and appease the academy team as much as possible without degenerating the component.

This introduced a bug on smaller viewports where the `vf-hero__content` would not resize accordingly. 

This fixes that.

### DIFF
--- a/components/vf-hero/CHANGELOG.md
+++ b/components/vf-hero/CHANGELOG.md
@@ -1,4 +1,4 @@
-### 3.3.3.
+### 3.2.3.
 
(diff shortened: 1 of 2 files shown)


components/vf-hero/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+### 3.3.3.
```
khawkins98: I think you meant 3.2.3

khawkins98 (approved):

PR #1515, opened 2021-05-07
Docs: component library navigation

khawkins98:
Attempts to improve the navigation flow (and align with typical practice) by using the vf-hero more.

Would benefit from some specific VF vf-hero imagery, but we're not there just yet, I think.

image

image


sturobson (approved):

khawkins98:
Added a few text content tweaks, mostly things that had become outdated.

PR #1527, opened 2021-05-19
the one that makes it possible to use HTML in a section header text item

sturobson:
Spilios needed to include a link in the section header text for a card container and it wasn't possible without the ` | safe` being in the .njk file.

This fixes that.


khawkins98:
safety first

khawkins98 (approved):

PR #1539, opened 2021-05-24
Fixed #1518 and #1523

### LINKED ISSUE #1518: Bug: Terms banner does not always stay on top
image

### LINKED ISSUE #1523: vf-card nunjucks template contains a potential error
image

This template takes variable `modifiers` from content but uses as `modifier`

nitinja:
Fixes #1518 and #1523

### DIFF
--- a/components/vf-card/vf-card.njk
+++ b/components/vf-card/vf-card.njk
@@ -31,7 +31,7 @@
   {%- endif -%}
   {%- if variant %} vf-card--{{variant}}{% endif %}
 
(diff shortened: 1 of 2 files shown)


sturobson:
Can you redo these as separate PRs? 

Even though they're quite small we try and stick to one PR for one thing.

I think these also both need the changelog updating so that they can be released to NPM too.

nitinja:
ok, we do. Also not sure the minor bug fixes really need changelog?

PR #1540, opened 2021-05-24
Added button link theme variant

nitinja:
Reference: https://docs.google.com/document/d/1pA436W5M0ZPhlplfIXhTzo9vThvKhE4kkdGMzUT9nHs/edit?ts=60a6779a

### DIFF
--- a/components/vf-button/CHANGELOG.md
+++ b/components/vf-button/CHANGELOG.md
@@ -1,3 +1,5 @@
+### 1.4.4
+* Added `link` theme button variant. This variant is similar to link style.
 ### 1.4.3
 
 * Added react implementation of button
--- a/components/vf-button/vf-button.config.yml
+++ b/components/vf-button/vf-button.config.yml
@@ -28,7 +28,7 @@ variants:
     context:
       text: a link variant
       button_href: "JavaScript:Void(0);"
-      theme: primary
+      theme: link
 
   # TODO: get rid of these on the next major release
   - name: regular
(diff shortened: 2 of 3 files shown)


sturobson (approved):

sturobson:
I kinda feel this came about out of confusion rather than necessity. 

As now we've created more CSS to make something more understandable -- where I would have expected the HTML element being an `<a>` and not a `<button>` would be sufficient.

Let's merge it in anyhow.

nitinja:
@sturobson not sure I fully agree.
I have been developing lots of enterprise apps: need for a "button that looks like a link" is very much real and we just cant use regular links as buttons for accessibility reason. [Here is detailed explanation](https://accessibility.oit.ncsu.edu/it-accessibility-at-nc-state/developers/accessibility-handbook/mouse-and-keyboard-events/links/link-behavior/).

sturobson:
I perhaps didn't make it clear. 

There already was a 'button as a link' option here … you would replace the `<button>` with an `<a href="">`.

What we've done here is added extra code to make up for confusion of a 'button link' in the documentation because of how it was worded, and the documentation that has gone with it.

nitinja:
@sturobson I beg to differ

`href=""` is an [invalid URL value](https://html.spec.whatwg.org/multipage/links.html#attr-hyperlink-href)  
`href="#"` takes users to top of the page, unless you `event.preventDefault()` so it's not usable as a button.
omitting href value or href attribute itself makes the link un-focusable by keyboard, so not accessible.

This component is very much required.

sturobson:
I'm clearly not explaining this well.

PR #1542, opened 2021-05-24
Fixes menu overlap in VF component site

### LINKED ISSUE #1518: Bug: Terms banner does not always stay on top

nitinja:
Fixes #1518

### DIFF
--- a/tools/vf-component-library/CHANGELOG.md
+++ b/tools/vf-component-library/CHANGELOG.md
@@ -1,3 +1,8 @@
+### 1.1.7
+
+* Fixed issue of overlapping menu on VF components site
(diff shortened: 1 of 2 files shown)


sturobson (approved):

sturobson:
just noting that _I think_ this was originally inline on the component library site because of how `vf-hero` and anything with `vf-u-fullbleed` would be on top of the tree.

nitinja:
> just noting that _I think_ this was originally inline on the component library site because of how `vf-hero` and anything with `vf-u-fullbleed` would be on top of the tree.

That will be useful info incase any related bug arises, thanks

PR #1543, opened 2021-05-24
the one that adds context to summaries 

sturobson:
some tweaks to update for 'real world usage' and added the `{% if context %}` for each variant.

A first step to a single component to rule them all.

### DIFF
--- a/components/vf-summary/vf-summary.config.yml
+++ b/components/vf-summary/vf-summary.config.yml
@@ -10,6 +10,61 @@ context:
   name: FirstName Surname
   summary__href: 'JavaScript:Void(0);'
 variants:
+  - name: article
+    context:
+      summary__href: 'JavaScript:Void(0);'
+      summary__title: EMBL's top social media posts of 2017 and what we learned from them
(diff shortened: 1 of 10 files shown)


components/vf-summary/vf-summary.config.yml
```diff
@@ -31,3 +86,5 @@ variants:
     hidden: true
     context:
       size: s
+  - name: event-easy
```
khawkins98, line 89: Is this intended to be a new variant?

sturobson: It was, an idea from the before times that never got into being used. So I’ve hidden it here. 

https://latest.visual-framework.dev//components/vf-summary/index.html?q=Summary#vf-summary--event-easy

khawkins98: ah, didn't realise that was live. Yeah, the event container. Was a nice design from Mark, but would probably need an overhaul given updated context.

sturobson:
I've a separate update to this that will suggest adding `role="presentation"` after `alt=""` if you are using an image but don't want to describe it for a screen reader.

This would be useful for training as the images used are quite abstract. 

For news - it would be a suggestion to always add good descriptive text (we can but hope).

khawkins98 (approved):

PR #1546, opened 2021-05-25
the one that updates to badges to make them consistent with buttons

sturobson:


sturobson:
I think it might be worth doing a major breaking change on buttons to match this too now? 

@khawkins98 thoughts?

components/vf-badge/vf-badge.scss
```diff
@@ -15,16 +15,9 @@ $vf-badge--primary-color--text: ui-color(white) !default;
 $vf-badge--primary--hover-color--background: color(blue--dark) !default;
 $vf-badge--primary--hover-color--border: color(blue--dark) !default;
 
-$vf-badge-rounded-border--radius: $vf-radius--xs !default;
-$vf-badge-pill-border--radius: vf-radius--pill !default;
-
 $vf-badge--outline-color--background: transparent !default;
```
khawkins98, line 18: Is vf-badge--outline being completed removed?

if so: 

1. The `vf-badge--outline-color--` variables should also be removed
2. I think the outline badge is used in quite a few places, are we sure we want it to completely go? (just double checking) ... or is the `--secondary` variant becoming an outline?

Perhaps a "what's changing and how to migrate" set of notes in the changelog would be useful?

sturobson: > The vf-badge--outline-color-- variables should also be removed

agreed

> I think the outline badge is used in quite a few places, are we sure we want it to completely go? (just double checking) ... or is the --secondary variant becoming an outline?

note the `2.0.0` version

for both buttons and badges (which will match figma)

primary - blue
secondary - blue outline
tertiary - grey

sturobson: @khawkins98 I've updated the changelog for this - I'll let you resolve this conversation and I'll merge the PR when approved

khawkins98 (approved):

PR #1547, opened 2021-05-26
the one that makes `vf-stack` dictate the flow of things a bit better

sturobson:
Initially - we want the `vf-global-header` to have a top margin of `8px` so it's close to the top of the viewport but not 'on top of it'.

This worked ok until we added `vf-stack`. 

But this is not because of `vf-stack` but because we have the GDPR `vf-banner` in the DOM as well as the `link` to the `vf-global-header` because it comes from content hub. 

This adds more ways to make the `vf-global-header` sit on the page correctly.

### DIFF
--- a/components/vf-global-header/vf-global-header.scss
+++ b/components/vf-global-header/vf-global-header.scss
@@ -19,6 +19,7 @@ $vf-global-header__link-text--color: ui-color(black);
   display: flex;
   flex-wrap: wrap;
   gap: space(200);
+  margin-top: space(200);
 
   & > [class*='logo'] {
     flex: 1; // necessary for when a logo is inside flex-based containers like vf-global-header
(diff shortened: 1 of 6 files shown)


components/vf-global-header/vf-global-header.scss
```diff
@@ -44,6 +44,14 @@ $vf-global-header__link-text--color: ui-color(black);
 // This fixes that:
 // Note: we duplicate the top margin that's already on line 22.
 
+// These are for static html pages
+link {
```
khawkins98: This would affect all contentHub imports. 

Perhaps it's better if this css lives within `vf-stack.scss`? or within `vf-content-hub-import` and scope it as: `.vf-stack link[data-embl-js-content-hub-loader]` 

_Note: that the `link > div` relationship on contentHub imports isn't always assured._

Aside: I don't think `link > div` ever happens as the HTML gets pulled out of the import.

sturobson: I think keeping it closer to where 'the problem is' is probably better

so not `vf-stack` but maybe the `vf-content-hub-import` would be better placed.

sturobson: @khawkins98 leaving this one for you to resolve conversation and I'll merge when approved

khawkins98 (approved):

PR #1548, opened 2021-05-26
the one that makes buttons stick to their naming convention rather than 'fake it'

sturobson:
Currently the 'secondary' button is actually the primary button with an outline.

This updates the variants to stick the naming convention of what we call them to be the same in the code. 

Will match up with badges PR too

### DIFF
--- a/components/vf-button/CHANGELOG.md
+++ b/components/vf-button/CHANGELOG.md
@@ -3,6 +3,11 @@
 * removes deprecated code
 * turns the primary, secondary, tertiary into actual things - rather than aliases.
 
+Mirgation Instructions
+
+* If you were using the "Outline Primary" variant you should use the "Secondary" variant now.
+  * This replaces the classes of vf-button--primary and vf-button--outline with vf-badge--secondary.
+
 ### 1.4.4
 * Added `link` theme button variant. This variant is similar to link style.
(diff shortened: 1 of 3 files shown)


components/vf-button/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+### 2.0.0-alpha.1
+
+* removes deprecated code
```
khawkins98, line 3: I think like the vf-badge pull request, this changelog would benefit from a mini "how to migrate" note.

sturobson: > a mini "how to migrate"

if the usage of existing primary, secondary, tertiary badges and buttons were implemented 'semantically' then I don't think there would be a need. 

So, yes - you're probably right -- but there probably needs more documentation as to when to use what where and why.

sturobson: @khawkins98 I've added the relevant migration details - I'll let you resolve this conversation and I will merge the PR once approved.

khawkins98 (approved):

components/vf-button/CHANGELOG.md
khawkins98: 👍

PR #1551, opened 2021-05-26
the one that updates tabs to match to current design direction

sturobson:
the one to better match

https://github.com/visual-framework/vf-core/discussions/1423

### DIFF
--- a/components/vf-tabs/CHANGELOG.md
+++ b/components/vf-tabs/CHANGELOG.md
@@ -1,6 +1,8 @@
 ### 2.0.0-rc.1
 
-* updates the styling to match the design direction.
+* Updates the styling to match the design direction.
+* This is a no-code-change update. It has been marked as visually breaking due to the major design change.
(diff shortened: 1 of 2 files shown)


components/vf-tabs/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+### 2.0.0-rc.1
+
+* updates the styling to match the design direction.
```
khawkins98: If I follow, this is a no-code-change needed migration that is mainly "visually breaking"?

sturobson: Yeah - potentially we could do a minor release -- but as it was such a big stylistic change I thought going with a major release would stop any products using it have a sudden 'wtf?' if they use dependabot etc etc

khawkins98: Cool. figured. I'll dive in and a note to the changelog about this — as I've already pinged you with enough things.

khawkins98 (approved):

PR #1552, opened 2021-05-26
the one that updates blockquote to allow for use with context and a bit more

sturobson:
This includes:

- an update to allow for using `vf-blockquote` as an 'include' in VF Eleventy sites
- includes the option for a `cite` element to help make block quote be a little more usable. 
- makes use of `vf-stack` for vertical layout
- lets `cite` just 'do it's thing' from the browser agent and just apply `font-style: italic` to the element.

### DIFF
--- a/components/vf-blockquote/vf-blockquote.njk
+++ b/components/vf-blockquote/vf-blockquote.njk
@@ -6,10 +6,8 @@
 {% endif %}
 
 <blockquote{% if id %} id="{{-id-}}"{% endif %} class="vf-blockquote{%- if override_class %} | {{override_class}}{% endif %} | vf-stack vf-stack--400">
-  {% if blockquote_text %}
-    <p class="vf-blockquote__text">{{- blockquote_text | safe -}}</p>
-  {% else %}
-    {{- html | safe if html else text -}}
+  {% if (blockquote_text) or (html) or (text) %}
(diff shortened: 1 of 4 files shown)


components/vf-blockquote/vf-blockquote.njk
```diff
@@ -1,3 +1,20 @@
-<blockquote{% if id %} id="{{-id-}}"{% endif %} class="vf-blockquote{%- if override_class %} | {{override_class}}{% endif -%}">
-{{- html | safe if html else text -}}
```
khawkins98, line 2: could we add a fallback to use `html` or `text` if `blockquote_text` is not set? (stops it from being a breaking change)

sturobson: I'll take a look

sturobson: it's a tongue twister but
```
{% if (blockquote_text) or (html) or (text) %}
    <p class="vf-blockquote__text">{{- (html) or (blockquote_text) | safe if (html) or (blockquote_text) else text -}}</p>
  {% endif %}
```

works for this

sturobson: @khawkins98 I'll let you resolve this and I'll merge when approved

khawkins98 (approved):

PR #1553, opened 2021-05-26
The one that enhances the `vf-summary`  `alt` text if the image is purely decorative.

sturobson:
This probably needs more work.

EBI Training had noticed some 'accessibility issues' which it transpires showed that the summary images did not have good alternative text for images, as well as repeated text across several instances.

Mostly, the images that are used in this page are purely decorative and don't really add any information to the content. 

To combat this there are some things that can be done to make AT (assistive technologies) realise that the image is decorative and ignore it.

This is by making sure

1)  the alt text is empty – `alt=""`
2) that `role="presentation"` is added to the `` element.

This is crudely added to the documentation and also added to the `--news` variant -- because EBI training are using the `--news-has-image` variant rather than the `--event` variant of the summary (but that is a much larger conversation).

This is a stop gap - in the hope that developers that need to have a summary that is an 'abstract', decorative image next to some text make sure the two things noted above are added accordingly.


sturobson:
Oh, merge this one after the other `vf-summary` PR

khawkins98 (approved):

PR #1562, opened 2021-06-01
bug: vf-card arrow alignment

khawkins98:
This fixes the vertical centring of the svg arrow on vf-card titles. Also aligns better with the Figma design kit.

This effectively lowers the position of the SVF by ~2px

### With fix

image

### Currently 

image

### Figma

image

### Note

- there is some slight browser inconsistency here. A SVG height of 1.1em or 1.2em are treated visually different in Chrome vs FF. 1.3em is consistent.

### DIFF
--- a/PUBLISHING.md
+++ b/PUBLISHING.md
@@ -35,10 +35,10 @@ A reference guide on how to do releases of the VF [monorepo](https://www.toptal.
 ### 4. Post-release
 
 1. commit and push changes to the `develop` branch
-    - commit message in a format of: `Component release 20200330-01`
+    - commit message in a format of: `Component release 20210531-01`
 1. add a tag"
     - see last tag `git describe --abbrev=0 --tags`
-    - add a semantic versioned tag `git tag -a v2.4.10 -m 'Release of precompiled CSS, JS, assets'`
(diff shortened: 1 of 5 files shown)


khawkins98:
The vf-section-header has the same issue which I've also applied the same fix.

And I've snuck in a small update to `PUBLISHING.md`

sturobson:
> The vf-section-header has the same issue which I've also applied the same fix.

I did notice this and wondered if there was some abstraction that could be made to reduce duplicating the css.

khawkins98:
Thinking on this a little more, the issue is that:

1. The title element line-height is 32px
2. Font size and 1em here is equal to 24px
3. We do want the 24px line height that the svg currently is (not 100% height for 32px)
4. We want to align it with Plex's cap line height/below the ascender

This particular handy little diagram seems that it [can only be sourced to stackoverflow](https://stackoverflow.com/a/25520581):
image

In this case, plex's ascender height appears to be about .1em. 

So a simpler approach may be to just add .1em padding-top.

I'll test that tomorrow — thanks for the prod @sturobson

khawkins98:
I've now updated this to directly target the font ascender. Less CSS change and it's more predictable across browsers — and easier to understand why it's there. 

image

image

image

nitinja (approved):

PR #1565, opened 2021-06-02
CI: Bump up to node 12 and 16

khawkins98:
Node 10 is now out of support and breaking on many things.

Also adds a note that node 10 is now the minimum version.


nitinja (approved):

khawkins98:
I've moved this down from node 16 to 14. 

For 16 to work, #1517 will need first to be resolved.

Issue #1572, opened 2021-06-04
Component: ELIXIR banner


khawkins98:
Many EBI services (and potentially EMBL ones) will need a VF 2 implementation of the ELIXIR banner.

Looks like:

![grafik](https://user-images.githubusercontent.com/928100/120829659-f3550d80-c55d-11eb-8df0-f69cfa91098e.png)

Options:

-  data-color: set the background color; options: orange, grey, blue, green, none (black text on white)
-  data-name: set part of the tite (data-name + "is part of the ELIXIR infrastructure")
-  data-use-cdr-logo: use the ELIXIR CDR logo instead of the kitemark
-  data-description: set the text in the small line of text
-  data-more-information-link: by default the banner will point to a page at www.elixir-europe.org. If you would prefer to link to a custom page explaining your relationship with ELXIR, add the full URL here
-  data-use-basic-styles: if you have not loaded the Foundation Framework CSS, set to true and will inject a few lines of css for basic formatting

VF 1.x docs: https://www.ebi.ac.uk/style-lab/websites/patterns/banner-elixir.html

We could potentially reuse some vf-banner .scss and .js code for this.

Prospective name `vf-banner-elixir`

Related issue: https://helpdesk.ebi.ac.uk/Ticket/Display.html?id=505212&results=3967f6a5ba9251acf6a8fea7cc6f94e5

PR #1576, opened 2021-06-07
Component: elixir banner

khawkins98:
For  #1572

- [x] finish building the template markup
- [ ] add any special css required

### DIFF
--- a/components/vf-banner-elixir/vf-banner-elixir.js
+++ b/components/vf-banner-elixir/vf-banner-elixir.js
@@ -41,6 +41,36 @@ function vfBannerElixir(scope) {
     return;
   }
 
+
+  let logo = elixirBanner[0].dataset.vfJsBannerElixirLogo;
+  let name = elixirBanner[0].dataset.vfJsBannerElixirName;
+  let description = elixirBanner[0].dataset.vfJsBannerElixirDescription;
+  let link = elixirBanner[0].dataset.vfJsBannerElixirLink;
+
+  if (!logo) {
+    logo = "https://ebi.emblstatic.net/web_guidelines/EBI-Framework/v1.4/images/logos/assorted/elixir_kitemark-60px.png";
+  }
+  if (logo == "CDR" || logo == "cdr") {
+    logo == "https://ebi.emblstatic.net/web_guidelines/EBI-Framework/v1.4/images/logos/ELIXIR/elixir-cdr.gif";
+  }
+  if (!name) {
+    name = "This service";
+  }
+  if (!description) {
+    description = "This is part of the ELIXIR distributed infrastructure for life-science information.";
+  }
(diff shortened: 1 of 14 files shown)


khawkins98:
This is now ready. 

It's a much-scaled option compared to the 1.x option — that's based on real-world usage, and to serve as an alpha-level reset for future development. 

This leverages vf-banner and vf-flag and currently avoids writing any new CSS.

image

components/vf-banner-elixir/vf-banner-elixir.react.js
```diff
@@ -0,0 +1,39 @@
...
+// We use vanilla JS templates for react for compatibility with create react app
+// ---
+import React from "react";
+import Fragment from "react-dom-fragment";
+// eslint-disable-next-line no-unused-vars
+import VfBanner-ElixirTemplate from "./vf-banner-elixir.precompiled.js"; // import templates before the nunjucks env
+import { vfNunjucksEnv } from "@visual-framework/vf-extensions-react/vf-extensions-react.js";
+
+// any JS actions needed on component insertion
+class VfBanner-ElixirCallback extends React.Component {
```
nitinja: I don't think anything with `-` is a valid name for a class, it wouldn't compile.

components/vf-banner-elixir/vf-banner-elixir.js
```diff
@@ -0,0 +1,75 @@
...
+  * The global function for this component
+  * @example vfBanner-Elixir(scope)
+ * @param {object} [scope] - the html scope to process, optional, defaults to `document`
+  */
+function vfBannerElixir(scope) {
+  /* eslint-disable no-redeclare */
+  var scope = scope || document;
+  /* eslint-enable no-redeclare */
+  const elixirBanner = scope.querySelectorAll("[data-vf-js-banner-elixir]");
+  if (!elixirBanner) {
```
nitinja: non-blocking comment: this will never be null, so unnecessary nullcheck

nitinja (approved):
React component will fail (in my opinion, could be wrong). otherwise everything's ok.

components/vf-banner-elixir/vf-banner-elixir.react.js
khawkins98: Ah, thanks there was a bug in the component generator (since fixed)

khawkins98: Pushed a fix

nitinja (approved):

components/vf-banner-elixir/vf-banner-elixir.js
khawkins98: Thanks. Updated. I've been burned by not including this check before — but in this case, yeah, redundant. good catch

PR #1577, opened 2021-06-07
Fixed null error

### LINKED ISSUE #1567: Bug: vf-back-to-top causes errors when no back to top element
When `vfBackToTop()` is invoked an error occurs if no element is found.

This is exhibited everywhere, including https://stable.visual-framework.dev/

```
Uncaught TypeError: element is null
    setVisible https://dev.assets.emblstatic.net/vf/develop/scripts/scripts.js:2948
    vfBackToTop https://dev.assets.emblstatic.net/vf/develop/scripts/scripts.js:2934
```

In these two sections of code, we need to check if the elements exist:

https://github.com/visual-framework/vf-core/blob/763069b22a02b44f0c1132adfe1c1e9c495bc7a7/components/vf-back-to-top/vf-back-to-top.js#L11

https://github.com/visual-framework/vf-core/blob/763069b22a02b44f0c1132adfe1c1e9c495bc7a7/components/vf-back-to-top/vf-back-to-top.js#L30

Can be done like in vfTabs:

https://github.com/visual-framework/vf-core/blob/763069b22a02b44f0c1132adfe1c1e9c495bc7a7/components/vf-tabs/vf-tabs.js#L13-L24

nitinja:
Fixes #1567 

Note that in case of 
```
const links = document.querySelectorAll("[data-vf-js-back-to-top]"); 
```
`querySelectorAll` call returns empty list if no element found, and we are iterating on that list, so no null check necessary.

### DIFF
--- a/components/vf-back-to-top/CHANGELOG.md
+++ b/components/vf-back-to-top/CHANGELOG.md
@@ -1,5 +1,6 @@
 ### 1.0.0-alpha.2
-* fixed a null reference error
+
+* Fixed a null reference error.
   * [Tracking issue](https://github.com/visual-framework/vf-core/issues/1567)
 
 ### 1.0.0-alpha.1
--- a/components/vf-back-to-top/CHANGELOG.md
+++ b/components/vf-back-to-top/CHANGELOG.md
@@ -1,3 +1,7 @@
+### 1.0.0-alpha.2
+* fixed a null reference error
+  * [Tracking issue](https://github.com/visual-framework/vf-core/issues/1567)
+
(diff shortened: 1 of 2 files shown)


khawkins98:
Thanks Nitin. Could you also add a changelog entry as well please.

nitinja:
ah, sorry, I forgot, adding it now

nitinja:
added changelog

khawkins98 (approved):

khawkins98:
Thanks much. Merging

PR #1597, opened 2021-06-18
enhancement: update colour tokens

khawkins98:
Driven by the changes in the design kit and colours:

* ~~Use revised primary and secondary text colours~~ removed in e07ce523f9e220a07af165c6ca62c4858cbb3de4
* Generate json for theme colours
* Display theme colours in the component library
* Deprecates old primary, secondary, tertiary themes in favour of vf-color__brand and interactive, text colours
* Changes mixin `set-type` to use the primary colour (off black) by default

We had also discussed moving to a "functional" name for these colours. For the moment I've left as "theme" as it would require more technical change and I worry it might be confused with "functions" in a programming sense... To be revisited.

The newer colour tokens still need to mapped into components.

### Screen grab of before after

image

image


nitinja (approved):

nikiforosk:
Thanks! We need to make sure that there is sufficient contrast between the text and the background, especially when the background isn't white. 

The tables at the bottom of your screengrab have two rows with gray background, does the contrast pass the AA standard there?

khawkins98:
> The tables at the bottom of your screengrab have two rows with gray background, does the contrast pass the AA standard there?

Yeah, my initial reaction was "that'll be too light". But it does still pass. The worst is the header — but it's still 7.46.

khawkins98:
I know while the revised main text colour (`#373a36`) is based off the design token colour steps, it's quite a bit lighter than the colour text used in most sites. 

- Gov.uk is `#0b0b0b`
- USWDS `#1c1c1c`
- here on github the body text is `#24292e`
- Many others are things like `#222222` 

Before we merge this in, a final check that we're happy with `#373a36`?

nikiforosk:
I've been thinking of this too, and I'd rather go for #0b0b0b instead of #000000 for primary and leave #373a36 as secondary. I think the issue is that #0b0b0b is not in our palette (unlike gov.uk).

By the way some components like vf-checkbox and vf-summary use even lighter colours for text such as #54585a and #707372. I think in those cases we should be using #373a36 instead.

khawkins98:
We have a vf-design catchup today. We can chat with design on this and see how we might be able to fit another grey into the colour scale.

grafik

cindyebi:
I just saw this thread today.

Is there a problem with #373a36 now that we make it a primary colour? If the colour passes the AA accessibility standard (which it does), is there a reason for us to add another shade of grey into the palette? Also, keep in mind that Neutral #800 is within the EMBL original palette as well. We discussed in January that we wanted to strictly keep and use the colours to use the original palette as much as possible.

khawkins98:
> Is there a problem with #373a36 now that we make it a primary colour?

I'd say look at the screenshot in the original issue. I think the worry is that the `#373a36` (800) looks a bit washed out, particularly when used for the section headers.

khawkins98:
Here are a couple more screenshots. I've also added this to the vf-design catchup to see if 800 is the darkest we can go before full black

grafik

grafik

khawkins98:
Another option would be to use the UI tokens to derive the primary text colour, where we could add a "Grey dark" colour 

grafik

```

  # VF Core Colours

  ui-color--black: 'rgb(0,0,0)'
  ui-color--grey: 'rgb(216,216,216)'
  ui-color--grey--light: 'rgb(243,243,243)'
  ui-color--yellow: 'rgb(255,250,220)'
  ui-color--red: 'rgb(211,47,47)'
  ui-color--white: 'rgb(255,255,255)'
  ui-color--off-white: 'rgb(250,250,250)'
```

khawkins98:
Following today's VF-Design catchup, we're going to:

1. Identify a few places we alternate colours to test
2. Test those with stakeholders (TBD)
3. Discuss at Tuesday's VF-UX catchup

khawkins98:
In e07ce523f9e220a07af165c6ca62c4858cbb3de4 I've reverted the revisions to the primary and secondary colours (so primary is black once again). 

I've done this so we can get the other under-the-hood technical changes in place while we await UX+Design feedback/investigations on the new primary text colour.

Issue #1587, opened 2021-06-14
Tokens: bring in functional colours, deprecate theming, grey body text


khawkins98:
With the ongoing design work to the VF by @cindyebi, we're:

- [x] Moving away from theming: https://stable.visual-framework.dev/design-tokens/theming/
- [x] Adding "functional" colours: https://www.figma.com/file/3bkXbDDIFkIH8fMEdbCGvY/Design-Library?node-id=0%3A1
- [ ] Swapping a black `#000` body text for a more accessible off-black

We should:

- [x] Deprecate [theme.palette.alias.yml](https://github.com/visual-framework/vf-core/blob/develop/components/vf-design-tokens/src/theme.palette.alias.yml)
- [x] Create functional.palette.alias.yml and add the corresponding colours:

![image](https://user-images.githubusercontent.com/928100/121878914-89cac100-cd0c-11eb-81e0-3b53f5e3e919.png)

khawkins98:
Also: 

- add some notes on when to use which functional colour
  - solid black to only be used for small text (less than 14px)
- @cindyebi to swap the secondary and primary typography colours

cindyebi:
Done swapping the typography colour labels on Figma @khawkins98

khawkins98:
Partly addressed (and much conversation) in #1597.

---

Copying some of @nikiforosk comments from Slack here:

> I haven't been able to find out why gov.uk ended up using #0B0C0C for text instead of some other shading of black (or dark grey) but this article suggests they've been using it since 2012: https://designnotes.blog.gov.uk/2019/07/29/weve-updated-the-gov-uk-colours-and-font/
> 
> Especially given that we don't have black defined in the EMBL colours so it doesn't really contradict our design system per se: https://wwwdev.embl.org/guidelines/design/page/design-guidelines/colours/
>
> My concern about #373A36 is that it was originally introduced as a secondary neutral colour but not specifically for text, and as you mentioned in github various other sites (which have hopefully explored this already) use darker tints.

--- 

I caught up with @cindyebi based on the above conversation and others and we have a planned way forward:

1. @cindyebi to pick an "almost black" that is near the gov.uk black (but not the same as to not encroach on their identity) and is also, perhaps, a touch warm like the other embl neutral colours
2. we'll use this new black in place of the current `neutral 900` (as embl does not specify an absolute black) as the new secondary text colour
[...]

PR #1606, opened 2021-06-25
enhancement: primary-text-as-grey

khawkins98:
For #1587.

Primary (big) text and headlines are now `#373a36` and small text is an off black.

Also:

* Improve reliability of mixin for `link-color` and `brand-color` when using the default color.
* Mixin for `inline-link` now uses `link-color` mixin.
* Use design tokens for colors, by using mixins for `text-color`  and `link-color`.
* Reduce `text-body--1` to 28px (was 32px) in size based on design input and feedback.
    * https://docs.google.com/document/d/13UJF2yw4C6IVXCQ1uljb6PWX20fY-gZNRlXtoTZbes4/edit#heading=h.fxkljuwfkv09

## Notes

* We now set the colour type (by default) using the tokens + `set-type` mixin — so we need to review components to ensure colours are as they should be
* @cindyebi and I have also discussed lowering the vf-text--body-1 from 32px to 28px, but that's best to do in a separate PR

### Current

image

### With changes

image


khawkins98:
Based on conversations, there's a desire to test this more widely in the browser.

Therefore we should merge this in, which will make it available on the testing latest.visual-framework.dev — we can then revise or rollback before sending into production.

nitinja (approved):

khawkins98:
Feedback has been unclear on using #373a36 for the main text.

So in in #1613 I've switched to using the clearly more accessible (but less perceptible change) of using #1a1c1a for the main text.

This importantly unblocks the next beta 2.5.0 rollup release — and ships a non-trivial accessibility win.

PR #1632, opened 2021-07-23
Added react component for vf-card

nitinja:


components/vf-card/vf-card.jsx
```diff
@@ -0,0 +1,75 @@
+import { VfClassNormalize } from "../../tools/vf-extensions-react/vf-extensions-react";
```
khawkins98, line 1: I think once shipped to npm this should be:

`import { VfClassNormalize } from "@visual-framework/vf-extensions-react/vf-extensions-react";`

PR #1636, opened 2021-07-29
enhancement: favicon urls

khawkins98:
This uses the production emblstatic CDN repo for both vf-favicon and embl-favicon to improve stability and perofrmance.

The url will not resolve until we cut the 2.5.0 release (later today).


khawkins98:
Force merging this one as I'm only around for a couple hours and it's quite minor.

PR #1650, opened 2021-09-03
deprecate: vf-activity-list and vf-activity-group

khawkins98:
These two components never saw wide adoption and neither the technical tooling nor use case was ever proven.

- https://stable.visual-framework.dev/components/vf-activity-list/
- https://stable.visual-framework.dev/components/vf-activity-group/

This deprecates them by following the guide at https://stable.visual-framework.dev/guidance/deprecating-components/

### DIFF
--- a/components/vf-activity-group/README.md
+++ b/components/vf-activity-group/README.md
@@ -4,6 +4,10 @@
 
 ## About
 
+This component has been <span style="color: rgb(228, 0, 70);">deprecated</span>.
+
+## Usage
+
 A list of updates on activities.
 
 ## Install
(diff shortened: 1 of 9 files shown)


cindyebi (approved):

components/vf-activity-group/README.md
```diff
@@ -4,6 +4,10 @@
 
 ## About
 
+This component has been <span style="color: rgb(228, 0, 70);">deprecated</span>.
```
nitinja, line 7: Nonblocking comment: style <span style="color: rgb(228, 0, 70);">deprecated</span> is not visible on preview
image

nitinja (approved):

components/vf-activity-group/README.md
khawkins98: yeah, this is actually expected. we strip the tags out in the component library view, partly as we also get the deprecated banner. However it's still useful in the markdown if you're viewing it somewhere else, like npm. 

https://www.npmjs.com/package/@visual-framework/vf-masthead

image

PR #1666, opened 2021-09-14
Feature: vf-button experemental angular support

khawkins98:
Adds experimental angular support. See README.md for more.

This is for an initial proof of concept to be circulated for feedback.

### DIFF
--- a/components/vf-button/CHANGELOG.md
+++ b/components/vf-button/CHANGELOG.md
@@ -1,7 +1,11 @@
+### 2.0.0-alpha.2
+
+* Adds experimental angular support. See README.md for more.
+
 ### 2.0.0-alpha.1
 
 * Requires at least `@visual-framework@vf-sass-config@2.6.1`.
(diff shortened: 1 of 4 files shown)


nitinja (approved):
As a test component, this should work fine (did not test it, but will do that). We are expected to follow all the standard angular practices including separate files for template, script and styles.
This also begs question: should we make separate repos for react/angular component library apart from main vf-core?

khawkins98:
> We are expected to follow all the standard angular practices including separate files for template, script and styles.

I don't think what I've done here runs counter to that -- or does it?

> This also begs question: should we make separate repos for react/angular component library apart from main vf-core?

I think we should keep the react and angular code in the same repo to keep things manageable for us, the project maintainers. If we had separate components for each component's react and angular versions, it would just be too much — at least not without being able to drop some of our other responsibilities 

From the perspective of a react or angular project, it shouldn't matter much either, as the bundler will ignore the assets it doesn't need.

PR #1672, opened 2021-09-21
Component: vf-bookmark

khawkins98:
Creates an initial vf-bookmark with a few options:

- label_inactive: Add as bookmark
- label_active: Bookmarked
- is_active: false
- modifier: button or inline
- icon: heart or star

for #1654

### DIFF
--- a/components/vf-bookmark/.npmignore
+++ b/components/vf-bookmark/.npmignore
@@ -1,0 +1,4 @@
+bin
+.github
+.travis.yml
+node_modules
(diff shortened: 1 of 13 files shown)


khawkins98:
The first draft of this component is ready for review.

image

Still to do:

- [ ] basic code cleanup
- [ ] accessibility testing
- [ ] more docs and do's don'ts
- [ ] compact version ... is this mobile only?

However this is probably nearly far enough that we can ship it as an alpha for testing in the Training projects.

khawkins98:
Showing sample active/inactive states

image

cindyebi (approved):

PR #1675, opened 2021-09-24
bug: content-hub-notifications

khawkins98:
Addresses issues when showing remote content on EMBL sites; https://gitlab.ebi.ac.uk/emblorg/backlog/-/issues/583

- when no results, embl-content-hub-loader would error if an empty message is not set
- embl-notifications was still using the deprecated `vf-banner--phase` variant


khawkins98:
Merging this in so I can prepare a release.

PR #1679, opened 2021-09-27
Component: vf-dropdown

nitinja:
Initial commit for vf-dropdown, issue https://github.com/visual-framework/vf-core/discussions/1299


nitinja:
Note that I added z-index metadata for variables, but couldn't get it working, so hardcoded values as of now. @khawkins98 do we have to run any command to compile these yaml values into scc/css?

cindyebi (approved):

khawkins98:
> Note that I added z-index metadata for variables, but couldn't get it working, so hardcoded values as of now. @khawkins98 do we have to run any command to compile these yaml values into scc/css?

Yeah, to get the design tokens to update you need to:

1. `cd components/vf-design-tokens`
2. `yarn install`
3. `gulp vf-tokens`

That will compile them and make them available. The updated files need to be committed.

Looking around the code I don't think these steps were ever documented — I'll add that to my to-do list.

components/vf-dropdown/vf-dropdown.njk
```diff
@@ -0,0 +1,25 @@
+{# Make sure any variables are listed inside the following if statement #}
+
+<div class="vf-dropdown" data-vf-js-dropdown style="width:{{width}}">
+  <div class="vf-dropdown__label-container">
+    <div class="vf-dropdown__label" style="text-align: {{labelAlign}}">{{label}}</div>
+    <div>
+      <svg class="arrow-icon" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 16 16" height="16" width="16">
```
khawkins98: should add an `aria-hidden="true"` to this, as screen readers don't need to read it out..

components/vf-dropdown/vf-dropdown.scss
```diff
@@ -0,0 +1,81 @@
...
+      .vf-dropdown__menu-item {
+        padding: 1rem 1.5rem;
+        cursor: pointer;
+
+        .vf-dropdown__menu-item-link {
+          text-decoration: none;
+          color: var(--vf-dropdown--item-color);
+        }
+        &:hover {
+          background-color: #CDE3F8;
```
khawkins98: There are quite a few `#` colours here, better to use the design tokens.

components/vf-dropdown/vf-dropdown.scss
```diff
@@ -0,0 +1,81 @@
...
+      box-shadow: 0px 3px 6px 1px #ddd;
+      padding: 0.5rem 0;
+
+      .vf-dropdown__menu-item {
+        padding: 1rem 1.5rem;
+        cursor: pointer;
+
+        .vf-dropdown__menu-item-link {
+          text-decoration: none;
+          color: var(--vf-dropdown--item-color);
```
khawkins98: Is this for optional overrides?

components/vf-dropdown/vf-dropdown.js
```diff
@@ -0,0 +1,22 @@
...
+  const components = document.querySelectorAll("[data-vf-js-dropdown]");
+
+  for (let component of components) {
+    component.addEventListener("click", (event) => {
+      // if target element has href defined, prevent it from navigating. Href is a non-js fallback
+      const d = component.getElementsByClassName(
+        "vf-dropdown__menu-container"
+      )[0];
+      d.style.display = d.style.display === "block" ? "none" : "block";
+      component.classList.toggle("menu-opened");
```
khawkins98: the "active" class is an ongoing question. but for consistency we should use either `active` or `vf-dropdown--active` or `vf-dropdown--open`

components/vf-dropdown/vf-dropdown.js
```diff
@@ -0,0 +1,22 @@
...
+export function vfDropdown() {
+  const components = document.querySelectorAll("[data-vf-js-dropdown]");
+
+  for (let component of components) {
+    component.addEventListener("click", (event) => {
+      // if target element has href defined, prevent it from navigating. Href is a non-js fallback
+      const d = component.getElementsByClassName(
+        "vf-dropdown__menu-container"
+      )[0];
+      d.style.display = d.style.display === "block" ? "none" : "block";
```
khawkins98: Can the display none/block be handled by the presence of the active class? To avoid extra inline styles.

components/vf-dropdown/vf-dropdown.jsx
```diff
@@ -0,0 +1,9 @@
+// vf-dropdown for React
+// See vf-extensions-react for usage guidance
+
+import React, { useEffect, useState } from "react";
+import "./vf-dropdown.scss";
+
+export function VFBackToTop({}) {
```
khawkins98: function name needs to be updated (`VFBackToTop`)

cindyebi:
@nitinja is it possible to see some screenshots/visuals for this?

components/vf-dropdown/vf-dropdown.scss
```diff
@@ -0,0 +1,81 @@
...
+    display: none;
+
+    .vf-dropdown__menu {
+      border-top: 1px solid #eee;
+      position: relative;
+      z-index: 8050;
+      box-shadow: 0px 3px 6px 1px #ddd;
+      padding: 0.5rem 0;
+
+      .vf-dropdown__menu-item {
```
khawkins98: Also a comment here that the Sass nesting isn't required (or always helpful).

components/vf-dropdown/vf-dropdown.njk
nitinja: you referring to vf-dropdown__label? I think that needs to be readable by SR?

components/vf-dropdown/vf-dropdown.scss
nitinja: that was incorrect variable, fixed

nitinja:
HI @khawkins98 fixed review comments except one

components/vf-dropdown/vf-dropdown.njk
khawkins98: just the svg element. in the interst of time I'll add the commit

khawkins98 (approved):

khawkins98:
I've also compiled the tokens and pushed those.

nitinja:
thanks

PR #1682, opened 2021-09-27
Initial commit for progress indicator

nitinja:

PR #1684, opened 2021-09-29
chore: drop-down and progress component cleanup

khawkins98:
Following work on #1682 and #1679 to introduce the new alpha components this cleans up a pair of rough edges.

- css custom prop usage
- ie11 fallback
- njk typo
- njk whitespace control
- add aria progress  bar role
  - maybe more work to do there in the future https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_progressbar_role
- add context conditionals
- z-index tokens
- remove unused `event` in `vf-dropdown.js`

@nitinja I thought it might be easier to show a few of these rather than explain all of them. If you like we can also discuss during the next VF call.

## Screenshots

Just as information

image

image

### DIFF
--- a/components/vf-dropdown/vf-dropdown.njk
+++ b/components/vf-dropdown/vf-dropdown.njk
@@ -1,5 +1,10 @@
-{# Make sure any variables are listed inside the following if statement #}
-
+{%- if context %}
+  {% set width = context.width %}
+  {% set labelAlign = context.labelAlign %}
+  {% set label = context.label %}
+  {% set id = context.id %}
+  {% set menuItems = context.menuItems %}
+{%- endif -%}
 <div class="vf-dropdown" data-vf-js-dropdown style="width:{{width}}">
   <div class="vf-dropdown__label-container">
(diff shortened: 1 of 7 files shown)


components/vf-dropdown/vf-dropdown.njk
```diff
@@ -1,5 +1,10 @@
-{# Make sure any variables are listed inside the following if statement #}
-
+{%- if context %}
```
nitinja, line 1: @khawkins98 this njk templates is new to me. Can you explain why its necessary to add the declarations even if works without them?

nitinja (approved):

components/vf-dropdown/vf-dropdown.njk
khawkins98: https://github.com/visual-framework/vf-core/blob/ad90662b87148957d29dda623fc3f5335bb0d96f/tools/vf-component-generator/templates/_component.njk#L1-L4

This approach is not required for all (or most) situations, but sometimes a `context` object is passed (or inherited) from a parent template, this ensure the necessary settings are available to the template.

Issue #1530, opened 2021-05-20
Documentation: don't use variants to show examples


khawkins98:
We currently make use of variants to show things that are not componentised functionality and design but are simply examples of other components....

- [vf-button--link](https://stable.visual-framework.dev/components/vf-button/#vf-button--link)
- [vf-figure alignments](https://stable.visual-framework.dev/components/vf-figure)
- probably others

We can simplify the developer options and enrich the overall documentation by creating a section of "Examples" in the component pages.

Fits with:  Discussion/notes: improving the component library browser #1175

khawkins98:
Raising this to high priority as we're getting increasing requests for "how should we use component X to do Y". I'll try to have a look this week.

PR #1689, opened 2021-10-07
Feature: allow examples of variants

khawkins98:
Adds technical support to facilitate #1530 by:

1. Adding a new component context variable `isExample: true`
2. Adding template support in the component library
3. `exampleMultiColumns: true` shows component previews in a smaller layout space

While these are technically new variants, they do not generate additional template files and are shown in a contextually appropriate way.

Incorporating examples also brings some required updates to the layout of the component pages — most of which have been overdue, such as removing the tree layout (whose function is planned to be replaced by the forthcoming mega menu).

WIP.

image

### DIFF
--- a/components/vf-badge/vf-badge.config.yml
+++ b/components/vf-badge/vf-badge.config.yml
@@ -8,6 +8,7 @@ variants:
     hidden: true
     context:
       text: alpha
+      theme: primary
 
   - name: primary
     context:
@@ -16,16 +17,19 @@ variants:
 
   - name: secondary
     context:
+      isExample: true
       text: alpha
       theme: secondary
 
   - name: tertiary
     context:
+      isExample: true
       text: alpha
       theme: tertiary
 
   - name: link - primary
     context:
+      isExample: true
       badge_href: "JavaScript:Void(0);"
       theme: primary
       text: beta
--- a/components/vf-card/vf-card.config.yml
+++ b/components/vf-card/vf-card.config.yml
@@ -2,7 +2,7 @@
 title: Card
 # Label shown on index pages
 label: Card
-status: beta
+status: live
 # preview: '@preview--nogrid'
 # The template used from /components/_previews
 #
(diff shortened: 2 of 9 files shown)


khawkins98:
image

khawkins98:
Before and after showing more contextually appropriate previews and better aligning with VF usage principles

Screenshot 2021-10-12 at 19-41-33 Local components

Screenshot 2021-10-12 at 19-41-14 Local components

nikiforosk:
This looks good, I think it will be very helpful to progressively separate actual variants from examples of how they can be applied.

I personally find the list on the sidebar quite useful (in fact more useful than searching) mostly because [I can recognise a component in the list more easily than trying to remember](https://www.nngroup.com/articles/recognition-and-recall/) what is called and searching for it. But I see we [have listed the components on the corresponding page](https://stable.visual-framework.dev/components/) so hopefully this will be a good substitute.

khawkins98:
> I personally find the list on the sidebar quite useful

Losing the tree on the individual component pages is a notable regression in usability, but removing it is the correct thing to do as:

- It's not a pattern we (currently) encourage others to do
- Occupying the left column prevents correct display of containers (like vf-card-summary)
- Removing the tree will force us to "eat our own dogfood" and work on navigation
- Our fallback remains the /search and /components pages

I have some thoughts about how we can replace the tree functionality, but it's better to address that as a separate chunk of work following this change.

nitinja (approved):

nitinja:
I found the examples section minimalistic and useful. But this prevents users from "quickly copying code of example". Instead, user has to do ???.

cindyebi (approved):

khawkins98:
> I found the examples section minimalistic and useful. But this prevents users from "quickly copying code of example". Instead, user has to do ???.

As these are examples of the main variants, the idea would be that the developer copies the code for the main variants and modifies that. However inspecting the HTML view would still be an option. 

It definitely needs ongoing consideration.

khawkins98:
Merged this in. Expect a couple more itterations, and will need to go through and activate the correct preview format per component.

PR #1695, opened 2021-10-13
Documentation: update component creation process

khawkins98:
We've had some information but it's been:

1. Hard to find
2. Out of date

This sorts that.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,0 +1,17 @@
+'use strict';
+
+const fs = require('fs');
+const path = require('path');
+const gulp = require('gulp');
+
+// Pull configuration from the package.json:
+// const {componentPath, componentDirectories, buildDestionation} = require('../vf-config');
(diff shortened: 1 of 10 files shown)


khawkins98:
Here's what the revised guidance looks like before/after.


Screenshot 2021-10-13 at 15-30-57 Components creating new ones and revising existing


Screenshot 2021-10-13 at 15-29-50 Component creation

nitinja (approved):

nitinja:
Lovely, I could add some points like, when I have more time
* Any existing components in other frameworks/libraries for reference?
* Accessibility considerations (like focus behavior and escape key function in case of a dialog)

khawkins98:
👍 

Some of that accessibility stuff is covered at https://stable.visual-framework.dev/guidance/accessibility/ -- but could certainly be better linked.

Issue #1700, opened 2021-10-22
Tabs: add a possibility to link to a particular tab


kasprzyk-sz:
Example: 

I'd like to link directly to the Password management tab on: 

https://www.embl.org/internal-information/it-services/embl-user-accounts-heidelberg/

Each tab has it's unique ID but adding it as an anchor to the URL, opens the first tab anyway.

![Screenshot 2021-10-22 at 10 40 17](https://user-images.githubusercontent.com/54020899/138422720-8fe7cc9e-839d-4560-bfba-c854e11550a6.png)

https://www.embl.org/internal-information/it-services/embl-user-accounts-heidelberg/#vf-tabs__section-73a123e3-b312-425a-aeb7-9d3e38c5f693

khawkins98:
This should be a fairly straightforward bit of JS to look for a `#vf-tabs__section` in the URL on pageload.

khawkins98:
Implemented in #1708

https://stable.visual-framework.dev/components/vf-tabs/

PR #1708, opened 2021-10-25
feature: vf-tabs `deeplinks

khawkins98:
Support activation of specific vf-tabs on page load `#vf-tabs__section=tab_id`

This feature is not designed with `vf-location-nearest` compatibility.

https://github.com/visual-framework/vf-core/issues/1700

### DIFF
--- a/components/vf-tabs/vf-tabs.js
+++ b/components/vf-tabs/vf-tabs.js
@@ -3,62 +3,27 @@
 /**
  * Finds all tabs on a page and activates them
  * @param {object} [scope] - the html scope to process, optional, defaults to `document`
+ * @param {boolean} [activateDeepLinkOnLoad] - if deep linked tabs should be activated on page load, defaults to true
  * @example vfTabs(document.querySelectorAll('.vf-component__container')[0]);
  */
-function vfTabs(scope) {
+function vfTabs(scope, activateDeepLinkOnLoad) {
   /* eslint-disable no-redeclare */
   var scope = scope || document;
+  var activateDeepLinkOnLoad = activateDeepLinkOnLoad || true;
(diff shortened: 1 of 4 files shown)


components/vf-tabs/vf-tabs.js
```diff
@@ -3,63 +3,28 @@
...
  * Finds all tabs on a page and activates them
  * @param {object} [scope] - the html scope to process, optional, defaults to `document`
+ * @param {boolean} [activateDeepLinkOnLoad] - if deep linked tabs should be activated on page load, defaults to true
  * @example vfTabs(document.querySelectorAll('.vf-component__container')[0]);
  */
-function vfTabs(scope) {
+function vfTabs(scope, activateDeepLinkOnLoad) {
   /* eslint-disable no-redeclare */
   var scope = scope || document;
+  var activateDeepLinkOnLoad = activateDeepLinkOnLoad || true;
```
nitinja, line 12: Nonblocking: This assignment can be simplified using es2015 default param values

nitinja: unless we need ie11 support

nitinja (approved):

components/vf-tabs/vf-tabs.js
khawkins98: We do use a polyfill for many IE11 things, and transpile down with Babel. worth testing if it will handle the default expressions.

khawkins98: Seems possible in principle, but would need to look at our specific Babel config and how it works in "the real world" https://babeljs.io/docs/en/babel-plugin-transform-parameters

Issue #1623, opened 2021-07-09
Navigation: in-page navigation


khawkins98:
How can we better facilitate linking within pages? Currently the `vf-links-list` does not scale well:

https://www.embl.org/sites/hamburg/
![image](https://user-images.githubusercontent.com/928100/125067240-46b60080-e0b4-11eb-8a59-b015c7049f9e.png)

Vertical and sticky menus may be an option
- https://www.covid19dataportal.org/submit-data
- https://experienceleague.adobe.com/docs/experience-manager-core-components/using/components/navigation.html?lang=en
- https://www.audi.de/de/brand/de/beratung-kauf/geschaeftskunden.html
  - https://www.audi.com/ci/en/guides/user-interface/components/navigation.html

It could be made horizontal and sticky, but risks being conflated with page-to-page vf-navigation.

a/c

- [ ] mockups
- [ ] user journeys
- [ ] summary slides for stakeholders
- [ ] action plan

sturobson:
Ahoy hoy! 

Just popping up to mention that I was planning on integrating `vf-cluster` as part of the 'on this page' navigation so that it doesn't rely on the 3 column grid and makes use of space where available. 

Spilios had made use of this somewhere by changing some classes here and there (from a copy/paste snippet I provided). 

That would at least - make better use of the space for text nodes to wrap as the space shrinks or get larger. 

aside - if the page is long enough that you have lots of 'on this page' links -- is the page _too_ long‽

khawkins98:
👋  hey Stu!

khawkins98:
We're testing a prototype evolution of the on-this-page navigation (perhaps to become `vf-navigation--on-page` as `vf-links-list` is probable to be deprecated #1649). 

![image](https://user-images.githubusercontent.com/928100/132357685-0b660400-84c1-4bcf-aac8-b659f98f7370.png)

https://www.embl.org/directory/

Stylistically this has some concerns due to the potential confusion with `vf-tabs` however we're hoping to get some real-world insights into usage (or lack of).

khawkins98:
It's been a month and I had a quick look at the metrics. 

The "on this page" does draw a number of clicks (about 10%). Most clicked were "About EMBL" "Sites" and "Find", which makes sense as:

- "Research" is already at the top 
- "Services" and "Training" are already heavily linked from across EMBL.org

So it does seem that users use it — as for satisfaction I can only say there have been no complaints and requests to use it. 

[...]

PR #1711, opened 2021-10-29
feature: vf-navigation--on-this-page

khawkins98:
Introduces a refined `vf-navigation--on-this-page` that has been trialed on https://wwwdev.embl.org/directory/ and will replace `vf-links-list-easy`

image

(download shows hover state)

Addresses: #1623

n.b. I've not fully implemented the look in #1623  as:

1. This uses more of the existing vf-navigation styling (writes less new code)
2. Still need to reconcile the horizontal padding in #1623 as we often have 5+ items in a menu


components/vf-navigation/vf-navigation.js
```diff
@@ -0,0 +1,47 @@
+// vf-navigation
+// only required for vf-navigation--on-this-page
+
+/**
+ * Function for JS scroll to top functionality
+ * That must be executed exactly once
+ * @example vfNavigationOnThisPage()
+ */
+export function vfNavigationOnThisPage() {
+  var sectionList = document.querySelectorAll("[data-vf-js-navigation-on-this-page-container='true'],[data-vf-js-navigation-on-this-page-container='1']")[0];
```
nitinja, line 10: Nonblocking: I wonder if we can just use `querySelector` instead of `querySelectorAll()[0]`

components/vf-navigation/vf-navigation.js
```diff
@@ -0,0 +1,47 @@
...
+  if (!sectionList || !section) {
+    // exit: either sections or section content not found
+    return;
+  }
+  if (sectionList.length == 0 || section.length == 0) {
+    // exit: either sections or section content not found
+    return;
+  }
+
+  Array.prototype.forEach.call(section, function(e) {
```
nitinja: No reason to still use `Array.prototype.forEach.call()` instead of `section.forEach()`, its supported on IE9+

nitinja (approved):

components/vf-navigation/vf-navigation.js
khawkins98: thanks for the tip, old habits die hard.

components/vf-navigation/vf-navigation.js
khawkins98: It's just the shorthand approach I've been using when there should only be one of a component on a page. Perhaps could be done more elegantly.

PR #1725, opened 2021-11-26
bug: vf-stack inheritance for contentHub

khawkins98:
contentHub html responses are nested deep in many layers of divs, so we ensure a default vf--stack applies to grid containers.

### DIFF
--- a/components/embl-content-hub-loader/embl-content-hub-loader.scss
+++ b/components/embl-content-hub-loader/embl-content-hub-loader.scss
@@ -41,14 +41,16 @@
 
 // contentHub html responses are nested deep in divitis,
 // so we ensure a default vf--stack applies to grid containers
-.vf-stack > .vf-content-hub-html > * > .embl-grid,
-.vf-stack > .vf-content-hub-html > * > .vf-grid {
-  margin-top:var(--vf-stack-margin--custom, var(--vf-stack-margin, 1rem));
+.vf-stack .vf-content-hub-html .embl-grid,
+.vf-stack .vf-content-hub-html .vf-grid,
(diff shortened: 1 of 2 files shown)


components/embl-content-hub-loader/embl-content-hub-loader.scss
```diff
@@ -38,3 +38,17 @@
 .vf-content-hub-html {
   --vf-stack-margin--custom: 0;
 }
+
+// contentHub html responses are nested deep in divitis,
+// so we ensure a default vf--stack applies to grid containers
+.vf-stack > .vf-content-hub-html > * > .embl-grid,
+.vf-stack > .vf-content-hub-html > * > .vf-grid {
+  margin-top:var(--vf-stack-margin--custom, var(--vf-stack-margin, 1rem));
```
nitinja: can we simplify it like, not sure if the direct children is optimal approach here? Besides, this won't work with non-direct `.vf-content-hub-html` children of `vf-stack`.
`vf-stack .vf-content-hub-html .embl-grid { ... }`

nitinja (approved):

components/embl-content-hub-loader/embl-content-hub-loader.scss
khawkins98: yeah, I actually need to revise this as I've realised for some pages there is embeded content inside tabs. A more agressive approach is propobably better for contentHub html.

khawkins98: I've now adjusted this approach.

nitinja (approved):

PR #1771, opened 2022-04-25
Enhancement: make vf-navigation--on-this-page more dynamic, responsive

khawkins98:
- allows for multiple on this page navigators
- activates the indicator only when the content is in focus
- is a breaking change! (requires a `data-vf-js-navigation-on-this-page-link` on the navigation links)

I think this is the *better* approach, but it's also disruptive as it has a breaking change.


nitinja (approved):

nitinja:
These changes should make this component more stable, thanks

khawkins98:
🥳 

As a footnote, I don't think `data-vf-js-navigation-on-this-page-container` is required any longer. But I left the references in the documentation.

Issue #1428, opened 2021-03-18
Enhancement: support Google Analytics gtag tracking


khawkins98:
Google Analytics is now strongly pushing users to switch from `ga` (analytics.js) to `gtag` (gtag.js). This means many VF sites will start switching. 

![image](https://user-images.githubusercontent.com/928100/111670162-d25f0500-8817-11eb-825a-ff0b3c4dbfb5.png)

If you don't want to use gtag you have to read the small print and find the "additional tracking methods." and then read [a mess of technical docs](https://support.google.com/analytics/answer/7538414).

![image](https://user-images.githubusercontent.com/928100/111670203-dd199a00-8817-11eb-9ac3-3a8ce15baca4.png)

## Impacts

The syntax is similar but not the same: https://developers.google.com/analytics/devguides/collection/upgrade/analyticsjs

```js
ga('create', 'GA_MEASUREMENT_ID', 'auto', 'trackerName');
ga('trackerName.send', 'event', [eventCategory], [eventAction], [eventLabel], [eventValue], [fieldsObject]);
```

```js
gtag('event', eventName, {
  'send_to': 'GA_MEASUREMENT_ID',
  'parameter1': 'value1',
  'parameter2': 'value2',
  // ...
});
```

### Example

analytics.js:

```js
// Creates the default tracker.
ga('create', 'GA_MEASUREMENT_ID', 'auto');

// Uses the default tracker to send the event to the
// Google Analytics property with tracking ID GA_MEASUREMENT_ID.
ga('send', 'event', 'Videos', 'play', 'Fall Campaign');
```

gtag.js:

```js
// Sends the event to the Google Analytics property with
// tracking ID GA_MEASUREMENT_ID set by the config command in
// the global tracking snippet.
gtag('event', 'play', {
  'event_category': 'Videos',
  'event_label': 'Fall Campaign'
});
```

### So...

This 

```js
      ga && ga("send", "event", "UI", "UI Element / " + parentContainer, linkName);
```

Would become:

```js
      gtag && gtag("event", "UI Element / " + parentContainer, {
          "event_category": "UI", 
          "event_label": linkName
       });
```
## We'll need to ...

1. A way to detect if a site is on `ga` or `gtag`
2. Some way to support event tracking without a mess of if/else statements
3. Some way to eventually deprecate and remove `ga` (probably a few years off)

For now priority is low, but we'll soon get many requests.

nikiforosk:
@sandykadam to follow up on this.

[...]

PR #1777, opened 2022-06-07
Feature: add initial GA4 support

khawkins98:
This addresses https://github.com/visual-framework/vf-core/issues/1428 and supports classic GA and GA4. I've done some testing, but further testing is recommended (note I've suggested this as 1.1.0-rc.0)

### DIFF
--- a/components/vf-analytics-google/vf-analytics-google.js
+++ b/components/vf-analytics-google/vf-analytics-google.js
@@ -146,8 +146,13 @@ function vfGaInit(vfGaTrackOptions) {
 
   if (typeof gtag === "undefined") {
     // if the site is still using legacy GA, set a dummy gtag function so we don't have to add a bunch of if statements
-    var gtag = function() {};
     vfGaLogMessage('GA4 dummy function has been set.');
+    window.gtag = function() {};
+  }
+  if (typeof ga === "undefined") {
+    // if the site is still using legacy GA, set a dummy gtag function so we don't have to add a bunch of if statements
+    vfGaLogMessage('GA UA dummy function has been set.');
+    window.ga = function() {};
   }
 
   // standard google analytics bootstrap
@@ -340,7 +345,7 @@ function vfGaTrackInteraction(actedOnItem, customEventName) {
 
   if (typeof gtag === "undefined") {
     // if the site is still using legacy GA, set a dummy gtag function so we don't have to add a bunch of if statements
-    var gtag = function() {};
+    window.gtag = function() {};
     vfGaLogMessage('GA4 dummy function has been set.');
   }
 
--- a/components/vf-analytics-google/vf-analytics-google.js
+++ b/components/vf-analytics-google/vf-analytics-google.js
@@ -429,9 +429,10 @@ function vfGaTrackInteraction(actedOnItem, customEventName) {
       gtag && gtag("event", "click ", {
         "vf_analytics": "true",
         "page_container": parentContainer,
+        "event_label": mailLink,
         "event_category": "UI",
         "event_type": "Email",
-        "email_adress": mailLink
+        "email_address": mailLink
       });
       vfGaLogMessage("Email", "Region / " + parentContainer, mailLink, lastGaEventTime, actedOnItem);
     } else if (href && href.match(filetypes)) {
@@ -442,6 +443,7 @@ function vfGaTrackInteraction(actedOnItem, customEventName) {
       gtag && gtag("event", "click", {
         "vf_analytics": "true",
         "page_container": parentContainer,
+        "event_label": filePath,
         "file_extension": extension,
         "event_category": "UI",
         "event_type": "Download",
@@ -504,6 +506,7 @@ function vfGaTrackInteraction(actedOnItem, customEventName) {
       gtag && gtag("event", "click", {
         "vf_analytics": "true",
         "page_container": parentContainer,
+        "event_label": linkName,
         "event_category": "UI",
         "event_type": "Webform",
         "link_text": linkName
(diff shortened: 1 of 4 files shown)


sandykadam:
Thanks @khawkins98, this is great. In-fact I was going to start discussion with team to prioritise this as G4 rolling out next year.

Will update you on testing of this.

nitinja:
Great, accepted, thanks

components/vf-componenet-rollup/scripts.js
```diff
@@ -20,7 +20,8 @@ vfBannerElixir();
 
 import { vfGaIndicateLoaded } from "vf-analytics-google/vf-analytics-google";
 let vfGaTrackOptions = {
-  vfGaTrackPageLoad: true
+  vfGaTrackPageLoad: true,
+  vfGa4MeasurementId: "G-N5ZLH2KNJP"
```
sandykadam: Is this intentional or left by mistake.

khawkins98: This is used by the visual-framework.dev doc site, but it probably raises more chances for confusion than help. So I made a clear generic that needs to be updated.

components/vf-analytics-google/vf-analytics-google.js
```diff
@@ -116,21 +122,30 @@ function vfGetMeta(metaName) {
...
-
-  // Need help
+  // Need help?
   // How to add dimension to your property
   // https://developers.google.com/analytics/devguides/collection/analyticsjs/custom-dims-mets
   // https://support.google.com/analytics/answer/2709829?hl=en
 
+  if (typeof gtag === "undefined") {
+    // if the site is still using legacy GA, set a dummy gtag function so we don't have to add a bunch of if statements
+    var gtag = function() {};
```
oss6: I encountered an issue when testing this. When using `var` the `gtag` is hoisted, hence it'll be `undefined` before the check (see https://stackoverflow.com/a/36911203).
A solution is to use the following `window.gtag = function() {}`

khawkins98: Hi Ossama! By pure coincidence I've been working through this code this morning, I'll be revising this PR with some fixes that include this issue.

components/vf-analytics-google/vf-analytics-google.js
```diff
@@ -189,6 +209,7 @@ function vfGaInit(vfGaTrackOptions) {
   // standard google analytics bootstrap
   if (vfGaTrackOptions.vfGaTrackPageLoad) {
     ga("send", "pageview");
```
oss6, line 226: Here `ga` may not be defined if one moves to ga4. I think we may need to use the same checks for `gtag`

khawkins98:
I've just pushed a fair number of changes (broken down into separate thematic commits) -- these address some issues that @oss6 spotted, in addition to making it a bit more resilient, improving the optional debug command and sending more appropriate events to GA4.

There's also a documentation task to be done, as GA4 has higher expectations on configuring the Google Analytics Profile for custom events (see section "IMPORTANT: Register custom definitions" at https://www.analyticsmania.com/post/how-to-track-custom-events-with-google-analytics-4/#custom-definitions) 

~I also want to point out in the readme that this approach does not work well at all when attempting to combine it with GA4 that is loaded through Google Tag Manager -- you should use the local JS approach:~ (I was wrong, see comment below)

khawkins98:
> I also want to point out in the readme that this approach does not work well at all when attempting to combine it with GA4 that is loaded through Google Tag Manager -- you should use the local JS approach:

I misspoke about that, Google Tag Manager should work fine. In my case, there was another bug causing the two clash.

khawkins98:
I'm still working my way through some of the complication between GA4 + UA + gtag + Google Tag Manager and am posting some things here before I forget them over the long weekend here:

1. In _principle_ you can use this gtag.js method with Google Tag Manager, but it seems unreliable and often leads to race conditions where this code may fail to work -- I think it's not realistic to support using this + Google Tag Manager. BUT, this is consistent with most of my reading on best practice. It is summed up well at https://measureschool.com/google-tag-manager-vs-global-site-tag/
    - > However, both these tools utilize some common resources that can lead to conflict in some cases. For example, the data layer is one such resource that is used by both tools. Therefore, it is important to be careful while adding both of them together. You will also need to test the implementation of tools by checking whether the correct data is sent to these tools. Overall, the process is doable but complicated. That’s why I would recommend not to install them together on the same page. 
    - That is: if you want to use Google Tag Manager, you should use that for your events. If you want to use gtag.js + custom JS, then vf-google-analytics is good for you. This code will provide particular value if you need a consistent way to collect event data across many websites, which is what we're after.
2. The basics of `vf-analytics-google` will be plug-and-play if using gtag+GA4, but to get the full value out if you need to add (currently) 4 "custom definitions": 
    -  grafik
    - This is an important step that will need to be done on each GA4 profile (adding the Custom Definitions will not retroactively collect data).
3. Good news: None of the changes in this PR should result in regressions for sites using UA + ga.js

This all needs to be reflected in the REAMDE.md

oss6:
Hi all, I tested the changes with the COVID-19 Data Portal, and everything seems to be fine :+1:

khawkins98:
> Hi all, I tested the changes with the COVID-19 Data Portal, and everything seems to be fine 👍

Nice. I'm hoping to get the readme updated soon.

khawkins98:
I've updated the documentation and made a few small changes to the way event messages are sent so the data format remains consistent in UA.

I believe this should now be ready for merging.

For the amount of effort that went into this PR, the diff reads quite straighforward!

sandykadam (approved):

PR #1800, opened 2022-11-14
fix/1578/add focus indicator to vf-tabs

kasprzyk-sz:


francisco-ebi (approved):
Thanks Szymon!

sandykadam:
@kasprzyk-sz @pacope92 Guys remember to update changelog file as well

kasprzyk-sz:
Right, I'll update accordingly

PR #1801, opened 2022-11-14
Make vf-dropdown keyboard accessible

francisco-ebi:
Updated the component to follow WebAIM guidelines for menubar: https://webaim.org/techniques/keyboard/#testing
Also added role and aria expanded labels


kasprzyk-sz (approved):
Looking good!

sandykadam (approved):

PR #1803, opened 2022-11-15
feat:1761/focus indicator

kasprzyk-sz:


components/vf-tree/vf-tree.scss
```diff
@@ -63,6 +63,11 @@
     transform: rotate(90deg);
     }
   }
+  &:focus:focus-visible {
```
francisco-ebi: I would wrote this one as: 
`&:focus:focus-visible svg {
    outline: -webkit-focus-ring-color auto 1px;
  }`
since it only has svg rule as its only child. It can stay as it is.

francisco-ebi (approved):
Thanks for adding the changelogs, looks good.

kasprzyk-sz:
Thanks for your comment @pacope92! I'll update the css as you suggested.

PR #1804, opened 2022-11-15
Update focus indicators of vf-banner

francisco-ebi:


kasprzyk-sz (approved):
Looks good! Are there any changes to the `top` variant needed? Currently, the outline is white on a yellow background.

components/vf-banner/vf-banner--alerts.scss
```diff
@@ -9,8 +9,8 @@
   }
   a {
     color: currentColor;
-    &:focus {
-      outline: 1px solid ui-color(white);
+    &:focus:focus-visible {
+      outline: 1px solid #373A36;
```
sandykadam: @pacope92 please don't use hardocded color code use css tokens which are already there in vf-core.
Like this https://github.com/visual-framework/vf-core/blob/develop/components/vf-social-links/vf-social-links.scss#L53.

Check for color variables in `vf-saas-starter` component https://stable.visual-framework.dev/assets/vf-sass-starter/vf-sass-starter.css

francisco-ebi: I have updated this part to use a variable that has the same color

PR #1806, opened 2022-11-17
feat:1781 vf tabs deep links

kasprzyk-sz:
Please test it in the preview mode


francisco-ebi (approved):
Works good!

sandykadam:
@pacope92 have we tested with different combination of component just like we did for embl pages with tab.

francisco-ebi:
No, I haven't tested that case yet. The combination that I think might need more work would be nested tabs (deep linking a tab that is inside another tab).

Issue #1754, opened 2022-01-31
Make the vf-dropdown component accessible using the tab button


MGS-sails:
For accessibility reasons, the vf-dropdown component should work when a user tabs over it and presses the enter button.

nikiforosk:
More specifically you should be able to access [vf-dropdown](https://stable.visual-framework.dev/components/vf-dropdown/) by pressing the tab key (like other links and buttons on a page), access the menu options with the arrows and expand it with enter (optionally) as mentioned in [the keyboard testing guide](https://webaim.org/techniques/keyboard/#testing):
 
![Screenshot 2022-02-01 at 10 36 52](https://user-images.githubusercontent.com/20280139/151954449-6a2b5d16-6881-4eb8-bceb-407edb89da87.png)

This component appears in a couple of pages at the moment (both require login):
- My learning: https://wwwdev.ebi.ac.uk/training/user 
- Online tutorial: https://wwwdev.ebi.ac.uk/training/online/courses/bioinformatics-terrified

nikiforosk:
Add ARIA labels and tab indexing 

@pacope92 Add the link to your prototype here

nikiforosk:
@pacope92 I tried this in latest and I can now access the dropdown with the keyboard (which wasn't possible before):

https://latest.visual-framework.dev/components/vf-dropdown/

I can also move through the menu items with the up/down arrow as suggested in the guide above. However, I noticed that the links within the similar menu in gov.uk are accessed via the tab (which is also [how it is on stable](https://stable.visual-framework.dev/components/vf-dropdown/)): https://www.gov.uk/

nikiforosk:
@pacope92 I looked into this a bit more and found the following example:

**[ARIA Example Disclosure Navigation Menu](https://www.w3.org/WAI/ARIA/apg/example-index/disclosure/disclosure-navigation.html#ex_label)**

- Menu opens with Space or Enter (as it is taken to be a button)
- When the menu is open you can use the tab key to move to the links
- You can also use the up/down arrow keys to move to the links but this is marked as optional

See [the keyboard support section](https://www.w3.org/WAI/ARIA/apg/example-index/disclosure/disclosure-navigation.html#kbd_label) for full description.

The navigation menu in gov.uk works like this but doesn't have support for the arrow keys: 
- Menu opens with Space or Enter (as it is taken to be a button)
- When the menu is open you can use the tab key to move to the links
- If you click on the arrow key when the menu is open then it scrolls down the page

[...]

PR #1812, opened 2022-11-21
Details with vf-dropdown

francisco-ebi:
Don't scroll on keyboard arrows press and navigate with Tab
Needs to be reviewed with Niki.
For #1754

### DIFF
--- a/components/vf-dropdown/vf-dropdown.js
+++ b/components/vf-dropdown/vf-dropdown.js
@@ -39,8 +39,8 @@ export function vfDropdown() {
       }
     });
     component.addEventListener("keyup", function(keyboardEvent) {
-      // if user presses Enter key, open the dropdown
-      if (keyboardEvent.keyCode === 13) {
+      // if user presses Enter or Space key, open the dropdown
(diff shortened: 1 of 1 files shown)


sandykadam (approved):

components/vf-dropdown/vf-dropdown.js
```diff
@@ -29,9 +31,16 @@ export function vfDropdown() {
...
+    component.addEventListener("keydown", function(keyboardEvent) {
+      if (currentPos !== null && isNavigationKey(keyboardEvent)) {
+        keyboardEvent.preventDefault();
+      }
+    });
     component.addEventListener("keyup", function(keyboardEvent) {
-      // if user presses Enter key, open the dropdown
-      if (keyboardEvent.keyCode === 13) {
+      // if user presses Enter or Space key, open the dropdown
+      if (keyboardEvent.code === 'Enter' || keyboardEvent.code === 'Space') {
```
sandykadam, line 43: @pacope92 I would suggest to use the keycodes for all key events, not sure how other language keyboards will work this code, so it is better to use keycodes instead of name.

Issue #1816, opened 2022-11-22
Outline of vf-tree not showing as full rectagnle


nikiforosk:
I noticed that sometimes the outline of a link in [vf-tree](https://stable.visual-framework.dev/components/vf-tree/) doesn't always show as a full rectangle:

![Screenshot 2022-11-21 at 09 33 58](https://user-images.githubusercontent.com/20280139/203028086-15ae125b-f6c7-47cc-a9c7-6615fd784f46.png)

This seems to happen when a parent  menu is expanded and I move backwards with the keyboard (using Shift+tab) both in stable and latest but I can't replicate it very reliably.

_Originally posted by @nikiforosk in https://github.com/visual-framework/vf-core/issues/1761#issuecomment-1321838321_

kasprzyk-sz:
A harmless way to fix it would be to increase the `z-index` for the element with the focus. 

Pull request: https://github.com/visual-framework/vf-core/pull/1820

nikiforosk:
It looks better [on latest now](https://latest.visual-framework.dev/components/vf-tree/), thank you @kasprzyk-sz:

![Screenshot 2022-11-22 at 15 46 31](https://user-images.githubusercontent.com/20280139/203358652-d8f40df6-a21b-4168-905e-cba8b50608fe.png)

nikiforosk:
The changes have now been released: https://stable.visual-framework.dev/components/vf-tree/

PR #1820, opened 2022-11-22
fix: increases the z-index for focus element

kasprzyk-sz:
Issue https://github.com/visual-framework/vf-core/issues/1816


francisco-ebi (approved):
Nice work

PR #1906, opened 2023-03-19
feat: Added Angular compatibility support for vf-badge component

bhushan-ebi:
@sandykadam please review and approve the PR


components/vf-badge/vf-badge.angular.component.ts
```diff
@@ -0,0 +1,40 @@
...
+  styleUrls: []
+})
+
+export class VfBadgeAngularComponent implements OnInit {
+  /* Initialize values based on input values */
+  @Input() badge_href = '';
+  @Input() theme = '';
+  @Input() text = '';
+  @Input() style = [];
+  @Input() html = '';
```
francisco-ebi, line 19: I'm assuming that through this input, you want to provide anyone that use the component, the option to put some HTML content inside the vf-badge. If that is the case, to display it properly you would need to do something like this: `<a [innerHTML]="content"></a>`

components/vf-badge/vf-badge.angular.component.ts
```diff
@@ -0,0 +1,40 @@
...
+  @Input() override_class = '';
+  @Input() id: string | undefined;
+  content =  '';
+  class = 'vf-badge ';
```
francisco-ebi, line 23: I'm not 100% sure for this one but if you are setting state inside the component based on data from outside, you would need to add `ngOnChanges` to update the inner state correctly.

components/vf-badge/vf-badge.angular.component.ts
```diff
@@ -0,0 +1,40 @@
...
+  `,
+  styleUrls: []
+})
+
+export class VfBadgeAngularComponent implements OnInit {
+  /* Initialize values based on input values */
+  @Input() badge_href = '';
+  @Input() theme = '';
+  @Input() text = '';
+  @Input() style = [];
```
francisco-ebi, line 18: You could specify the different types of badges, like this: 
`@Input() style: Array<'tertiary' | 'secondary' | 'primary'> = [];`
It would be something nice to have, as a developer

components/vf-badge/vf-badge.angular.component.ts
bhushan-ebi: Thanks @pacope92. Changes added with latest commit. Please review.

components/vf-badge/vf-badge.angular.component.ts
bhushan-ebi: Thanks @pacope92. Changes added with latest commit. Please review.

components/vf-badge/vf-badge.angular.component.ts
bhushan-ebi: Thanks @pacope92. Changes added with latest commit. Please review.

sandykadam (approved):

PR #2054, opened 2024-06-21
Changed : Issue-2038 - Updated documentation for typos, enabled variants

bhushan-ebi:


components/vf-social-links/vf-social-links.scss
```diff
@@ -80,6 +80,11 @@
   }
 }
 
+.dark-mode {
+  display: inline-block;
+  background-color: set-ui-color(vf-ui-color--black);
+  padding: 0 20px;
+}
```
sandykadam: @bhushan-ebi  Why there is a need to define this class explicitly when it is already defined in Sass including BG color here file https://github.com/visual-framework/vf-core/blob/develop/components/vf-social-links/vf-social-links.scss#L84-L119

dark-mode will work as it is even in current page of news if you add dark-mode class - https://www.ebi.ac.uk/about/news/

bhushan-ebi: @sandykadam it's actually not defined there but instead its a selector for the elements having class=dark-mode. It'll apply to the child elements and I had to define .dark-mode class and then apply it to main container so that it shows with black background and child elements are in white.

sandykadam: Sorry I didn't get you why exactly it is required?
Below is example on live I added class.

Screenshot 2024-06-21 at 17 09 42
Screenshot 2024-06-21 at 17 09 46

bhushan-ebi: 'dark-mode' selector is working correctly and I didn't add any change there. Since you added it to UL, [class*='dark-mode'] selector made the icons inside white. But that didn't change the UL (parent element) to black(or any other colour). Just for the example I've added the 'dark-mode' class so that the parent shows black. Users can override it in their code. I can remove it if it's better. Please check the screenshot attached.

Screenshot 2024-06-21 at 17 43 51

sandykadam (approved):

components/vf-social-links/vf-social-links.scss
bhushan-ebi: @sandykadam I've deleted the .dark-mode class

PR #2086, opened 2024-08-02
Chore: Issue 1647 - Dependency updates for multiple packages

bhushan-ebi:


tools/vf-component-library/src/site/_includes/layouts/components.njk
```diff
@@ -283,7 +283,7 @@ This approach is best for bare-bones Nunjucks environments, such as precompiled
 %}{% endraw %}
                 </code></pre>
               </details>
-              {% set angularComponents = "vf-badge, vf-banner, vf-hero, vf-button, vf-blockquote" %}
+              {% set angularComponents = "vf-badge, vf-banner, vf-hero, vf-button, vf-blockquote, vf-tabs" %}
```
sandykadam: should we have this change in Tabs angular PR?

.github/workflows/nodejs.yml
```diff
@@ -12,9 +12,9 @@ jobs:
         node-version: [14.x, 16.x]
```
sandykadam: I think you should update this node version to 16 & 18. Then I think builds will work for both PRs

tools/vf-component-library/src/site/_includes/layouts/components.njk
bhushan-ebi: Added this change to the Tabs PR.

.github/workflows/nodejs.yml
bhushan-ebi: This change is done.

bhushan-ebi:
Hi @sandykadam , If I use node-version: [18.x] Node CI build succeeds but other checks (Header, Pages changed, Redirect rules, netlify) fail. But when I use node-version: [16.x, 18.x] Node CI build as well as all these checks fail. Any suggestions?

PR #2091, opened 2024-08-06
Chore: Issue 1647 - Dependency updates for multiple packages

bhushan-ebi:


sandykadam (approved):

sandykadam:
before I approved it was working so we should have merged that first and then we could have commit new changes, your last commit is breaking again. So remove that lets merge what it was there before approval.

bhushan-ebi:
Hi @sandykadam 

I've raised a new PR https://github.com/visual-framework/vf-core/pull/2096 (from other branch) after merging the approved PR to develop. IT still seems to be failing.

Regards,
Bhushan

PR #2110, opened 2024-08-26
Updated : Issue 1750 - Made the image in news feed clickable

bhushan-ebi:


sandykadam:
@bhushan-ebi I don't see this working on deploy site, I wanted to check how does it work as we had long discussions in past and this requirement was not working as expected so it was on hold.

sandykadam (approved):

PR #2186, opened 2025-02-03
chore: Issue 1932 - Test cases for Angular components

bhushan-ebi:
vf-badge, vf-blockquote, vf-button, vf-hero


sandykadam:
@bhushan-ebi  Does it need to have to be in "lib" folder ? can it stay in root.
Also how are you making sure test case works if anyone make changes or push the code

bhushan-ebi:
> @bhushan-ebi Does it need to have to be in "lib" folder ? can it stay in root. Also how are you making sure test case works if anyone make changes or push the code

@sandykadam It doesn't need to be but since the testcase file is written for the Typescript (ts) extension I've kept it in 'lib' folder along with the .ts file to avoid confusion.

sandykadam (approved):

PR #2212, opened 2025-02-25
Component release 20250225-01

bhushan-ebi:


sandykadam:
@bhushan-ebi build is failing, update component-library and release new version

bhushan-ebi:
> @bhushan-ebi build is failing, update component-library and release new version

@sandykadam this is fixed now.

PR #2231, opened 2025-03-10
fix: Issue 2129 - Changes added to fix cropped display of progress in…

bhushan-ebi:
…dicator


sandykadam (approved):

sandykadam:
FYI this is not used anywhere, but we can make use of it for loading pages, data etc

PR #2289, opened 2025-07-01
Vf chatbot

bhushan-ebi:


components/vf-chatbot-selector/vf-chatbot-selector.scss
```diff
@@ -94,7 +94,16 @@
...
+      visibility: hidden;
+      position: absolute;
+      border: 0;
+      padding: 0;
+      margin: 0;
+      height: 1px; 
+      width: 1px;
+      overflow: hidden;
+  }
   input {
```
sandykadam, line 107: @bhushan-ebi if you are using this for accessibility, then we already have class `vf-u-sr-only` which need to added to HTML element

PR #2370, opened 2026-04-14
Minimal footer

mshujamalik-ebi:


sandykadam:
@mshujamalik-ebi on white varient links hover doesn't look correct, please check

components/vf-minimal-footer/vf-minimal-footer.scss
```diff
@@ -9,4 +9,8 @@
   .vf-minimal-footer .vf-logo__image {
     width: 80%;
   }
+}
+
+.vf-u-background-color-ui--white .vf-footer__link:hover {
+    color: #000;
```
sandykadam: don't hardcode colors, there are already classes of colors, use those variables
e.g `--vf-ui-color--black`

mshujamalik-ebi: fixed!

bhushan-ebi:
Hi @mshujamalik-ebi, please note the issues in this PR as per my current observation:

1. The Readme.md doesn't have the correct content as per the footer content provided by Ux team [here](https://docs.google.com/document/d/1EoVdgVzQmqWyB3lPzj3sl51Pop3zA4Qoazcty-RuMhA/edit?usp=sharing) and as discussed earlier
2. The page needs to look the way it's shown [here](https://www.figma.com/design/7MlAdBtZv5cORXDS3QGGdF/VF---Footer-lighter-version-exploration?node-id=1153-4500&p=f&t=FujEbv3czX1mV7xA-0) as discussed.
3. The code would just show 2 variants as per the .njk file you added. It needs to account for the different variants (as per design) which should each show a different HTML. You'll need to add these variants in .config.yml file and when each context is read in .njk file, it would render the particular version.
4. I'd suggest to have dark and light variants rendered based on styling instead of having different HTML for them.

Regards,
Bhushan

PR #2374, opened 2026-04-28
Remove terms of use link from Banner component

mshujamalik-ebi:


package-lock.json
```diff
@@ -0,0 +1,8974 @@
+{
```
bhushan-ebi: @mshujamalik-ebi This file is probably added by mistake. Can you please remove it as it's not needed.

bhushan-ebi (approved):

PR #2390, opened 2026-06-24
chore: Issue 2389 - Added new configuration for k8s deployment

bhushan-ebi:


sandykadam:
@bhushan-ebi this is external domain so it needs to go on external k8s, plz update accordingly then we can merge

sandykadam (approved):

